PR #5, opened 2026-05-31
feat(openlogi-hid): route HID++ writes to directly-attached devices

AprilNEA:
## What & why

Today every control path assumes a device hangs off a **Logi Bolt receiver** at a pairing slot. Directly-attached devices (USB cable or Bluetooth) are second-class: Bluetooth ones list but are read-only, and wired ones (no battery — e.g. a corded **G502**) are dropped entirely. This makes them first-class.

Three blockers removed:

| Before | After |
|---|---|
| `probe_direct` required a battery → wired mice dropped | **Hybrid discriminator**: keep a direct device if it reports a battery **OR** exposes a control feature (adjustable DPI `0x2201/0x2202` or ReprogControls `0x1b04`). A receiver's secondary `0xff` interface exposes neither, so the phantom-entry guard still holds. |
| `with_device` / `open_target_channel` bailed unless `Receiver::Bolt` | New `DeviceRoute { Bolt { receiver_uid, slot } \| Direct { vendor_id, product_id } }`; `open_route_channel` resolves both in one place. Direct devices are addressed at HID++ self-index `0xff` and re-found by vid/pid. |
| `DpiTarget` required a `receiver_uid` → direct devices got `dpi_target = None`, so DPI/SmartShift/capture were never wired up | The GUI builds a `Bolt` or `Direct` route per device; DPI, SmartShift, and gesture/button capture all route through it. |

`DpiTarget`/`GestureTarget` collapse into `DeviceRoute`. The CLI `diag` commands and `SharedChannel` reuse follow the same route, so `openlogi diag {features,dpi,smartshift}` work on direct devices too.

## Test plan

Run in the project devenv (real-Xcode Metal env for the GUI):

- [x] `cargo clippy --workspace --all-targets -- -D warnings` → clean
- [x] `cargo fmt --check` → clean
- [x] `cargo test` → non-GUI 14 passed, GUI 6 passed, 0 failed
- [x] Cargo.lock untouched; gpui pin (`eb2223c0`) unchanged

## Still needs real-hardware verification (no G502 on hand)

- [ ] **Enumeration filter** (`transport.rs`): only Logitech VID + usage page `0xff00` / usage `0x0002` are enumerated — **unchanged here**. If a wired G502 exposes HID++ on a different vendor usage page (some G-series use `0xff43`), it won't appear at all. This is the first thing to confirm with `openlogi diag features`.
- [ ] **Feature discriminator**: assumes HID++ 2.0 with `0x2201/0x2202`/`0x1b04`. A variant speaking only legacy HID++ 1.0 registers would be skipped.
- [ ] Two identical direct mice on one host are indistinguishable by vid/pid (first match wins) — documented as a v0 limitation.

### DIFF
--- a/crates/openlogi-cli/src/cmd/diag/dpi.rs
+++ b/crates/openlogi-cli/src/cmd/diag/dpi.rs
@@ -14,10 +14,10 @@ pub struct DpiArgs {
 }
 
 pub async fn run(args: DpiArgs) -> Result<()> {
-    let (uid, slot, name) = first_online_device().await?;
-    println!("device: {name} (slot {slot}, receiver {uid})");
+    let (route, name) = first_online_device().await?;
+    println!("device: {name} ({route})");
 
-    let before = openlogi_hid::get_dpi(Some(&uid), slot)
+    let before = openlogi_hid::get_dpi(&route)
         .await
         .context("read current DPI")?;
     println!("  current DPI: {before}");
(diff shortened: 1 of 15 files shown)


radfaraf:
I have a G502 X which is wired. Is there anything I can help you with to get it working here? Here are images of what it shows in the v0.5.3 for that mouse on my mac. Not a programmer but I could test stuff and give you information as far the results if that could get you further along. 

Screenshot 2026-06-06 at 5 47 35 PM

Screenshot 2026-06-06 at 5 47 55 PM

PR #20, opened 2026-05-31
fix: detect MX Master 4 over macOS Bluetooth

### LINKED ISSUE #17: Not working with Logitech M535
I'm trying to connect my M535 mouse but the app is not detecting it while on macOS settings is detecting the mouse in pairing mode.

Image

If I try to first pair it on macOS setting ad then open OpenLogi the mouse is not registered.

Image

threepeakschallenge:
## Summary
- include Logitech generic mouse HID nodes as direct-device candidates
- add a macOS `hidutil --ndjson` fallback for MX Master 4 when HID probing yields no inventory
- synthesize the MX Master 4 model identity so the GUI can render the device over Bluetooth

## Notes
This fixes detection/presence for MX Master 4 over macOS Bluetooth. It does not claim full HID++ feature write support when macOS rejects output reports for the BLE HID node.

## Test plan
- `cargo check -p openlogi-hid`
- `cargo test -p openlogi-hid hidutil_parser -- --nocapture`

Fixes #17

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4462,6 +4462,7 @@ dependencies = [
  "futures-lite 2.6.1",
  "hidpp",
  "openlogi-core",
+ "serde_json",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


AprilNEA:
Thanks for the contribution and the MX Master 4 investigation — it was really helpful.

I’m going to merge #28 instead, since it addresses the root cause more generally at the HID++ transport layer for BLE-direct devices, rather than adding a device-specific fallback. If MX Master 4 still has issues after #28, we can revisit a narrower follow-up.

PR #21, opened 2026-05-31
fix: refresh device inventory when reopening GUI

threepeakschallenge:
## Summary
- refresh the global `AppState` when a new `AppView` receives a non-empty inventory
- refresh inventory before focusing an existing main window opened via dock/tray reopen paths
- update the startup path so a pre-existing global state does not remain stale

## Test plan
- `cargo check -p openlogi-gui`

This is split from the MX Master 4 Bluetooth detection change so the GUI lifecycle fix can be reviewed independently.

### DIFF
--- a/crates/openlogi-gui/src/main.rs
+++ b/crates/openlogi-gui/src/main.rs
@@ -191,8 +191,8 @@ fn main() -> Result<()> {
             // Install the hook-shared AppState up front, then open the window at
             // launch; closing it leaves the app live in the menu bar.
             cx.update(|cx| {
+                let cache = asset::AssetResolver::new();
                 if !cx.has_global::<AppState>() {
-                    let cache = asset::AssetResolver::new();
                     cx.set_global(AppState::with_runtime_shared(
                         initial_config,
                         &inventories,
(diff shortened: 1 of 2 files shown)


crates/openlogi-gui/src/main.rs
```diff
@@ -337,6 +341,27 @@ fn main_window_options(cx: &mut gpui::App) -> WindowOptions {
...
+                warn!(error = %e, "could not refresh inventory before opening window");
+                Vec::new()
+            }
+        };
+        refreshed_inventories.as_slice()
+    } else {
+        inventories
+    };
+
+    if !inventories.is_empty() && cx.has_global::<AppState>() {
```
AprilNEA, line 358: Should this refresh even when `inventories` is empty? An empty inventory can be a valid successful enumeration (“all devices disconnected”), but this guard skips the refresh and can leave stale devices in `AppState`. Maybe only skip refresh when `enumerate_blocking()` failed.

AprilNEA (changes requested):

AprilNEA:
Thanks for splitting this out — closing this one.

After this PR was opened, 882c388 ("open the window immediately; never block startup on enumeration") landed, which removed enumerate_blocking() because the synchronous block_on probe could hang forever on an asleep device and freeze launch. This PR re-introduces that exact main-thread block on every reopen, so it (a) no longer compiles against master (hence the conflict), and (b) is now unnecessary: the 2 s inventory watcher keeps the global AppState fresh independent of window presence, so a reopened window already reads a ≤2 s-old list.

The MX Master 4 detection piece is the valuable half — let's keep that going in #40.

Issue #25, opened 2026-06-01
Doesn't work with MX Master 3S for Mac


markcharlesobrien:
I start OpenLogi with my MX Master 3S for Mac paired over Bluetooth. OpenLogi says "No device connected. Plug in or pair a supported Logitech device – it'll show up here automatically." But it doesn't.

I plug it in with a USB-C cable connected directly to my Mac, nothing happens.

I go to "Add Device," it says 'Searching for devices... Make sure the device is on and in pairing mode."

I put the mouse into pairing mode and nothing happens.

Logi Options + is uninstalled and I've rebooted.

Logi Options+, when installed, finds the mouse immediately.

Now I have to reinstall Logi Options+ to get my scroll wheels and side buttons back.

Mouser did the same thing.

All these programs going around claiming to free us from Logitech's terrible software, but none of them work.

M2 Ultra Mac Studio, macOS 15.7

markcharlesobrien:
Two useless replies appear immediately, "pullfrog" with some frog icon, and "linear-code" which takes me to a site I've never heard of that wants me to make an account. Wtf?

davidbudnick:
Hello @markcharlesobrien 👋🏻 

Sorry to hear you are having issues with the application, I as well have the MX Master 3S and it is paring correctly.

1. What version are you currently on of OpenLogi?
2. Are you using the Logitech Logi Bolt Receiver or over Bluetooth? (I know there is a current issue with Bluetooth should be solved in #20)

markcharlesobrien:
The app says v0.2.0 and I am connected over Bluetooth

bpaiement:
Same issue with an **MX Master 3** (non-S). App stays on "No device connected" even though the mouse works fine. Tried re-pairing and switching channels. No Logi Options+ installed.

### Environment
- OpenLogi **v0.2.0** · macOS **26.5** (25F71) · Apple **M4**
- MX Master 3, VendorID `0x046D` / ProductID `0xB023`, connected via **Bluetooth LE**
- Accessibility ✅
- Tested with adding Input Monitoring ✅

### Diagnostics
- `OPENLOGI_LOG=debug`: `HID++ candidate interfaces count=0` every scan
- `ioreg -r -c IOHIDDevice -l`: device shows `Transport = Bluetooth Low Energy`, every interface `PrimaryUsagePage = 1`, no `0xFF00` interface

latte-z:
[...]

Issue #19, opened 2026-05-31
UI Broken for MX Master 3S + MX Keys Mini


kodi:
<img width="1212" height="890" alt="Image" src="https://github.com/user-attachments/assets/69bdc63d-bde6-42f7-bf43-c9fcd819a14c" />

1) No mouse preview
2) Clicking on a MX Keys mini - nothing happens, same view
3) Toolbar sometimes shows keyboard battery, or mouse battery (maybe lets show both?)

<img width="232" height="138" alt="Image" src="https://github.com/user-attachments/assets/ac1ee199-4357-4dc9-bc67-bb666195d34d" />

davidbudnick:
@kodi your first issue you listed above => "No mouse preview" should be solved with #24.

AprilNEA:
Thanks for the report — the mouse preview / asset matching part should be fixed in v0.3.3 by the displayName fallback for devices whose live PID does not match the asset registry (the MX Master 3S case here).

Please try v0.3.3 or newer once available and let me know if the MX Master 3S preview still does not render.

For the MX Keys Mini part: keyboard support / interaction is tracked separately in #30, so I’ll keep that discussion there rather than mixing it with this UI/asset issue.

PR #26, opened 2026-06-01
fix(assets): resolve assets by displayName when no PID matches (MX Master 3S over BTLE)

jcbrizu:
## Problem

On an **MX Master 3S** connected through a Logi Bolt receiver, OpenLogi shows no device image — the GUI falls back to the synthetic silhouette and the asset depot is never downloaded (`sync: no matching depots for connected devices`).

**Root cause:** over BTLE the device's HID++ `DeviceInformation` (feature `0x0003`) reports `model_ids = [0xb034]`, `extended_model_id = 1`, i.e. `config_key = "1b034"`. The asset registry, however, keys the MX Master 3S as **`2b043`** (a different transport's PID). `resolve_in_index` only matches by PID — strict `{ext}{pid}` and bare-PID suffix — and the device never reports `b043` under any transport, so nothing matches.

Observed with `openlogi list` on hardware:

```
slot 2 ● MX Master 3S (mouse, wpid=b034)
        model_ids=[b034,0000,0000] ext=01 transports=btle
```

## Fix

Add a final resolution step that bridges the firmware **codename** to the registry **`displayName`** (exact, case-insensitive). `"MX Master 3S"` lines up even when no PID does. This generalises to any BTLE-connected device whose live PID is absent from the registry.

The codename is threaded through `resolve_in_index`, `AssetResolver::resolve`, and the asset sync path (`collect_models` now carries `(DeviceModelInfo, Option<String>)`). PID matching still takes precedence; the name match is a last resort, so existing devices are unaffected.

## Verification

- `cargo test -p openlogi-assets -p openlogi-gui` — green, including new tests for `Index::find_by_display_name` and for the codename fallback in `resolve_in_index` (a BTLE 3S model misses by PID, resolves by codename).
- On hardware (MX Master 3S / Bolt / macOS 26): the log now shows `asset matched via codename↔displayName fallback depot="mx_master_3s"`, the `mx_master_3s` depot downloads, and the mouse image renders.

Likely also helps the missing-image symptom reported in #19 / #25.


AprilNEA (dismissed):

AprilNEA (approved):

AprilNEA:
Rebased onto `master` (dropping the merge commit) and landed directly:

- `7505fa5` — your fix, authored by you
- `f64f40c` — a one-line `serial_number: None` follow-up the `btle_3s_model` test helper needed after master added that field to `DeviceModelInfo` (otherwise the merged tree didn't compile)

Thanks @jcbrizu! Closing since the change is now on master.

PR #29, opened 2026-06-01
feat: add wired G-series keyboard RGB control

davidbudnick:
Adds wired Logitech G-series keyboard support (G513, G512, G610, G810, G815, G Pro…).

- Detects wired G keyboards via the `0xff43` HID++ usage page (Bolt/Unifying was `0xff00`-only)
- Solid-colour per-key RGB over `0x8080`, wired to the lighting panel (swatches, on/off, brightness slider)
- All connected devices shown in the menu-bar tray, with stable ordering + refresh hysteresis so flaky scans don't reshuffle
- CLI: `openlogi diag lighting <RRGGBB>`

### Screenshots
Screenshot 2026-06-02 at 10 26 56 AM

Screenshot 2026-05-31 at 11 04 00 PM


crates/openlogi-gui/src/components/lighting_panel.rs
```diff
@@ -0,0 +1,197 @@
...
+                            .child(tr!("BRIGHTNESS")),
+                    )
+                    .child(
+                        div()
+                            .text_xs()
+                            .text_color(rgb(ACCENT_BLUE))
+                            .child(format!("{}%", lighting.brightness)),
+                    ),
+            )
+            .child(Slider::new(&self.brightness).horizontal())
```
AprilNEA, line 135: The slider's `SliderState` is initialized only once in `new()`, but `render()` reads the current device's `lighting.brightness` from `AppState`. When switching devices or after another control changes lighting, the label can show the current brightness while the thumb still holds the previous value; releasing it can then write the stale slider value into the newly selected device. Please keep the slider state synchronized with `AppState::lighting().brightness` when the active device/config changes, or make the slider controlled from `AppState` instead of long-lived local state.

crates/openlogi-gui/src/components/device_view.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::theme::Palette;
+
+pub fn device_view(
+    record: &DeviceRecord,
+    lighting_panel: &Entity<LightingPanel>,
+    pal: Palette,
+) -> AnyElement {
+    // Wired G-series keyboards arrive through the direct path as `Unknown`, so
+    // treat those as keyboards too and offer the lighting controls.
+    let is_keyboard = matches!(record.kind, DeviceKind::Keyboard | DeviceKind::Unknown);
```
AprilNEA: This treats every `DeviceKind::Unknown` as a keyboard and shows lighting controls for it. The comment says wired G-series keyboards arrive as `Unknown`, but this is broader than that: any future unknown non-pointer device would get a keyboard RGB panel and attempts to write feature `0x8080`. Could this be narrowed to confirmed keyboards, or at least to direct wired unknown devices with a lighting-capable signal/probe?

crates/openlogi-cli/src/cmd/diag/lighting.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub color: String,
+    /// Brute-force probe: fire many candidate RGB commands ~1.5s apart so you
+    /// can watch which one flashes the keyboard.
+    #[arg(long)]
+    pub spam: bool,
+}
+
+pub async fn run(args: LightingArgs) -> Result<()> {
+    let hex = args.color.trim_start_matches('#');
+    let rgb = u32::from_str_radix(hex, 16)
```
AprilNEA, line 22: The command documents `RRGGBB`, but this accepts any length that parses as hex. For example `f` becomes `#00000f`, and `1000000` silently truncates through the later 24-bit extraction. Please validate `hex.len() == 6` and that all bytes are ASCII hex digits before parsing, so the CLI matches the error message and help text.

crates/openlogi-hid/src/write.rs
```diff
@@ -224,6 +308,106 @@ async fn toggle_smartshift_on_channel(
...
+        }
+        variants.push((
+            "V9 g810 setGroupKeys 0x12 + commit 5a",
+            vec![verylong(0x0c, 0x3a, &p), long(0x0c, 0x5a, &[])],
+        ));
+    }
+    for (label, reports) in &variants {
+        println!("  -> {label}");
+        for rep in reports {
+            let _ = writer.write_output_report(rep).await;
```
AprilNEA: This ignores every write error during the diagnostic spam sequence, so the command can print progress and finish successfully even if no report was accepted by the device. Since this is meant to diagnose hardware behavior, returning or at least logging the first failed report would make failures visible instead of looking like a non-responsive keyboard/protocol variant.

crates/openlogi-hid/src/lib.rs
```diff
@@ -28,5 +28,6 @@ pub use route::{DIRECT_DEVICE_INDEX, DeviceRoute};
 pub use smartshift::{SmartShiftMode, SmartShiftStatus};
 pub use write::{
     FeatureEntry, SharedChannel, WriteError, dump_features, get_dpi, get_smartshift_status,
-    set_dpi, set_dpi_on, toggle_smartshift, toggle_smartshift_on,
+    set_dpi, set_dpi_on, set_keyboard_color, spam_lighting, toggle_smartshift,
```
AprilNEA: Do we want `spam_lighting` to be part of the public `openlogi-hid` API? It reads like a reverse-engineering probe for the CLI rather than a stable library operation. If it needs to stay, consider documenting it as diagnostic/experimental and avoid swallowing write errors in the implementation; otherwise keeping it internal to the CLI path would reduce the public surface.

AprilNEA (changes requested):
Thanks for the PR — the overall direction looks good and CI is green, but I think a few correctness/UI issues should be fixed before merging. The main blockers are the lighting slider state drifting from the selected device/config, treating every `Unknown` non-pointer device as a keyboard, and accepting non-`RRGGBB` values in the CLI despite the documented contract.

crates/openlogi-hid/src/lib.rs
davidbudnick: I ended up just removing it for right now, I completely forgot the code was still in there from my debugging session last night 🤦🏻

crates/openlogi-hid/src/write.rs
davidbudnick: ✅

crates/openlogi-cli/src/cmd/diag/lighting.rs
davidbudnick: ✅

crates/openlogi-gui/src/components/device_view.rs
davidbudnick: ✅

crates/openlogi-gui/src/components/lighting_panel.rs
davidbudnick: ✅

davidbudnick:
@AprilNEA Thanks for your comments they all should be addressed, please let me know if you find any other issues!

AprilNEA (approved):

AprilNEA:
Thanks @davidbudnick — solid groundwork.

Follow-ups for your FYR
- Re-apply stored lighting on device arrival / startup (color is lost on replug).
- A wired non-keyboard USB device (enumerates as Unknown) currently gets an inert Lighting tab — the real fix is a 0x8080 capability probe.
- Name the reverse-engineered 0x8080 frame bytes in set_keyboard_color.
- Tray menu silently caps at 8 device rows.

davidbudnick:
> Thanks @davidbudnick — solid groundwork.
> 
> Follow-ups for your FYR
> 
> * Re-apply stored lighting on device arrival / startup (color is lost on replug).
> * A wired non-keyboard USB device (enumerates as Unknown) currently gets an inert Lighting tab — the real fix is a 0x8080 capability probe.
> * Name the reverse-engineered 0x8080 frame bytes in set_keyboard_color.
> * Tray menu silently caps at 8 device rows.

@AprilNEA 👋🏻 

I will work to move forward with your suggestions, I have found there are ~53 devices missing from the assets you have hosted within `openlogi.org` created a GH issue and repo which holds the remaining missing devices. #116.

Looking forward on them being added to the assets bucket so I can move forward with the new pull request!

PR #57, opened 2026-06-01
feat(i18n): add zh-TW (Traditional Chinese, Taiwan) locale

PeterDaveHello:
This pull request adds support for Traditional Chinese (Taiwan) as a separate locale (`zh-TW`), improves locale detection logic for Chinese variants, and updates translation files and documentation accordingly.

The changes ensure users can explicitly select between Simplified Chinese (`zh-CN`), Traditional Chinese for Hong Kong (`zh-HK`), and Traditional Chinese for Taiwan (`zh-TW`), and that the correct translations are loaded based on their system or user-selected locale.

### DIFF
--- a/crates/openlogi-gui/src/i18n.rs
+++ b/crates/openlogi-gui/src/i18n.rs
@@ -4,8 +4,8 @@
 //! at compile time by the `rust_i18n::i18n!` macro in `main.rs`. Call sites use
 //! the [`tr!`](crate::tr) helper (or `rust_i18n::t!`) with the **English string
 //! as the key** — a missing entry falls back to that English text, so the file
-//! carries only the translated `ja` / `ru` / `zh-CN` / `zh-HK` columns;
-//! English is the key itself.
+//! carries only the translated `ja` / `ru` / `zh-CN` / `zh-HK` / `zh-TW`
+//! columns; English is the key itself.
 //!
 //! The current locale is a process-global atomic inside `rust_i18n`. Setting it
 //! re-localizes both our own call sites *and* gpui-component's built-in widget
@@ -19,16 +19,17 @@ use openlogi_core::config::AppSettings;
 /// Locales the GUI ships, as `(code, native name)`. The codes match the
 /// sub-keys in `locales/app.yml`; `en` / `zh-CN` / `zh-HK` also match
 /// gpui-component's bundled `ui.yml`, so choosing one localizes the framework's
-/// own widgets too. `ja` and `ru` are *not* in `ui.yml`, so under Japanese or
-/// Russian our app strings localize but the framework's built-in widget strings
-/// fall back to English.
+/// own widgets too. `ja`, `ru`, and `zh-TW` are *not* in `ui.yml`, so under
+/// those locales our app strings localize but the framework's built-in widget
+/// strings fall back to English.
 /// Order here is the order shown in the Settings picker (after "Follow system").
 pub const SUPPORTED: &[(&str, &str)] = &[
     ("en", "English"),
     ("ja", "日本語"),
     ("ru", "Русский"),
     ("zh-CN", "简体中文"),
-    ("zh-HK", "繁體中文"),
+    ("zh-HK", "繁體中文（香港）"),
+    ("zh-TW", "正體中文（臺灣）"),
 ];
 
 /// Resolve the locale to apply, preferring an explicit stored `setting`, then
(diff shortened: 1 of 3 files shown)


crates/openlogi-gui/src/i18n.rs
```diff
@@ -17,18 +17,19 @@
 use openlogi_core::config::AppSettings;
 
 /// Locales the GUI ships, as `(code, native name)`. The codes match the
-/// sub-keys in `locales/app.yml`; `en` / `zh-CN` / `zh-HK` also match
-/// gpui-component's bundled `ui.yml`, so choosing one localizes the framework's
-/// own widgets too. `ja` and `ru` are *not* in `ui.yml`, so under Japanese or
-/// Russian our app strings localize but the framework's built-in widget strings
-/// fall back to English.
+/// sub-keys in `locales/app.yml`; `en` / `zh-CN` / `zh-HK` / `zh-TW` also
+/// match gpui-component's bundled `ui.yml`, so choosing one localizes the
```
AprilNEA: The current pinned gpui-component ui.yml appears to ship en/zh-CN/zh-HK, but not zh-TW. So choosing zh-TW localizes OpenLogi app strings, but gpui-component built-in strings will still fall back to English. Please update this comment to avoid claiming zh-TW matches gpui-component unless we also add/upgrade gpui-component zh-TW support.

crates/openlogi-gui/src/i18n.rs
```diff
@@ -60,11 +68,22 @@ fn match_supported(code: &str) -> Option<&'static str> {
         Some("ja") => Some("ja"),
         Some("ru") => Some("ru"),
         Some("zh") => {
-            let traditional = matches!(
-                subtags.find(|t| matches!(*t, "hans" | "hant" | "tw" | "hk" | "mo")),
-                Some("hant" | "tw" | "hk" | "mo")
-            );
-            Some(if traditional { "zh-HK" } else { "zh-CN" })
+            let rest: Vec<&str> = subtags.collect();
+            let has_hans = rest.contains(&"hans");
```
AprilNEA: This is likely what clippy is objecting to: we allocate a Vec and then scan it multiple times with contains(). Could we parse script/region in one pass and then match the resolution priority explicitly? For example:

```rust
let mut script = None;
let mut region = None;

for subtag in subtags {
    match subtag {
        "hans" | "hant" => script = Some(subtag),
        "cn" | "sg" | "tw" | "hk" | "mo" => region = Some(subtag),
        _ => {}
    }
}

match (script, region) {
    (Some("hans"), _) => Some("zh-CN"),
    (_, Some("hk" | "mo")) => Some("zh-HK"),
    (_, Some("tw")) | (Some("hant"), _) => Some("zh-TW"),
    _ => Some("zh-CN"),
}
```

crates/openlogi-gui/src/i18n.rs
```diff
@@ -93,8 +112,12 @@ mod tests {
...
         assert_eq!(match_supported("zh-Hans-CN"), Some("zh-CN"));
         assert_eq!(match_supported("zh-CN"), Some("zh-CN"));
-        assert_eq!(match_supported("zh-Hant-TW"), Some("zh-HK"));
+        assert_eq!(match_supported("zh-Hans-HK"), Some("zh-CN"));
+        assert_eq!(match_supported("zh-Hant-TW"), Some("zh-TW"));
+        assert_eq!(match_supported("zh-TW"), Some("zh-TW"));
+        assert_eq!(match_supported("zh-Hant"), Some("zh-TW"));
         assert_eq!(match_supported("zh-HK"), Some("zh-HK"));
+        assert_eq!(match_supported("zh-Hant-HK"), Some("zh-HK"));
         assert_eq!(match_supported("ja"), Some("ja"));
```
AprilNEA, line 122: Could you also add a small zh-TW check to locale_file_resolves_keys()? The new column is otherwise mostly untested beyond locale mapping, so a missing/typoed zh-TW app.yml entry would silently fall back to English.

AprilNEA (changes requested):
Thanks for adding zh-TW. I think this needs a small follow-up before merging: clippy is currently failing, and there are a couple of i18n edge cases/comments to fix inline.

crates/openlogi-gui/src/i18n.rs
PeterDaveHello: Sure.

AprilNEA (approved):

PR #77, opened 2026-06-02
feat: SmartShift 0x2110 support (MX Master 2S) + diag smartshift --sensitivity

laofun:
## Summary

Adds support for the original `0x2110` SmartShift feature so devices that
predate `0x2111 SmartShiftWheelEnhanced` (e.g. the MX Master 2S) can toggle
the ratchet and adjust sensitivity, plus a `diag smartshift --sensitivity`
flag.

## Changes

- **openlogi-hid:** a private `SmartShift` enum wraps whichever feature a
  device exposes — `0x2111` Enhanced (MX Master 3/3S) or the original
  `0x2110`. It tries `0x2111` first and falls back to `0x2110` **only** on a
  missing-`0x2111` error, so transport/protocol failures are never masked by
  a second open attempt. `toggle_smartshift` and `get_smartshift_status` now
  work on both. New `set_smartshift_sensitivity` writes the auto-disengage
  threshold while preserving the current Free/Ratchet mode.
- **openlogi-cli:** `diag smartshift --sensitivity N` (1–255; 255 = permanent
  ratchet) sets sensitivity without toggling, rejects `0` (a device no-op),
  conflicts with `--leave-flipped`, and verifies the read-back value + mode.

## Testing

- Unit tests cover byte-encoding parity (Free/Freespin = 1, Ratchet = 2) and
  the fallback decision.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo test --workspace` all pass.
- Device writes smoke-tested on a physical MX Master 2S: toggle round-trips
  Free↔Ratchet and `--sensitivity` writes verify via read-back.

### DIFF
--- a/crates/openlogi-cli/src/cmd/diag/smartshift.rs
+++ b/crates/openlogi-cli/src/cmd/diag/smartshift.rs
@@ -1,5 +1,7 @@
 //! `openlogi diag smartshift` — SmartShift toggle round-trip.
 
+use std::num::NonZeroU8;
+
 use anyhow::{Context, Result};
 use clap::Args;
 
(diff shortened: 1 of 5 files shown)


devIlluSioN:
SmartShift switch doesn’t work with the 3S either. Is it the same thing?

PR #78, opened 2026-06-02
Use device-reported DPI capabilities

### LINKED ISSUE #50: Query real device DPI range instead of hardcoding 6400
## Context

OpenLogi currently caps the GUI DPI slider at `6400` via a local constant in `crates/openlogi-gui/src/components/dpi_panel.rs`:

```rust
const MIN_DPI: f32 = 200.;
const MAX_DPI: f32 = 6400.;
const STEP_DPI: f32 = 50.;
```

Git history shows this value was introduced during the early GUI DPI slider / preview-dot work as a UI tuning range, not as a device-derived limit. It does not reflect the true capabilities of newer devices such as MX Master 3S / MX Master 4, which can support 8K DPI.

## Findings

- The HID write path itself is not limited to 6400; it writes DPI as `u16`.
- Logitech Options+ exposes an **“Extend sensor range to 8K DPI”** toggle for supported devices.
  - With the toggle off, devices commonly expose a max around 4000 DPI.
  - With the toggle on, supported devices expose up to 8000 DPI.
- Solaar handles this by querying the device-reported DPI list from HID++ instead of hardcoding a max.
- The relevant standard feature is HID++ `0x2201 AdjustableDpi`.
  - OpenLogi currently implements `getSensorCount`, `getSensorDpi`, and `setSensorDpi`.
  - It does **not** yet implement `getSensorDpiList`.
- Some devices may use HID++ `0x2202 ExtendedAdjustableDpi`; OpenLogi already knows about this feature ID in inventory, but the DPI write path currently uses only `0x2201`.

### 1. Move/add protocol-level DPI list support in `openlogi-hidpp`

Add a typed Adjustable DPI feature wrapper to `openlogi-hidpp`, or move the current local wrapper from `openlogi-hid` into `openlogi-hidpp`:

```rust
pub async fn get_sensor_dpi_list(&self, sensor_index: u8) -> Result<Vec<u16>, Hidpp20Error>
```

This method should implement HID++ `0x2201` function 1 (`getSensorDpiList`) and parse both:

- explicit 2-byte DPI entries
- range encoding entries using the `0xE000 | step` format followed by the range end

### 2. Expose an OpenLogi-level API from `openlogi-hid`

Add a route-based public API for GUI/CLI callers:

```rust
pub async fn get_dpi_list(route: &DeviceRoute) -> Result<Vec<u16>, WriteError>
```

or a richer capability type:

```rust
pub struct DpiCapabilities {
    pub values: Vec<u16>,
}
```

`openlogi-hid` should remain responsible for opening channels, resolving routes, and mapping HID++ errors into OpenLogi errors.

### 3. Use the real range in the GUI

Update the DPI panel so the slider range comes from the active device's reported DPI list:

- MX Master 3-style devices should naturally cap around 4000.
- MX Master 3S / MX Master 4 with 8K mode enabled should naturally cap at 8000.
- If querying the list fails, use a conservative fallback and log the failure.

### 4. Document the Options+ 8K toggle behavior

Until OpenLogi can toggle the hidden/internal 8K range flag itself, the UI/docs should explain:

- If a supported 8K mouse only reports up to 4000 DPI, enable **“Extend sensor range to 8K DPI”** once in Logi Options+.
- OpenLogi should then read the extended list from the device.

## Open questions

- Can the Options+ 8K toggle be safely controlled through a known hidden/internal HID++ feature such as `0x1806 CONFIG_DEVICE_PROPS`?
- Should OpenLogi eventually support `0x2202 ExtendedAdjustableDpi` for devices that expose only the extended feature?
- What fallback range should the GUI use when DPI list querying is unavailable?

## Acceptance criteria

- The GUI no longer hardcodes `6400` as the DPI maximum for all devices.
- OpenLogi can read and expose the active device's supported DPI values/range.
- Supported 8K devices can display/use up to 8000 DPI when their firmware reports that range.
- Existing DPI read/write behavior remains intact.

AprilNEA:
## Summary
- implement HID++ AdjustableDpi 0x2201 in openlogi-hidpp, including DPI list parsing and validation
- expose DpiInfo/DpiCapabilities from openlogi-hid and use real supported DPI values in CLI diagnostics
- wire the GUI DPI panel and preset cycle path to lazy per-device DPI capability discovery instead of hardcoded 6400 limits

## Validation
- cargo fmt --check
- cargo check -p openlogi-hid -p openlogi-cli -p openlogi-gui
- cargo test -p openlogi-hidpp adjustable_dpi
- cargo test -p openlogi-hid
- cargo clippy -p openlogi-hidpp --all-targets -- -D warnings
- cargo clippy -p openlogi-hid --all-targets -- -D warnings
- cargo clippy -p openlogi-cli --all-targets -- -D warnings
- cargo clippy -p openlogi-gui --all-targets -- -D warnings

Closes #50


AprilNEA:
Rebased onto master and pushed directly: commits ce5e8d7 (feat) + 33dbb71 (review-fix follow-up). The earlier fc67169 (GitHub App token inputs) was already on master, so the rebase dropped it. CI and Pullfrog were green on the final revision. Closing — the changes are merged into master.

PR #88, opened 2026-06-02
Polish empty state UI and improve user hints for UX/usability

markosnarinian:
# Improvements
1. **Remove text_muted color**: Makes the "Plug in or pair a device…" hint readable instead of dimmed. Users skip dimmed text.
2. **Center text**: Aligns instruction and warning lines with the heading for visual symmetry.
3. **Improve whitespace**: Change whitespace between in the page that shows when no devices are connected for improved aesthetics.
4. **Instruct users to pair in their computer's settings for direct Bluetooth connections**: Adds a hint to the empty state for users with Bluetooth-only devices, preventing confusion when the device doesn't appear automatically.
5. **Quieter "Add device" button**: Reduces vertical padding so it doesn't dominate the empty state and prevents it from stealing attention from the "pair in your computer's settings" instruction to the user.
6. **Pluralize "No device connected"**: Corrects grammar.

# Screenshots

## Now
image

## Before
image

### DIFF
--- a/crates/openlogi-gui/locales/app.yml
+++ b/crates/openlogi-gui/locales/app.yml
@@ -33,12 +33,12 @@ _version: 2
   zh-HK: 結束 OpenLogi
   zh-TW: 結束 OpenLogi
 
-"Plug in or pair a supported Logitech device — it'll show up here automatically.":
-  ja: 対応する Logitech デバイスを接続またはペアリングすると、ここに自動的に表示されます。
-  ru: "Подключите или сопрягите поддерживаемое устройство Logitech — оно автоматически появится здесь."
-  zh-CN: 接入或配对受支持的 Logitech 设备，它会自动显示在这里。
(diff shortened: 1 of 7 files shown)


AprilNEA:
This direction makes sense to me. The “Add Device” flow is really for receiver-based pairing, while direct Bluetooth devices need to be paired through the OS Bluetooth settings.
One small UX concern is that users may still see the Add Device button next to this hint and assume it also handles direct Bluetooth pairing. As a follow-up, it might be clearer to split the two paths:
- Add Device for Bolt / Unifying receiver pairing.
- Open Bluetooth Settings for direct Bluetooth devices.
Ideally, the Bluetooth path could open the system Bluetooth settings directly instead of only mentioning it in text. This PR is fine as a wording/UI improvement, but that might be a better next step for usability.

AprilNEA (approved):

PR #97, opened 2026-06-03
fix(gui): stop idle CPU drain from the connected status dot

davidbudnick:
## Problem

While the app sits idle, it continuously pins a CPU core (and drains battery) **whenever a device is connected**. This is a CPU drain, not a memory leak — RSS stays flat.

## Root cause

`status_dot()` in `crates/openlogi-gui/src/app.rs` rendered a **Connected** device's indicator with an infinite animation:

```rust
.with_animation(("gallery-status-breath", idx), Animation::new(2200ms).repeat(), …)
```

In GPUI's immediate-mode renderer a `.repeat()` animation forces a full re-render on **every display frame for as long as the element is visible**, so the render loop never goes idle. (CPU is low for the first few seconds after launch — the device is still enumerating → `Offline` → static dot — then jumps the moment it shows `Connected`.)

## Fix

Replace the infinite "breathing" animation with a static glow on the connected dot. The per-card `idx`, which existed only to key the animation, is no longer needed and is threaded out of `status_dot`/`device_card`.

## Before / after

Measured with a debug build, an MX Master 3S connected, and the main window frontmost.

| metric | before (`.repeat()` animation) | after (static glow) |
| --- | --- | --- |
| idle CPU (steady) | **66–70%**, never drops | **~2%** |
| GPU work (Metal submissions / 3 s) | **126** | **18** |
| rendering | every frame (continuous) | on events only |
| RSS | ~116 MB | ~116 MB |

The "before" CPU samples were a flat `66 / 68 / 70 / 68 / 67 / 67` — never dropping, the signature of per-frame rendering. The debug build inflates the absolute percentage; the point is eliminating the continuous render (a pinned core → idle).

## Test plan

- `cargo clippy -p openlogi-gui` — clean
- `cargo fmt -p openlogi-gui --check` — clean
- `cargo test -p openlogi-gui` — 11 passed
- Connect a device: the status dot shows a static glow and idle CPU returns to baseline.

### DIFF
--- a/crates/openlogi-gui/src/app.rs
+++ b/crates/openlogi-gui/src/app.rs
@@ -1,10 +1,8 @@
-use std::time::Duration;
-
 use gpui::{
-    Animation, AnimationExt as _, AnyElement, AppContext as _, BorrowAppContext as _, BoxShadow,
-    Context, Div, Entity, FontWeight, InteractiveElement, IntoElement, ParentElement, Render,
-    SharedString, StatefulInteractiveElement as _, Styled, Subscription, Window, div, ease_in_out,
-    img, point, prelude::FluentBuilder as _, px, relative, rgb,
+    AnyElement, AppContext as _, BorrowAppContext as _, BoxShadow, Context, Div, Entity,
+    FontWeight, InteractiveElement, IntoElement, ParentElement, Render, SharedString,
+    StatefulInteractiveElement as _, Styled, Subscription, Window, div, img, point,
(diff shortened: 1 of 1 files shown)


AprilNEA:
Thanks for catching this, @davidbudnick — genuinely great find.

Could you rebase onto master and re-point the fix at app.rs::status_dot?

davidbudnick:
@AprilNEA All rebased!

PR #114, opened 2026-06-03
ci: split out Nix CI; rename CI workflow to Rust CI

AprilNEA:
Splits CI into two clearly-named workflows.

## `nix.yml` (new — "Nix CI")
Builds the flake's `.#openlogi` package on `macos-latest` (`nix build .#openlogi -L`).

- **Why:** the Nix package's `cargoHash` goes stale whenever `Cargo.lock` changes (dep bump, release version bump, or a gpui pin bump) — a stale hash fails the build. This catches it automatically instead of surprising the next release/nixpkgs bump. (We just hit exactly this: 0.4.0 `sha256-bY/…` → 0.4.1 `sha256-xtO1…`.)
- **Gated** to files that actually affect the Nix build (`Cargo.lock`, `flake.*`, `nix/**`, the committed `.icns`, the workflow itself), so it doesn't run on every Rust change — Rust CI covers those.
- darwin-only package; arm64 alone catches cargoHash drift (platform-independent).

## `ci.yml` → display name "Rust CI"
Only the workflow's `name:` changed (CI → Rust CI). **Job names are unchanged**, so any required status checks are unaffected; the file and its triggers stay the same.

Uses `DeterminateSystems/nix-installer-action@v22` (flakes on by default) + `actions/checkout@v6`.

## cargoHash maintenance (`nix/refresh-cargo-hash.sh`)

The flake builds the working tree (local `src`), so `nix-update` can't refresh
the `cargoHash` — it tracks a remote version a local src doesn't have. Added
`nix/refresh-cargo-hash.sh`: it sets a fake hash, builds, and writes back the
value Nix reports (the vendor FOD fails before gpui compiles, so it's fast).
The `nix.yml` build above is the fail-closed guard; this script is the one-command fix.

Verified locally: `nix build .#openlogi` reports the committed cargoHash, and the
script round-trips to `cargoHash already up to date`.

Supersedes #133 (the remove-the-flake alternative) — keeping the flake with low-maintenance cargoHash instead of dropping it.


AprilNEA:
Superseded by #262, which removes the self-hosted flake entirely rather than keeping it behind a CI guard.

This workflow was meant to *catch* the `cargoHash` drift automatically; #262 removes that drift at the root by dropping the flake (OpenLogi is already covered by nixpkgs NixOS/nixpkgs#527640 and homebrew-cask). The `bindgenHook` fix and the version/hash bump go away with the package. Closing.

Issue #95, opened 2026-06-02
Linux event-hook support


recchia:
## Summary

Track first-class **Linux support for the OS-level mouse event hook** (`openlogi-hook`). Today the crate compiles on Linux but `Hook::start` returns `HookError::Unsupported`, so button remapping, event suppression, and per-app profile switching are unavailable. Device-side features (discovery, DPI, SmartShift, battery) ride on `async-hid`/`hidpp` and may already work on Linux with minor adjustments — but that needs verification before any hook work.

I'd like to take this on, but I'm opening a tracking issue first to align on direction (the hook API is still marked unstable) and to avoid colliding with the in-flight Windows work in #93 / #73.

## Current state (as of this writing)

- `crates/openlogi-hook/src/lib.rs` exposes a clean, platform-agnostic contract: `Hook::start/stop`, `MouseEvent::{Button, Scroll}`, `EventDisposition::{PassThrough, Suppress}`, `has_accessibility()`, `prompt_accessibility()`, and `frontmost_bundle_id() -> Option<String>`.
- macOS is implemented in `macos.rs` via `CGEventTap` on a dedicated `CFRunLoop` thread; frontmost app comes from `NSWorkspace`.
- Non-macOS targets make `Hook` uninhabited (`never: Infallible`) and return `Unsupported`.
- Per-app profile overlays poll `frontmost_bundle_id()` (~1 Hz in `openlogi-gui::app_watcher`) keyed on macOS bundle IDs.

## The core constraint: display server, not effort

The hard part of Linux isn't writing Rust — it's that the two display servers expose very different capabilities:

- **X11**: tractable. `XInput2` / `XGrabButton` can capture specific buttons and decide pass-through vs. suppress, mapping cleanly to the `CGEventTap` model. Active window via `_NET_ACTIVE_WINDOW` + `WM_CLASS` gives a `frontmost_bundle_id()` equivalent.
- **Wayland**: by design forbids global input interception and focused-window introspection from arbitrary clients. There is no userspace `CGEventTap` equivalent. Realistic options are (a) `evdev` grab (`EVIOCGRAB`) + `uinput` re-injection — display-server-agnostic but requires `input`-group/udev permissions, grabs the whole device, and is fragile on the re-injection path; or (b) emerging portals (`libei`). Neither gives a clean focused-app id on Wayland (would need a compositor-specific path, e.g. a GNOME Shell extension).

### Phase 0 — Verify the HID/device layer on Linux
[...]

PR #119, opened 2026-06-04
feat(hook): implement Linux evdev+uinput mouse hook

cserby:
## Summary

- Enumerates mouse devices via `evdev::enumerate()` (devices with `BTN_LEFT`), grabs each exclusively, and re-injects pass-through events via a paired `uinput` virtual device. `Suppress` events are consumed without reaching the desktop.
- Supports fractional scroll via `REL_WHEEL_HI_RES` / `REL_HWHEEL_HI_RES` (÷120 per the kernel convention), with automatic fallback to integer `REL_WHEEL` / `REL_HWHEEL` on devices that don't expose hi-res axes.
- `HookError` gains `NoDeviceFound` and `Linux(io::Error)` variants (Linux-only). The `Unsupported` stub is now Windows-only.
- Shutdown uses a per-thread pipe to interrupt blocking `poll()` cleanly.

## Known limitations

- **No hotplug** — devices are enumerated at `Hook::start` only; a mouse plugged in afterwards won't be hooked until restart.
- **No per-app profiles** — `frontmost_bundle_id()` always returns `None` on Linux (the roadmap item P1.4).

## Manual testing (Linux)

Build and run the included smoke-test example with `sudo` (requires access to `/dev/input` and `/dev/uinput`):

```sh
cargo build --example print_events -p openlogi-hook
sudo ./target/debug/examples/print_events
```

Move the mouse, scroll, and click buttons — events should print to stdout and still reach the desktop normally. Ctrl-C shuts down cleanly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -654,6 +654,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "block"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1403,6 +1415,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2931af7e13dc045d8e9d26afccc6fa115d64e115c9c84b1166288b46f6782c2"
 
 [[package]]
+name = "ctrlc"
+version = "3.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0b1fab2ae45819af2d0731d60f2afe17227ebb1a1538a236da84c93e9a60162"
+dependencies = [
+ "dispatch2",
+ "nix 0.31.3",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "darling"
 version = "0.20.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


recchia:
Thanks @cserby — this is great, and going straight to `evdev`+`uinput` instead of X11 was the right call: it covers Wayland sessions without compositor-specific code. I'm picking this up from two angles so nothing gets duplicated.

**1) Testing #119 on real Linux hardware.** 
Environment: Ubuntu 24.04, GNOME on Wayland, MX Master 3S and MX Anywhere 3.

Specifically I'll try to confirm/repro the items already flagged in review, since they need real hardware:
- The **Wayland path end-to-end** (exclusive grab + `Suppress` + re-inject under an actual Wayland session, not just X11).
- The **100% CPU spin on device unplug** (`POLLHUP`/`POLLERR` not handled in `wait_readable`) — easy to trigger with a wireless receiver that sleeps/disconnects.
- The **silent-failure path** when `/dev/uinput` is inaccessible or `EVIOCGRAB` fails — does a `Suppress` callback then pass everything through with no error surfaced?
- The **low-res scroll companion leak** when hi-res axes are active.
- A clean **cross-platform `cargo build`** of the workspace from a Linux host.

I'll report back with logs and the exact udev/permission setup needed to run without `sudo`.

**2) Claiming Phase 0 — HID/device-layer verification on Linux** (from the tracking issue #95).
This is out of scope for this PR (it's `openlogi-hid`, not the hook crate), and it's still open, so I'm taking it to avoid collisions:
- `async-hid`/`hidpp` enumeration of the Bolt receiver + BT-direct/wired devices over `hidraw`.
- DPI (`0x2201`), SmartShift (`0x2111`/`0x2110`), and battery read/write.
- Whether Linux needs report-size handling analogous to the Windows short/long-report channel in #73.
- Documented udev rules for non-root `hidraw` access to Logitech receivers.

I'll open it as a separate PR so it doesn't entangle with the hook work here. If anyone's already on Phase 0, give me a shout so we don't double up.

PR #131, opened 2026-06-05
Migrate macOS ObjC FFI to objc2, fixing the menu-bar leak (#99)

### LINKED ISSUE #99: Memory continuously increases to over 1GB+
It is unclear whether there is a memory leak or normal usage, as the memory usage ranges from over 100MB to over 1GB during use.

Image

OS: macOS 26.5.1 
OpenLogi Version: v0.4.0
Connection method: Bluetooth

AprilNEA:
## Why

Issue #99 reports memory climbing to 1 GB+. There were two distinct things behind it:

1. **A real leak.** `status_item.rs::nsstring()` returned a `+1`-owned `NSString` (`NSString::alloc(nil).init_str`) that was never released, and it ran on every 2 s tray refresh via `set_title`. That is exactly the ~26k leaked `CFString`s the reporter's `leaks` run found. The root cause isn't one missing `release` — it's the *representation*: `cocoa`/`objc` 0.x model every object as a raw, hand-retained `id`, so retain/release is an unenforced runtime discipline.
2. **Footprint churn.** `enumerate_hidpp_devices()` rebuilt a whole `HidBackend` (a macOS `IOHIDManager` it schedules and, on drop, cancels) on every ~2 s inventory tick — needless create/teardown that kept the process busy and its heap dirty around the clock.

## What this does

**Migrate the macOS Objective-C FFI from `cocoa`/`objc` 0.x to `objc2`** (already in the tree via gpui/async-hid), so the leak class can't be *written*:

- `Retained<T>` owns each AppKit object and releases it once on `Drop`. `nsstring()` and the autorelease-pool workaround are gone — every string is `NSString::from_str(s)` used as a borrowed temporary. A `+1`-with-no-owner is unrepresentable.
- The `OpenLogiMenuTarget` click target is now a `define_class!` subclass with the event channel in a typed `#[ivars]` (no `ClassDecl`/`add_method`, no `OnceLock<MENU_TX>` read in the callbacks).
- `NSMenu`/`NSMenuItem` are `MainThreadOnly` (`!Send`), so the tray's retained state moves from `OnceLock` statics into a main-thread `thread_local` — "main-thread only" is enforced by the type system, not a doc comment.
- `permissions.rs` (`CBCentralManager.authorization`) and the hook's `frontmost_bundle_id` (`NSWorkspace`, which is `AnyThread` so it's sound on the watcher thread) also move to `objc2`. `cocoa`/`objc` 0.x leave both crates' **direct** dependencies (they remain in `Cargo.lock` only transitively via gpui).

**Reuse one HID backend** in a `LazyLock` instead of rebuilding it per poll — the usage async-hid intends; it also keeps the device set warm between polls.

Plus a scoped `crates/openlogi-gui/src/platform/CLAUDE.md` documenting the conventions.

## What deliberately stays

The CGEventTap stays on `core-graphics`/CF. `objc2-core-graphics` *does* expose the tap API, but the Accessibility-revoke freeze-hazard run-loop machinery is load-bearing — out of scope here, byte-for-byte unchanged. Only the `NSWorkspace` read in that file moved.

## Verification

- `cargo fmt --check`, full-workspace `cargo clippy --all-targets -- -D warnings`, and all tests pass (hook+hid 26, gui 11).
- The `zed`/gpui-component git pins in `Cargo.lock` are unchanged — the lock diff is only the new `objc2-*` crates.io packages.
- **Runtime, against a live MX Master 4:** the dev bundle launches with no crash/panic; the `computermouse.fill` status item renders; the menu builds correctly (AX dump: device row `MX Master 4 · NN%`, separator, `Open OpenLogi`, `Quit OpenLogi`, spare rows hidden); and clicking **Quit OpenLogi** exits the app — confirming the `define_class!` action callback → ivar channel → `cx.quit()` chain fires end to end.

Closes #99.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4278,6 +4278,39 @@ version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d49e936b501e5c5bf01fda3a9452ff86dc3ea98ad5f283e1455153142d97518c"
 dependencies = [
+ "bitflags 2.11.1",
+ "block2",
+ "libc",
+ "objc2",
+ "objc2-cloud-kit",
+ "objc2-core-data",
+ "objc2-core-foundation",
+ "objc2-core-graphics",
+ "objc2-core-image",
+ "objc2-core-text",
+ "objc2-core-video",
+ "objc2-foundation",
+ "objc2-quartz-core",
+]
+
+[[package]]
+name = "objc2-cloud-kit"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "73ad74d880bb43877038da939b7427bba67e9dd42004a18b809ba7d87cee241c"
+dependencies = [
+ "bitflags 2.11.1",
+ "objc2",
+ "objc2-foundation",
+]
+
+[[package]]
+name = "objc2-core-data"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b402a653efbb5e82ce4df10683b6b28027616a2715e90009947d50b8dd298fa"
+dependencies = [
+ "bitflags 2.11.1",
  "objc2",
  "objc2-foundation",
 ]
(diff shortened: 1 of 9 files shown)


AprilNEA:
Addressed both non-blocking questions in 30287ce:

1. **Shared `HID_BACKEND` under concurrent callers** — documented why it's sound: async-hid declares the backend `Send + Sync`, `enumerate` only reads an `IOHIDManagerCopyDevices` snapshot, and sharing one long-lived `IOHIDManager` across threads is the same model hidapi uses. No behaviour change.
2. **Tray mutators with no `MainThreadMarker`** — added `debug_assert_main_thread()` to them. They were already memory-safe off-main (the `!Send` `TrayState` only lives in the main thread's TLS, so an off-main call no-ops rather than touching AppKit), but a future off-main caller is now a loud debug-build failure instead of a silent no-op. Free in release.

Issue #98, opened 2026-06-03
Master3 has detected the device, but all configuration changes aren’t taking effect


latte-z:
<img width="1100" height="778" alt="Image" src="https://github.com/user-attachments/assets/72df05c9-38c7-4dcb-8184-09591a001e5a" />

Master3 has detected the device, but all configuration changes aren’t taking effect. For example, I set the DPI button to play/pause, but it doesn’t work.

The mouse pointer seems to drift when moved. Additionally, there’s no setting for the scroll wheel’s direction.

AprilNEA:
0.5.3 ships a device-classification fix (#127) that restores any config panels hidden by a misdetected device, so it's worth updating and retrying (`brew update && brew upgrade --cask openlogi`).

That said, the specific things you hit look like separate issues, not that bug:
- *"DPI button → play/pause doesn't fire"*: on macOS the DPI button isn't visible to the event tap, tracked in #145.
- *Scroll-wheel direction setting*: not implemented yet, tracked in #126.
- *Pointer drift*: please open a dedicated issue with your model + connection type if it persists on 0.5.3.

After updating to 0.5.3, let us know which of these remain.

latte-z:
> 0.5.3 ships a device-classification fix ([#127](https://github.com/AprilNEA/OpenLogi/issues/127)) that restores any config panels hidden by a misdetected device, so it's worth updating and retrying (`brew update && brew upgrade --cask openlogi`).
> 
> That said, the specific things you hit look like separate issues, not that bug:
> 
> * _"DPI button → play/pause doesn't fire"_: on macOS the DPI button isn't visible to the event tap, tracked in [[Bug]: Broken Remapping DPI Button to Middle Click #145](https://github.com/AprilNEA/OpenLogi/issues/145).
> * _Scroll-wheel direction setting_: not implemented yet, tracked in [[Feature]: Inverted Scrolling Support #126](https://github.com/AprilNEA/OpenLogi/issues/126).
> * _Pointer drift_: please open a dedicated issue with your model + connection type if it persists on 0.5.3.
> 
> After updating to 0.5.3, let us know which of these remain.

Sorry for being late, I just got back from traveling. I tried the new 0.6.3 version. Here’s what I noticed:

[...]

PR #147, opened 2026-06-05
fix: gate device config panels on HID++ capabilities (closes #127)

### LINKED ISSUE #127: [Bug]: master breaks support for MX Anywhere 3S
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

master branch

### Operating system

macOS

### OS version & architecture

macos tahoe

### Device model

mx anywhere 3s

Logi Bolt receiver

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

Image

Bugs and pointer options dissapeared and now only color options appear which are irrelevant to my mouse.

### Steps to reproduce

1. Launch OpenLogi GUI
2. Connect MX Anywhere 3S

### `openlogi list` output

```text

```

### Logs

```shell

```

AprilNEA:
## Problem

On `master`, a **Bluetooth-direct / wired** MX Anywhere 3S / MX Master 3 shows **only a Lighting (color) tab** — the Buttons and Pointer tabs vanish, so there's no way to remap back/forward or anything else (#127).

## Root cause

Two pieces interact:

- `inventory.rs::probe_direct` (the receiver-less path: Bluetooth-direct, USB-C) hard-codes `kind: DeviceKind::Unknown`. There's no Bolt pairing register on this path to supply a real kind, and nothing else filled it in.
- `app.rs::supports_lighting` (added in #29 for wired G-series keyboards) treats **any** `Unknown` + `DeviceRoute::Direct` device as a lighting-capable keyboard.

So a Bluetooth-direct mouse is indistinguishable from a wired keyboard: `is_configurable_pointer(Unknown)` is `false` (no Buttons/Pointer) while `supports_lighting` is `true` (Lighting only). A Bolt-connected unit reports `Mouse(0x02)` from the pairing register and is unaffected — which is why the same model works over the Bolt receiver but breaks over Bluetooth.

## Fix

Read the device's marketing type from HID++ feature **`0x0005` (DeviceTypeAndName)** and use it for the kind:

- Folded into the existing `probe_features` session, so it's **one extra short round-trip** on an already-open device — no new `Device::new` / `enumerate_features` handshake.
- **Direct path:** uses the `0x0005` type directly → a Bluetooth-direct mouse is now `Mouse`, restoring the Buttons + Pointer tabs and dropping the spurious Lighting tab.
- **Bolt path:** keeps the authoritative pairing-register kind, falling back to `0x0005` **only** when that register reads `Unknown` — so currently-correct devices are untouched.

`map_device_type` maps the `0x0005` enum to our `DeviceKind`; `resolve_device_kind` encodes the precedence (primary wins unless `Unknown`, then probe) and is unit-tested.

## Verification

- `cargo test -p openlogi-hid` — 3 new precedence tests pass.
- `cargo clippy -p openlogi-hid --all-targets` clean under `-D warnings`.
- Not yet verified on hardware — @AprilNEA will test on an Anywhere 3S / Master 3 over Bluetooth. The tab gating is deterministic given `kind == Mouse`; this just needs to confirm the device answers `0x0005` at index `0xff` on the direct path.

## Notes / out of scope

- #98 (filed before #29) describes a different mix — the DPI button isn't visible to the macOS `CGEventTap` (it's outside button range 0–4; see `hook_runtime.rs`), plus pointer drift and a scroll-direction request. If that user is on a post-#29 build over Bluetooth this fix restores their tabs, but the DPI-button limitation is separate and not addressed here.
- A residual edge case — a Bolt register misreading kind as `Keyboard(0x01)` rather than `Unknown` — would not be covered, but there's no evidence of it.

Fixes #127

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4664,6 +4664,7 @@ name = "openlogi-hid"
 version = "0.5.1"
 dependencies = [
  "async-hid",
+ "futures-concurrency",
  "futures-lite 2.6.1",
  "num_enum",
  "openlogi-core",
--- a/crates/openlogi-core/src/device.rs
+++ b/crates/openlogi-core/src/device.rs
@@ -60,6 +60,64 @@ impl DeviceKind {
     }
 }
 
+/// What a device can be *configured* to do, derived from the HID++ feature
+/// table it reports (feature `0x0001`). This is the source of truth for which
+/// configuration panels the UI offers — a panel shows iff the device exposes
+/// the feature that drives it. Gating on capability rather than on
+/// [`DeviceKind`] is what keeps a misclassified device from losing its panels
+/// (issue #127): kind is an identity guess, capability is what the firmware
+/// actually announced.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize)]
+pub struct Capabilities {
+    /// Reprogrammable buttons — HID++ `0x1b00`–`0x1b04` (ReprogControls).
+    pub buttons: bool,
+    /// Adjustable pointer resolution — HID++ `0x2201` / `0x2202` (AdjustableDpi).
+    pub pointer: bool,
+    /// Controllable lighting — HID++ `0x8040`/`0x8070`/`0x8071`/`0x8080`/`0x8081`
+    /// (RGB) or `0x1981`–`0x1983`/`0x1990` (backlight/illumination).
+    pub lighting: bool,
+}
+
+impl Capabilities {
+    /// Derive capabilities from the set of HID++ feature IDs a device reports.
+    /// Membership of a driving feature ID flips the corresponding flag.
+    #[must_use]
+    pub fn from_feature_ids(ids: &[u16]) -> Self {
+        const BUTTONS: [u16; 5] = [0x1b00, 0x1b01, 0x1b02, 0x1b03, 0x1b04];
+        const POINTER: [u16; 2] = [0x2201, 0x2202];
+        const LIGHTING: [u16; 9] = [
+            0x8040, 0x8070, 0x8071, 0x8080, 0x8081, 0x1981, 0x1982, 0x1983, 0x1990,
+        ];
+        let has = |family: &[u16]| ids.iter().any(|id| family.contains(id));
+        Self {
+            buttons: has(&BUTTONS),
+            pointer: has(&POINTER),
+            lighting: has(&LIGHTING),
+        }
+    }
+
+    /// Best-effort capabilities for a device we could not probe (offline /
+    /// never reached), guessed from its [`DeviceKind`]. Used only as a fallback
+    /// when no measured [`Capabilities`] exist — a sleeping mouse should still
+    /// show its button/pointer panels so its bindings (host-side) stay
+    /// configurable.
+    #[must_use]
+    pub fn presumed_from_kind(kind: DeviceKind) -> Self {
+        match kind {
+            DeviceKind::Mouse | DeviceKind::Trackball => Self {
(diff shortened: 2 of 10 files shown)


AprilNEA:
Pushed a follow-up that strengthens the fix and addresses the review note.

**Why the precedence flipped.** #127's reporter is on a **Logi Bolt receiver**, not Bluetooth-direct. On the Bolt path a device only renders the lighting-only UI when it's classified as a `Keyboard`, which means the receiver's pairing register is *misreporting* the MX Anywhere 3S as a keyboard. The first revision only filled an `Unknown` register from `0x0005`, so it would **not** have fixed that user. 

The device's own HID++ `0x0005` (DeviceTypeAndName) report is authoritative, so `resolve_device_kind` now prefers it for any online device and falls back to the pairing-register kind only when the device is offline (no probe) or `0x0005` returns a type we don't model. This corrects a register that reports the *wrong concrete* kind — covering both the Bolt and the Bluetooth-direct cases in #127.

**Re: the discarded-round-trip note.** The `0x0005` read is unconditional again, but it's no longer discarded — it's now the *primary* kind source. It runs only for online devices that just answered the battery + device-info reads, so it's one cheap short round-trip with a real purpose.

The `bolt_kind` value the register reports is logged on the `paired slot` debug line, so a `RUST_LOG=openlogi_hid=debug` run on the affected device will confirm exactly what the receiver was reporting.

AprilNEA:
## Reframed: capability-driven gating (the root fix)

The earlier commits made the `DeviceKind` proxy more accurate (0x0005, asset cross-check, precedence). That treated the symptom. The disease was gating **capability** decisions on **identity** at all — whether to show a DPI panel has nothing to do with "is this a mouse", only with whether the device exposes `0x2201`/`0x2202`.

This PR now pivots to the essential line:

- **openlogi-core** — `Capabilities { buttons, pointer, lighting }` derived from the device's HID++ feature-ID set (`from_feature_ids`); `presumed_from_kind` is the offline fallback.
- **openlogi-hid** — `probe_features` captures the feature-ID list that `enumerate_features` **already returns** (zero extra round-trips) and returns a `ProbedFeatures` struct. The Bluetooth-direct peripheral discriminator reuses those caps, so the redundant `exposes_peripheral_feature` probe (an extra `Device::new` + root lookups) is **deleted**.
- **openlogi-gui** — `DeviceRecord` carries `Option<Capabilities>`; the snapshot merge carries the last-known set forward so a sleeping device keeps its panels. `tabs_for` shows each panel iff its capability is set. `is_configurable_pointer` and the `supports_lighting` `Unknown+Direct` heuristic are **deleted**.

`kind` is now identity-only (icon/label) — a wrong kind is a cosmetic glitch, never a lost panel.

**What this resolves from the review:**
- The asset-vs-0x0005 precedence concern (review finding #1) is defanged — `kind` no longer gates function, so a registry typo only mislabels an icon.
- The future keyboard-support mirror bug is pre-empted: a keyboard's button config will gate on `0x1b04`, not `kind == Keyboard`.

**No #29 regression:** a wired G-series keyboard exposes `0x8080` (PerKeyLighting) in its feature table — `set_keyboard_color` itself resolves `0x8080` to write — so `caps.lighting` is true for it and the Lighting tab stays.

Verified: `cargo test` (core/hid/gui) + `cargo clippy -D warnings` green; whole workspace builds. Still pending hardware confirmation on the affected Anywhere 3S / Master 3.

PR #24, opened 2026-05-31
feat(openlogi-gui): expand UI to 19 fully-translated locales

davidbudnick:
## Context
- adds 13 UI locales (da de el es fi fr ko nb nl pl pt-BR pt-PT sv), settings picker now 20 languages
- each new locale is full parity with en.yml (all 196 keys), per-locale yml in the crowdin layout
- rebased on current master; dropped the old asset/layout changes master already superseded

## Testing
- cargo test -p openlogi-gui i18n:: green — parity + resolve tests cover all 19 non-en locales

PR #158, opened 2026-06-06
feat(i18n): update italian language

elax46:
Update Italian (it) localization entries to `openlogi-gui/locales/app.yml.` Provides Italian translations for many UI strings across device/pointer/lighting panels, menu bar items, about/help, SmartShift panel, binding popovers, actions and thumb-wheel texts.


AprilNEA:
Thanks for putting this together! It got raced by #24, which landed full Italian (and 18 other locales) via the Crowdin pipeline — and restructured the files from a single app.yml into per-locale files (locales/it.yml), so this diff no longer applies. Translation fixes now flow through Crowdin so they don't get overwritten on the next sync — if you'd like to keep improving the Italian strings there, happy to get you access. Closing as superseded.

PR #162, opened 2026-06-06
chore(nix): bump cargo hash and version

brhutchins:
The nix flake wasn't building, because the cargo hash was out of date.

- Updates `cargoHash` to present state
- Updates `version` to present version
- Adds `cargoDepsName = finalAttrs.pname`
  - Without setting the dependencies name to a static value, the hash gets invalidated on every version bump of local crates. This creates a large maintenance surface for the nix flake, because the hash needs to be updated with every change.
  - By setting it to `pname`, the hash stays valid as long as external dependencies stay consistent, i.e., only actual changes to `Cargo.lock` will invalidate it.
  - If this were in `nixpkgs`, it would make sense to retain the default for `cargoDepsName` and rely on the autobump behaviour; for a local flake, which will always reflect the state of the current branch, minimising hash churn seems like the better option

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -16,7 +16,7 @@
 
 rustPlatform.buildRustPackage (finalAttrs: {
   pname = "openlogi";
-  version = "0.4.0";
+  version = "0.5.3";
 
   # Build from the working tree (target/.git/etc. filtered out). cargo-bundle
   # uses the committed crates/openlogi-gui/icon/AppIcon.icns.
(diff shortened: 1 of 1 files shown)


AprilNEA:
Thanks for digging into this, @brhutchins — and for the `cargoDepsName` pointer, which sent me down a useful rabbit hole.

We've decided to remove the self-hosted flake entirely instead (#262): OpenLogi is already in nixpkgs (NixOS/nixpkgs#527640) and homebrew-cask, so the in-repo flake adds no distribution path those don't already cover, while its single `cargoHash` is a recurring maintenance cost on every `Cargo.lock` change.

One note on `cargoDepsName = pname`, since it's a reasonable thing to reach for: it doesn't actually stop the hash churn here. `fetchCargoVendor` copies `Cargo.lock` into the vendored FOD, and `Cargo.lock` embeds the version of all 10 local `openlogi*` workspace crates — so a release version bump changes the vendored content and therefore the recursive NAR hash. `cargoDepsName` only renames the deps derivation (its store path), which isn't part of the content hash, so the hash still invalidates on every release. (Checked against the pinned nixpkgs `rev 4df1b88`: `build-rust-package/default.nix` wires `cargoDepsName` only to `name`; `fetch-cargo-vendor.nix` sets `outputHashMode = "recursive"`.) It's great for store-path stability and cache reuse — just not for hash stability across version bumps.

Closing in favor of #262, but genuinely appreciate the PR. 🙏

PR #173, opened 2026-06-08
feat(linux): install artifacts and docs

cserby:
## Summary

- **`packaging/linux/udev/70-openlogi.rules`** — `TAG+="uaccess"` for Logitech `hidraw` nodes and `uinput`; no group membership required on systemd-logind systems. Includes non-systemd `GROUP="input"` fallback instructions in comments.
- **`packaging/linux/systemd/openlogi-agent.service`** — packaged user-unit template for `/usr/lib/systemd/user/`; complements the per-user unit written by the `launch_at_login` setting.
- **`packaging/linux/desktop/openlogi.desktop`** — XDG application launcher entry.
- **`packaging/linux/install.sh` / `uninstall.sh`** — POSIX `sh`, `set -eu`; copies binaries and all artifacts to system paths, reloads `udevadm`, and performs best-effort icon cache and desktop database updates.
- **`docs/INSTALL-linux.md`** — full Linux install guide: prerequisites, build from source, udev rules (uaccess + non-systemd fallback), `install.sh` usage, autostart via `systemctl`, verification steps, known limitations.
- **`README.md`** — Linux subsection under `## Install` with the quick-start snippet and a link to the full guide.

## Test plan

- [ ] `desktop-file-validate packaging/linux/desktop/openlogi.desktop` — clean
- [ ] `shellcheck packaging/linux/install.sh packaging/linux/uninstall.sh` — clean (requires shellcheck)
- [ ] `udevadm verify packaging/linux/udev/70-openlogi.rules` — clean (requires udevadm ≥ 252)
- [ ] `bash -n packaging/linux/install.sh && bash -n packaging/linux/uninstall.sh` — no syntax errors
- [ ] Full `install.sh` run after `cargo build --release` — binaries, rules, unit, desktop, icon all land in expected paths
- [ ] `udevadm trigger` after install — `/dev/uinput` and `/dev/hidraw*` become accessible without `sudo`
- [ ] `uninstall.sh` — all installed files removed, udevadm reloaded


recchia:
Nice to see the packaging come together — and thanks for carrying the Bluetooth/uhid matcher forward into the comment block; generalizing it to *:046D:* is a good call.
One thing I'd flag: this adds packaging/linux/udev/70-openlogi.rules as a new file, but udev/70-openlogi.rules already exists in master and neither this PR nor #179 removes it. If both land as-is, the tree ends up with two copies of the rules that will drift apart on the next fix. Suggest making this a git mv of the existing file into packaging/linux/udev/ so there's a single source of truth.

cserby:
> Nice to see the packaging come together — and thanks for carrying the Bluetooth/uhid matcher forward into the comment block; generalizing it to _:046D:_ is a good call. One thing I'd flag: this adds packaging/linux/udev/70-openlogi.rules as a new file, but udev/70-openlogi.rules already exists in master and neither this PR nor #179 removes it. If both land as-is, the tree ends up with two copies of the rules that will drift apart on the next fix. Suggest making this a git mv of the existing file into packaging/linux/udev/ so there's a single source of truth.

Thank you @recchia for pointing this out, I was not aware of your work in the area being merged upstream.

recchia:
Confirmed on the new push — single source of truth restored, and the rewritten rules look good (generalizing the Bluetooth matcher to *:046D:* covers more transports than my original). One small thing from my hardware testing that got lost in the rewrite, in case it's worth a line in INSTALL-linux.md: on my setup (Bolt receiver and Bluetooth-direct), already-connected devices needed a replug/power-cycle after installing the rules before the uaccess ACL applied — udevadm trigger alone didn't re-grant it on the existing hidraw nodes. Since install.sh will mostly run with the device already connected, a one-line note could save some "rules don't work" reports. Resolving this thread — thanks for the quick turnaround on the relocation.

PR #174, opened 2026-06-08
Integrate Crowdin localization workflow

AprilNEA:
## Summary
- split GUI translations from a single rust-i18n v2 app.yml into per-locale v1 YAML files
- add Crowdin configuration using en.yml as the source locale and locale-specific downloads
- add Crowdin CLI to devenv with upload/download tasks and document them

## Verification
- cargo test -p openlogi-gui i18n
- devenv tasks list | rg "i18n|openlogi:"
- pre-push hooks: cargo fmt, cargo clippy


AprilNEA:
Addressed Greptile review feedback in 39a0a51:

- Fixed the DPI preset add chip to translate the existing `Add` key instead of looking up `+ Add`.
- Completed missing locale keys for `it`, `ru`, and `zh-TW`; all locale files now match `en.yml` key coverage.
- Added a test that asserts every locale file has the same keys as `en.yml`.
- Moved language-switch window refresh/menu rebuild into `AppState::set_language` so future callers cannot forget it.

Verification:
- `cargo test -p openlogi-gui i18n`
- pre-push hooks including `cargo clippy`

AprilNEA:
Rebuilt the Crowdin/i18n changes on top of the latest master and force-pushed the updated feat/i18n branch. The PR is now mergeable; conflicts should be resolved.

PR #184, opened 2026-06-08
Fix macOS volume and media key posting

jonstuebe:
## Summary
- Post macOS volume and media controls as NSSystemDefined subtype 8 NX events instead of standard keyboard events
- Wire Volume Up/Down/Mute plus Play/Pause/Next/Previous through the shared media-key helper
- Add macOS objc2 AppKit/CoreGraphics bindings needed to create and post NSEvents

## Validation
- cargo check -p openlogi-core
- cargo fmt --check

Fixes mouse-button bindings for Volume Up/Down/Mute on macOS, where kVK_Volume* keyboard events are ignored by the system volume handler.


rwkyyy:
@AprilNEA can we get this merged please?

PR #66, opened 2026-06-01
Fix security scan findings

jasoncantor:
## Summary

Fixes six security findings from the repository-wide Codex Security scan:

- pins the release packaging `cargo-bundle` install
- requires minisign verification for updater artifacts
- signs and publishes updater DMG signatures in the release workflow
- hardens GUI and CLI asset sync paths and checksum enforcement
- binds direct HID routes to DeviceInformation identity, not only VID/PID
- bounds Bolt discovery-name parsing to prevent oversized-length panics

## Impact

Release/update flows now fail closed unless updater artifacts are signed and the app is built with the minisign public key. Asset sync rejects path traversal, symlink destinations, and checksum mismatches. Direct device writes require the expected serial/unit identity before a route is accepted.

## Validation

Contributing checklist from `docs/DEVELOPMENT.md`:

- `env CARGO_INCREMENTAL=0 RUSTFLAGS='-D warnings' cargo fmt --all -- --check`
- `env CARGO_INCREMENTAL=0 RUSTFLAGS='-D warnings' cargo clippy --workspace --all-targets -- -D warnings`
- `env CARGO_INCREMENTAL=0 RUSTFLAGS='-D warnings' cargo test --workspace`

Additional focused checks run during the security fix:

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p openlogi-assets safe_component_path`
- `cargo test -p xtask collect_assets`
- `cargo test -p openlogi-hidpp parse_discovery_name`
- `cargo test -p openlogi-hid direct_identity`
- `cargo check -p openlogi-cli`
- `xcodebuild -downloadComponent MetalToolchain`
- `cargo test -p openlogi-gui signed_source`
- `cargo check --workspace`

PR #201, opened 2026-06-10
fix(assets): keep sync writes inside the cache root and verified

AprilNEA:
## Why

`index.json` depot keys, per-depot file names, and `manifest.json` variant names are remote data, but the sync paths joined them raw onto local directories. A tampered asset host (or poisoned cache/manifest) could:

- **write outside the cache root** — a depot key or file name like `../LaunchAgents/x` traverses on join (GUI runtime sync, CLI bundle sync, and the glow cache's depot dir)
- **redirect a write through a planted symlink**
- **plant unverified bytes** — the GUI's fallback fetched files the registry doesn't list, with no hash to check, and a fresh download was never re-verified against its registry `sha256`

## What

- `http::safe_component_path` validates every remote-named component down to a single normal path component (no separators, no `.`/`..`, non-empty) before any join; all sync joins go through it
- the destination is refused if a symlink sits there (`symlink_metadata` check at the single write chokepoint)
- `fetch_entry_if_stale` verifies a fresh download against the registry `sha256` and deletes it on mismatch — nothing unverified survives on disk
- the GUI now **skips** names the registry doesn't list instead of fetching them unverified; with that fallback gone, `fetch_file`/`fetch_file_to_dir` became private to `openlogi-assets`
- the glow cache (`ensure_glow_png` / `read_baked_mask`) gates its depot dir the same way — a write path beyond what #66 covered

## Credit

The path-traversal/symlink/checksum findings and the original patch shape are from #66 by @cantorjf, re-derived on current master.

---
Stacked on #197 (the clippy unblock): merge that first and this diff collapses to the one `fix(assets)` commit.

## Verification

- `cargo test -p openlogi-assets -p openlogi-gui -p openlogi-cli` — all pass (2 new traversal tests)
- `cargo clippy --workspace --all-targets -- -D warnings` clean

### DIFF
--- a/crates/openlogi-assets/src/http.rs
+++ b/crates/openlogi-assets/src/http.rs
@@ -97,8 +97,7 @@ impl AssetClient {
     /// Fetch `<base>/index.json`, write it into `dir`, and return the parsed index.
     pub fn fetch_index_to_dir(&self, dir: &Path) -> Result<Index> {
         let (raw, index) = self.fetch_index_raw()?;
-        let local = dir.join(INDEX_NAME);
-        fs::write(&local, &raw).with_context(|| format!("write {}", local.display()))?;
+        write_replace(&dir.join(INDEX_NAME), &raw)?;
         Ok(index)
     }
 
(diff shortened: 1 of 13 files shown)


AprilNEA:
Both review findings addressed in a292005:

- **glow_path** now routes through the same validated depot dir as the glow IO fns (returns `Option`), so the read-side lookup is inside the boundary too.
- **TOCTOU**: replaced check-then-write with `write_replace` — temp file via `create_new` (O_EXCL never follows a planted link) + `rename`, which *replaces* a symlink at the destination rather than writing through it. Also applied to `index.json`, and it makes sync writes atomic for concurrent readers. Unit-tested incl. the planted-symlink case.

Issue #218, opened 2026-06-12
[Bug]: Device list flaps (appears then vanishes ~2s) + device images never load — Bolt receiver, macOS 27 beta


buliwyf42:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.6

### Operating system

macOS

### OS version & architecture

macOS 27 (beta), Apple Silicon

### Device model

MX Master 4 + MX Master 3S + MX Mechanical Mini, all paired to one Logi Bolt receiver

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [ ] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

## Summary

Two issues on a fresh OpenLogi setup:

1. **The device list flaps** — all paired devices appear for ~1s, then the GUI flips back to "No devices connected", oscillating roughly every 2 seconds. They never stay long enough to configure.
2. **Device images never load** — every device card shows the generic chip/silhouette icon instead of an actual mouse/keyboard render.

## Setup

- OpenLogi 0.6.6 (Homebrew cask `openlogi@latest`)
- macOS 27 **beta**, Apple Silicon
- Logi Bolt receiver (USB, PID 0xC548)
- Paired to the receiver: slot 1 = MX Mechanical Mini (keyboard), slot 3 = MX Master 3S (mouse), slot 4 = MX Master 4 (mouse)

## What I ruled out

- Only ONE `openlogi-agent` process running (no duplicate / lock contention).
- Logi Options+ uninstalled; `LogiPluginService` is NOT running (6×1s process sampling showed it absent every time; no unified-log entries in 3 min). So this is not the documented "both apps compete for HID++" case.
- Input Monitoring granted to OpenLogi; GUI footer shows "Accessibility granted".
[...]

PR #221, opened 2026-06-12
fix(hid): keep last-known-good devices when a receiver read is incomplete (#218)

buliwyf42:
## Summary

Fixes the **device-list flapping** half of #218: on a Logi Bolt receiver the GUI oscillated between showing all paired devices and "No devices connected" roughly every 2 s, so they never stayed long enough to configure. Reported on 0.6.6, macOS 27 beta, Apple Silicon, with an MX Master 4 + MX Master 3S + MX Mechanical Mini on one receiver.

This PR is scoped to the flap only. The **device-images** half of #218 is already addressed by #142 (Settings → Assets runtime sync, no CLI needed), #137 (match every `modelId`), and #140 (retry sync with backoff), so I've left it to those.

## Root cause

The agent's inventory resilience rests on one rule, stated in `watchers/inventory.rs`:

> `enumerate()` returning **`Err`** = "couldn't check" → keep the last good snapshot.
> `enumerate()` returning **`Ok`** = authoritative → forward it (an empty `Ok` is a genuine disconnect).

A transient HID++ register read broke that rule. `count_pairings()` and the per-slot `get_device_pairing_information()` are independent HID++ 1.0 register reads; on the macOS 27-beta IOHID stack (with the brand-new MX Master 4 adding notification traffic) they time out intermittently. A failed slot read was silently skipped, so `probe_one` returned `Ok` with a **truncated** `paired` list — possibly empty — and a whole-receiver probe timeout dropped the receiver entirely. `enumerate()` already *detected* the shortfall (`paired.len() != pairing_count`) but only logged a warning.

That truncated `Ok` reached `Orchestrator::refresh_inventory`, which unconditionally overwrites `InventoryState::Ready`; the IPC `inventory()` poll served the short/empty list and the GUI rendered "No devices connected" until the next good cycle ~2 s later. The GUI's own `INVENTORY_MISS_GRACE = 2` rides out 1–2 transient empties, but a run of ≥3 (plausible on this stack) drops the devices anyway — and the agent-side snapshot flapped regardless.

I also checked the duplicate-agent spawn race from #207's review (the `open -n` path in `ipc_client.rs`). It's bounded — `just_disconnected` skips the spawn gate on the drop tick, `SPAWN_RETRY_PERIOD = 30 s` rate-limits respawns, and the agent's `single_instance::acquire("agent.lock")` makes any duplicate exit — so it can't produce a 2 s flap, and the report's own evidence (one agent process; isolated *cold* agent runs flap) rules it out. This PR doesn't touch `ipc_client.rs`, so it won't conflict with #207.

## The fix

Make `enumerate()` honest about partial reads so the existing keep-last-snapshot machinery engages, instead of papering over it in every consumer.

- **`openlogi-hid/inventory.rs`**
  - New `InventoryError::Incomplete`.
  - `receiver_read_complete(pairing_count, paired_len)`: a Bolt read is complete only when `count_pairings()` succeeded **and** at least that many slots read. An unreadable count or a short read is incomplete; reading *more* than the count is still complete.
  - `probe_one` reports that completeness; a probe **timeout** counts as incomplete too.
  - On an incomplete cycle `enumerate()` **skips cache eviction** (a device merely unread this cycle keeps its entry) and returns `Incomplete` rather than publishing a truncated list.
  - The one-shot `enumerate()` free fn (CLI / `diag`) has no last-known-good to fall back on, so it **retries** an `Incomplete` cycle a few times before giving up — the report's two isolated runs read 3 devices and 0; a retry would have read 3 both times.
- **`openlogi-agent-core/watchers/inventory.rs`** — factor the tick→event decision into `WatchState::classify` (no behaviour change; the `Err` arm already kept the last snapshot) so it's unit-testable.

A genuine unpair/disconnect still reads cleanly as an `Ok` carrying the receiver's real, smaller count, so the list still shrinks when devices actually leave.

## Tests

New unit tests (host-only, no hardware):

- `receiver_read_complete`: `Some(3)/3` and `Some(0)/0` complete; `Some(3)/2`, `Some(3)/0`, `None/*` incomplete.
- `WatchState::classify`: an incomplete read after a success **keeps** the last snapshot; a clean empty `Ok` is still forwarded as a disconnect; persistent initial failure reports `Unavailable` once, then recovers.

```
cargo fmt --all -- --check                                                                          # clean
cargo clippy -p openlogi-hid -p openlogi-agent-core -p openlogi-cli --all-targets -- -D warnings    # clean
cargo test  -p openlogi-hid -p openlogi-agent-core                                                   # 46 + 21 pass
```

I couldn't build `openlogi-gui` locally (it compiles Metal shaders against a full Xcode toolchain I don't have), but the GUI is untouched, `InventoryError` has no exhaustive match anywhere, and the new variant is additive — so the GUI build is unaffected. CI covers it.

## Manual hardware verification still needed

I don't have the Bolt-receiver hardware, so the live repro needs a hand (ideally @buliwyf42, who filed it):

1. Build the agent from this branch and run it against the 3-device Bolt setup on macOS 27 beta.
2. Watch the Devices view / agent logs for ~60 s. Expected: the list stays populated; on a bad cycle the agent logs `receiver read incomplete this cycle … keeping last snapshot` and emits **no** snapshot, instead of flapping to "No devices".
3. Power one device off → it should drop from the list within a cycle or two (a clean `Ok` with the smaller count), confirming genuine disconnects still propagate.

Closes the flapping half of #218.

### DIFF
--- a/crates/openlogi-agent-core/src/watchers/inventory.rs
+++ b/crates/openlogi-agent-core/src/watchers/inventory.rs
@@ -35,6 +35,53 @@ pub enum InventoryEvent {
     Unavailable,
 }
 
+/// The watcher's cross-tick memory, factored out of the poll loop so the
+/// tick → event decision is unit-testable without spawning the thread or
+/// touching real HID.
+#[derive(Default)]
+struct WatchState {
+    /// Set once any enumeration has completed. After that, a failed tick keeps
+    /// the last good snapshot forever instead of ever reporting `Unavailable`.
+    succeeded: bool,
+    /// Consecutive failures, counted only before the first success.
+    initial_failures: u8,
+}
+
+impl WatchState {
+    /// Decide what (if anything) a watch tick emits.
+    ///
+    /// - `Ok(snapshot)` — a completed enumeration (an empty one included: that's
+    ///   a genuine disconnect) — is forwarded so the agent's device set tracks
+    ///   reality.
+    /// - `Err(..)` means "couldn't fully check" — a transient HID++ timeout or a
+    ///   partial receiver read ([`openlogi_hid::InventoryError::Incomplete`]):
+    ///   emit nothing, so the agent keeps its last good device set and live
+    ///   bindings instead of wiping them for ~one period — the flap behind #218.
+    ///   Before the *first* success there is no good set to keep, so persistent
+    ///   initial failure is reported once as [`InventoryEvent::Unavailable`]; the
+    ///   loop keeps retrying and a later success recovers.
+    fn classify(
+        &mut self,
+        result: Result<Vec<DeviceInventory>, openlogi_hid::InventoryError>,
+    ) -> Option<InventoryEvent> {
+        match result {
+            Ok(inv) => {
+                self.succeeded = true;
+                Some(InventoryEvent::Snapshot(inv))
+            }
+            Err(e) => {
+                warn!(error = ?e, "enumerate failed during watch tick — keeping last snapshot");
+                if self.succeeded {
+                    return None;
+                }
+                self.initial_failures = self.initial_failures.saturating_add(1);
+                (self.initial_failures == INITIAL_FAILURE_LIMIT)
+                    .then_some(InventoryEvent::Unavailable)
+            }
+        }
+    }
+}
+
 /// Spawn the watcher and return a receiver of inventory events. The
 /// channel is unbounded so a slow consumer cannot back-pressure the HID
 /// poll loop into stalling on a real device disconnect.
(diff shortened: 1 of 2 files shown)


buliwyf42:
Closing in favor of #222 — it fixes the same transient-probe-failure root cause behind #218, and more thoroughly than this PR.

We independently landed on the same core signal (a Bolt read is healthy only when `paired.len() == pairing_count`), but #222's per-node `NodeLedger` is the better design: it replays each node's last good inventory for a grace window instead of dropping the whole enumerate cycle, so a partial read on one receiver no longer suppresses a cleanly-read second one, and cache eviction keeps running. It also evicts a wedged channel, which this PR didn't touch.

The only bit here that #222 doesn't cover is the one-shot `enumerate()` free fn used by the CLI `list`/`diag` commands — I left a note about that on #222. Thanks for the fast turnaround!

PR #222, opened 2026-06-12
fix(hid): replay a node's last inventory through transient probe failures

### LINKED ISSUE #218: [Bug]: Device list flaps (appears then vanishes ~2s) + device images never load — Bolt receiver, macOS 27 beta

AprilNEA:
## Problem

`Enumerator::enumerate()` returns `Ok` with an empty or partial device list when a receiver *probe* fails or times out — the inventory watcher only guards against `Err` from OS-level enumeration. A transient HID++ hiccup (busy receiver, timed-out feature walk, failed open) therefore masquerades as "checked: no devices", and the agent wipes live bindings/DPI state for devices that never went anywhere.

Worse, a sick channel was only ever evicted when its OS node vanished. `read_report` parks forever on `HidError::Disconnected` *expecting* the watcher to evict the channel (see the comment in `transport.rs`), but for a node that stays visible while its channel is wedged, that eviction never came — the probe failed every tick from then on. This is the agent-side half of the 0.6.6 device-loss reports (#218).

## Fix

New `NodeLedger` (per-HID-node probe-health ledger) in `openlogi-hid`:

- Each probe now reports a **health** verdict alongside its inventory: a Bolt probe is healthy when the pairing enumeration is complete (`paired.len() == pairing_count`); a direct probe is healthy when its capabilities walk succeeded.
- On an unhealthy or timed-out probe (and on a failed open), the ledger **replays the node's last good inventory** for up to `NODE_MISS_GRACE = 3` consecutive misses, so a transient failure no longer erases the node's devices. Recovery resets the count; after the grace expires, whatever the live (possibly partial) probe returned is surfaced.
- After `CHANNEL_EVICT_AFTER = 2` consecutive misses, the node's HID++ channel is **dropped and reopened on the next tick** — completing the recovery loop `transport.rs` documents ("until the inventory watcher evicts the channel"). A wedged channel can now actually heal.
- Vanished nodes are pruned from the ledger so it can't grow unboundedly.

Pure device-classification helpers moved out of `inventory.rs` into a new `mappings.rs` (file-size limit); no behavior change there.

## Notes

- The ledger is generic over the node key (`async_hid::DeviceId` is `#[non_exhaustive]` and cfg-gated per OS), which is also what makes it unit-testable.
- Agent-internal only — no IPC wire-format change, no protocol bump.

Fixes #218

## Tests

- 6 new ledger tests: replay-within-grace, grace expiry, recovery reset, eviction threshold, healthy-`None` clearing, vanished-node pruning.
- `cargo fmt` / `clippy --workspace --all-targets -D warnings` / full test suite green.

### DIFF
--- a/crates/openlogi-hid/src/inventory.rs
+++ b/crates/openlogi-hid/src/inventory.rs
@@ -11,34 +11,33 @@ use hidpp::{
     channel::HidppChannel,
     device::Device,
     feature::{
-        CreatableFeature,
-        device_information::DeviceInformationFeature,
-        device_type_and_name::{DeviceType as HidppDeviceType, DeviceTypeAndNameFeature},
-        unified_battery::{
-            BatteryLevel as HidppBatteryLevel, BatteryStatus as HidppBatteryStatus,
-            UnifiedBatteryFeature,
-        },
+        CreatableFeature, device_information::DeviceInformationFeature,
+        device_type_and_name::DeviceTypeAndNameFeature, unified_battery::UnifiedBatteryFeature,
     },
     receiver::{
         self, Receiver,
         bolt::{
-            DeviceConnection as BoltDeviceConnection, DeviceKind as BoltDeviceKind,
-            Event as BoltEvent, Receiver as BoltReceiver,
+            DeviceConnection as BoltDeviceConnection, Event as BoltEvent, Receiver as BoltReceiver,
         },
         unifying::{
-            DeviceConnection as UnifyingDeviceConnection, DeviceKind as UnifyingDeviceKind,
-            Event as UnifyingEvent, Receiver as UnifyingReceiver,
+            DeviceConnection as UnifyingDeviceConnection, Event as UnifyingEvent,
+            Receiver as UnifyingReceiver,
         },
     },
 };
 use openlogi_core::device::{
-    BatteryInfo, BatteryLevel, BatteryStatus, Capabilities, DeviceInventory, DeviceKind,
-    DeviceModelInfo, DeviceTransports, PairedDevice, ReceiverInfo,
+    BatteryInfo, Capabilities, DeviceInventory, DeviceKind, DeviceModelInfo, DeviceTransports,
+    PairedDevice, ReceiverInfo,
 };
 use thiserror::Error;
 use tokio::time::timeout;
 use tracing::{debug, warn};
 
+use crate::mappings::{
+    map_battery_level, map_battery_status, map_device_type, map_kind, map_unifying_kind,
+    normalize_serial_number, resolve_device_kind,
+};
+use crate::node_ledger::NodeLedger;
 use crate::route::DIRECT_DEVICE_INDEX;
 use crate::transport::{enumerate_hidpp_devices, open_hidpp_channel};
 
@@ -225,6 +224,10 @@ pub struct Enumerator {
     /// steadily-connected node is opened once here and reused until it
     /// disconnects.
     channels: HashMap<async_hid::DeviceId, CachedChannel>,
+    /// Per-node last-good inventory + consecutive-failure counts: replays a
+    /// node's snapshot through transient probe failures and decides when its
+    /// cached channel must be dropped and reopened (see [`crate::node_ledger`]).
+    ledger: NodeLedger<async_hid::DeviceId>,
     tick: u64,
 }
 
@@ -260,6 +263,12 @@ impl Enumerator {
     /// HID candidate concurrently (so one asleep node that burns the whole
     /// `PROBE_BUDGET` can't stall the others), reusing each device's cached
     /// immutable data when it's present and fresh.
+    ///
+    /// A node the OS still lists but whose probe fails (receiver registers
+    /// unanswered, probe timeout, open failure) is **not** reported as absent:
+    /// its last completed inventory is replayed for a bounded grace and its
+    /// channel is reopened, so a transient HID++ glitch can't masquerade as
+    /// "no devices" (#218) — see [`crate::node_ledger`].
     pub async fn enumerate(&mut self) -> Result<Vec<DeviceInventory>, InventoryError> {
         self.tick = self.tick.wrapping_add(1);
         let tick = self.tick;
@@ -272,6 +281,7 @@ impl Enumerator {
         // lookups — an actual open happens only when a new device appears.
         let mut active: Vec<(async_hid::DeviceInfo, Arc<HidppChannel>)> = Vec::new();
         let mut seen_nodes: HashSet<async_hid::DeviceId> = HashSet::new();
+        let mut open_failures: Vec<async_hid::DeviceId> = Vec::new();
         for dev in candidates {
             let node = dev.id.clone();
             seen_nodes.insert(node.clone());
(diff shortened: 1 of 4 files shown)


buliwyf42:
This is a cleaner take on the same #218 root cause than my #221 (now closed in favor of this) — the per-node ledger handles the multi-receiver "one bad receiver suppresses the others" case and the cache-eviction starvation that a whole-cycle guard doesn't.

One small gap the ledger doesn't reach: the one-shot `enumerate()` free fn (CLI `list` / `diag`) builds a fresh `Enumerator` per call, so its ledger has no prior snapshot to replay — a transient hiccup there still surfaces as an empty/partial list. A small bounded retry on that path (re-running the enumerate a couple of times on an unhealthy/empty result) would cover it. Happy to send that as a tiny follow-up on top of this if useful.

AprilNEA:
Yes please — send the follow-up! You've read the gap exactly right: a fresh `Enumerator` has no snapshot to replay, and your 3-then-0 isolated-runs observation in #218 is precisely that case. Two notes for it:

- Master just gained #226 (a 5 s request timeout inside `HidppChannel::send`), so a dead probe now fails fast instead of burning the whole `PROBE_BUDGET` — a short bounded retry on the one-shot path should be cheap.
- If you'd like to bring over the `WatchState::classify` refactor + watcher tests from #221 in the same PR, they're conflict-free (this PR didn't touch the watcher) and the testability win stands.

And thanks for the gracious close on #221 — the diagnosis there matched what landed here almost line for line, and the `receiver_read_complete` test matrix was genuinely good.

One more ask, honestly the most valuable one: you have the only known repro rig (macOS 27 beta, MX Master 4 on Bolt). Now that this is merged, would you run a master build against it for a minute? Expected: the device list stays populated (a bad cycle logs a replay instead of flapping to "No devices"), and powering a device off still drops it within a cycle or two. If the flap survives on your stack, just reopen #218.

buliwyf42:
Following up here, since the live re-test on #218 lands squarely on this PR's code.

On the repro rig (macOS 27 beta, MX Master 4 on Bolt) the ledger absorbs short glitches exactly as designed — but it surfaced a **sustained** failure: a paired online device's deep `probe_features` hangs *every* cycle, burning the full 5 s `PROBE_BUDGET`. Since `probe_one` is the unit wrapped in `timeout(PROBE_BUDGET)`, firing it discards the whole node — including the pairing-register reads that succeeded — so the receiver yields nothing → `count=0` "No devices" (confirmed in the GUI; only a physical receiver replug cleared it; a standalone `openlogi list` hangs identically, so it's device/OS-level, not capture contention).

The telling part: the **Unifying** path already guards this with `UNIFYING_SLOT_PROBE` (3.5 s per-slot cap → partial instead of starving the budget). The **Bolt** slot probe has no equivalent — `probe_or_reuse` is awaited bare in `probe_bolt_slot`, so one hung Bolt device blows the whole-receiver budget.

I've opened #251 to mirror that guard onto Bolt (a `BOLT_SLOT_PROBE` cap → the hung device falls back to cached/identity data while the rest of the receiver still enumerates). Writeup + logs on #218.

PR #226, opened 2026-06-12
fix: add a request timeout inside HidppChannel::send (fix hang root at the right layer)

### LINKED ISSUE #52: hidpp: add a request timeout inside HidppChannel::send (fix hang root at the right layer)
## Problem

`hidpp`'s `HidppChannel::send` awaits a oneshot that is only resolved when the read thread sees a matching response. On device silence (asleep / on another host / gone) it **never resolves**, so the future hangs forever. There is no per-request timeout in the protocol layer.

OpenLogi currently works around this **per caller**, which is the wrong altitude:
- `openlogi-hid::inventory::enumerate` wraps each probe in `timeout(PROBE_BUDGET, …)`.
- `openlogi-gui::hardware` wraps DPI/SmartShift writes in `timeout(WRITE_BUDGET, …)`.

Any future `send()` caller (new feature, pairing flow, etc.) is exposed again, and the enumerate-side timeout makes "drop a channel mid-I/O" a routine path (which runs `HidppChannel::Drop`'s blocking `join()` — see review finding #2).

## Proposal

Add a timeout **inside `HidppChannel::send`** (the vendored `openlogi-hidpp` fork): race the response receiver against a timer; on elapse, remove the pending-message entry and return `ChannelError::NoResponse`. `send_and_forget` needs no timeout (it only writes).

- The fork is tokio-free, so this needs a runtime-agnostic timer — e.g. `futures-timer` (small, `futures`-based). Add it to `crates/openlogi-hidpp/Cargo.toml`.
- Once `send()` is bounded, the per-caller `PROBE_BUDGET` / `WRITE_BUDGET` timeouts can be simplified or removed.
- **Candidate to upstream** to `lus/logy`.

## Acceptance
- A `send()` against an unresponsive device returns `NoResponse` within the budget (no hang, no thread/channel leak).
- No pending-message leak on timeout.
- enumerate + DPI/SmartShift writes no longer need their own timeout wrappers (or keep them as cheap belt-and-suspenders).

Context: review of #45 (findings #2, #10).

mvanhorn:
## Summary

`HidppChannel::send` no longer hangs forever when a device goes silent. The oneshot receiver is now raced against a runtime-agnostic `futures-timer` delay (the crate stays tokio-free; `channel.rs` already uses `futures::select`), and on elapse the request's entry is removed from `pending_messages` and the existing `ChannelError::NoResponse` is returned. `send_and_forget` stays untimed since it only writes.

## Why this matters

#52 (maintainer-authored, status: confirmed, from the PR #45 review): when a device is asleep, on another host, or unplugged, the read thread never sees a matching response and the future never resolves, so every new `send()` caller is re-exposed to the hang and works around it individually (`PROBE_BUDGET` in inventory, `WRITE_BUDGET` in the GUI hardware path). The protocol layer now owns the timeout with a 5s default matching those caller budgets; the per-caller wrappers can stay as belt-and-suspenders, so no caller changes are needed.

## Testing

New tests in `channel.rs` on the mock raw channel: an unanswered request returns `NoResponse` within the budget and leaves `pending_messages` empty (no lazy-sweep reliance), answered requests are unaffected, and the timeout path coexists with concurrent pending requests. The package suite passes locally.

Fixes #52


mvanhorn:
Appreciate you getting this in, @AprilNEA. HidppChannel::send won't hang forever on a silent device anymore.

PR #233, opened 2026-06-12
Add Linux port support and packaging

jvannoyx4:
## Summary

This PR adds Linux support for OpenLogi, including device access setup, Linux-compatible button actions, GUI desktop integration, and Linux release packaging.

## What changed

- Added Linux mappings for desktop/workspace actions that were previously macOS-specific
- Added Linux action injection through uinput for keyboard and pointer actions
- Added GNOME workspace shortcut support for previous/next workspace actions
- Added Linux GUI window chrome support, including working minimize/maximize/close controls and draggable titlebar behavior
- Added app ID/icon integration so OpenLogi appears correctly in desktop launchers and task bars
- Updated Linux permissions handling and udev rules for hidraw, input, and uinput access
- Added/updated Linux install documentation
- Added Linux release packaging:
  - `.deb`
  - `.rpm`
  - portable `.tar.gz`
- Updated release workflow so Linux tarballs are included in future release assets

## Validation

Tested locally on Linux:

- `cargo check -p openlogi-gui -p xtask`
- `cargo test -p openlogi-core linux_accelerators`
- `sh -n` on Linux install/uninstall/package scripts
- Built release binaries
- Generated Linux package artifacts:
  - `openlogi-0.6.8-linux-amd64.tar.gz`
  - `openlogi_0.6.8_amd64.deb`
  - `openlogi-0.6.8-1.x86_64.rpm`


recchia:
Real-world confirmation that the new `SUBSYSTEM=="input"` rule is needed: on Ubuntu 26.04 (GNOME Shell 50.1, native Wayland) with the currently packaged `70-openlogi.rules` installed, everything is ACL'd correctly **except** the event nodes:

- `/dev/uinput` — ✅ `uaccess` ACL present (`user:<seat-user>:rw-`)
- Logitech `hidraw` — ✅ covered by the VID 046D rules
- `/dev/input/event*` — ❌ `root:input 0660`, no ACL → the evdev hook gets `EACCES` opening the mouse nodes (an MX Master 3S, plus the pointer subdevice of an MX Keys keyboard — kernel node "MX Keys Mouse"), so button capture fails even though frontmost detection (GNOME extension) and HID++ device control work

So the only workaround today is the keylogger-capable `input` group; the scoped per-vendor `uaccess` rule in this PR is the right fix. I'll test the updated rule from this branch on the same machine and follow up.

recchia:
Follow-up as promised: tested the updated rule from this branch on the same machine (Ubuntu 26.04, GNOME Shell 50.1, native Wayland, systemd-logind), installed via `cp` + `udevadm control --reload-rules` + `udevadm trigger` — no re-login needed.

Results:

- ✅ Both Logitech mouse event nodes (MX Master 3S via Bolt receiver, plus the pointer subdevice the MX Keys keyboard exposes — kernel node "MX Keys Mouse") picked up the `uaccess` ACL (`user:<seat-user>:rw-`)
- ✅ Functional check: `open(O_RDWR)` + exclusive `EVIOCGRAB` + release succeeds on both nodes as the unprivileged seat user — the exact sequence the evdev hook performs
- ✅ `/dev/uinput` still opens for writing
- ✅ Correctly scoped: a non-Logitech pointer device (Elan touchpad, `ID_INPUT_MOUSE` but different vendor) did **not** get an ACL, and keyboards remain inaccessible — so this avoids handing the user blanket `input`-group access

LGTM from the Wayland/GNOME side.

recchia:
Heads-up on a Wayland app_id / `StartupWMClass` overlap with #191

This PR and #191 (Wayland frontmost backends) both set the GUI's Wayland `app_id` + the desktop file's `StartupWMClass`, but with different values, so they'll conflict — and it's semantic, not just textual:

- This PR: `app_id: Some(APP_ID)` (`APP_ID = "openlogi"`) and `StartupWMClass=openlogi`
- #191: `app_id: Some("org.openlogi.openlogi")` and `StartupWMClass=org.openlogi.openlogi`

In #191 the `gnome_shell` frontmost backend identifies OpenLogi's *own* window by this exact string, and `StartupWMClass` has to match it for GNOME to group the window under the launcher icon — so the two values have to agree on one canonical id.

I'd propose standardizing on **`org.openlogi.openlogi`**, since that's already the established identifier across the repo:
- `crates/openlogi-gui/Cargo.toml` → `identifier = "org.openlogi.openlogi"`
- `xtask/src/manifest.rs` → `const APP_ID: &str = "org.openlogi.openlogi"`
- consistent with the macOS bundle-id family (`org.openlogi.agent`, `org.openlogi.openlogi.dev`)

Two small follow-ons if we go with that:
1. This PR's new `APP_ID = "openlogi"` const in `main.rs` would duplicate/diverge from the existing `APP_ID` in `xtask/manifest.rs` — worth reusing one source of truth.
2. The desktop file is `openlogi.desktop` while the app_id is reverse-DNS; `StartupWMClass=org.openlogi.openlogi` bridges that fine, so no rename is required (though `org.openlogi.openlogi.desktop` would be the strict freedesktop convention if we ever want it).

Happy to rebase #191 onto whatever lands here — just want us aligned on the one id string before either merges so the frontmost self-identification doesn't silently break.

recchia:
Quick update: I've aligned #191 on `org.openlogi.openlogi` so there's nothing blocking on my side.

Rather than leave a bare literal, I hoisted the value into `openlogi_core::brand::APP_ID` (the platform-free brand module that already holds the shared URLs / deep-link vocab) and the GUI now references that const; the `.desktop` `StartupWMClass` keeps a documented literal copy since it can't import Rust.

So when this PR lands, the conflict in `main_window_options` should resolve to just `app_id: Some(brand::APP_ID.into())`, and the local `APP_ID = "openlogi"` const here can drop in favor of the shared one. Happy to rebase #191 on top whenever it's convenient — just say the word.

PR #237, opened 2026-06-13
fix(hid): retry the one-shot enumerate() through a transient probe miss (#218)

buliwyf42:
Follow-up to #222 (merged), as invited in [its review](https://github.com/AprilNEA/OpenLogi/pull/222#issuecomment-4691532623).

## Gap

#222's per-node `NodeLedger` replays a node's last good inventory across watcher ticks — that fixes the flapping device list. But the **one-shot** `enumerate()` free fn (used by the CLI `list` / `diag` commands) builds a *fresh* `Enumerator` per call, so its ledger is empty: a transient probe miss has nothing to replay and still surfaces as an empty/partial list. That's exactly the "two back-to-back isolated runs read 3 devices and 0" case in #218.

## Fix

**`openlogi-hid`** — split the enumerate pass into `enumerate_reporting_health`, which also returns whether every probed node answered cleanly this pass (a probe timeout, an open failure, or a receiver read short of its pairing count is unhealthy).
- The polling watcher keeps calling `enumerate()` and ignores the flag — the ledger already handles its per-node replay, so nothing changes there.
- The one-shot `enumerate()` retries while a node is unhealthy, **reusing the same `Enumerator`** so its ledger accumulates a snapshot a later attempt can replay and the opened channel stays warm. Bounded to 4 attempts, 300 ms apart. #226's 5 s `HidppChannel::send` timeout makes a dead probe fail fast, so the retry is cheap.

A legitimately-skipped node (a receiver's secondary HID interface) reports `healthy: true` after a completed feature walk, so it doesn't trigger needless retries; zero candidates returns immediately (no spin).

**`openlogi-agent-core`** — factor the watcher's tick→event decision into `WatchState::classify` and unit-test it (invited in the #222 review; the watcher was untouched there). No behaviour change — the `Err` arm already kept the last snapshot. Tests assert: an enumerate failure after a success keeps the last snapshot; a clean empty `Ok` is still forwarded as a genuine disconnect; persistent initial failure reports `Unavailable` once, then recovers. (Adds `async-hid` as a dev-dependency only, to build an `InventoryError::Hid` in those tests.)

## Verification

```
cargo fmt --all -- --check                                                                          # clean
cargo clippy -p openlogi-hid -p openlogi-agent-core -p openlogi-cli --all-targets -- -D warnings    # clean
cargo test  -p openlogi-hid -p openlogi-agent-core                                                   # 61 + 23 pass (3 new watcher tests)
```

I can't build `openlogi-gui` locally (its Metal shaders need a full Xcode toolchain I don't have), but nothing here touches the GUI — `InventoryError` stays a single-variant enum with no exhaustive match anywhere, and CI covers the GUI.

Refs #218, #222.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4699,6 +4699,7 @@ dependencies = [
 name = "openlogi-agent-core"
 version = "0.6.12"
 dependencies = [
+ "async-hid",
  "bincode",
  "interprocess",
  "openlogi-core",
--- a/crates/openlogi-agent-core/Cargo.toml
+++ b/crates/openlogi-agent-core/Cargo.toml
@@ -25,3 +25,5 @@ workspace = true
 
 [dev-dependencies]
 bincode = "1.3"
+# Only to construct an `InventoryError::Hid` in the watcher's classify tests.
+async-hid = { workspace = true }
--- a/crates/openlogi-agent-core/src/watchers/inventory.rs
+++ b/crates/openlogi-agent-core/src/watchers/inventory.rs
@@ -48,6 +48,54 @@ pub enum InventoryEvent {
     SystemWake,
 }
 
+/// The watcher's cross-tick memory, factored out of the poll loop so the
+/// tick → event decision is unit-testable without spawning the thread or
+/// touching real HID.
+#[derive(Default)]
+struct WatchState {
+    /// Set once any enumeration has completed. After that, a failed tick keeps
+    /// the last good snapshot forever instead of ever reporting `Unavailable`.
+    succeeded: bool,
+    /// Consecutive failures, counted only before the first success.
+    initial_failures: u8,
+}
+
+impl WatchState {
+    /// Decide what (if anything) a watch tick emits.
+    ///
+    /// - `Ok(snapshot)` — a completed enumeration (an empty one included: that's
+    ///   a genuine disconnect) — is forwarded so the agent's device set tracks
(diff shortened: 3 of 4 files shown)


buliwyf42:
Re: the Unifying health-signal P2 — good catch, but that coarseness is by design from #222, not introduced here. The Unifying path only surfaces *online* devices (via arrival events) — it doesn't poll offline slots the way Bolt does — so `paired.len() == pairing_count` can't be its health signal: a legitimately-offline Unifying device would fall short of the count every cycle and wedge it into a permanent "unhealthy" retry. Hence the coarser `pairing_count.is_some()`.

This PR's one-shot retry just keys off whatever `healthy` the probe already reports, so it inherits that asymmetry rather than changing it. A tighter Unifying signal (e.g. an online-count baseline) is a reasonable follow-up, but it has to account for offline devices and arrival-drain timing, so I'd keep it out of this small retry-focused change. (The #218 repro is a Bolt receiver, where the signal is already tight.)

Addressed the loop-range nit, and rebased onto master to resolve the #223 conflict — back to MERGEABLE.

FWIW: I re-tested this branch against the real repro rig (macOS 27 beta, MX Master 4 on Bolt). The one-shot retry demonstrably works — `openlogi list` hit 3 deep-probe timeouts but returned all 3 devices on the 4th attempt. But the deeper #218 failure (a *sustained* deep-probe hang that only a receiver replug clears) is separate and survives; I've written it up on #218.

PR #251, opened 2026-06-13
fix(hid): bound the Bolt per-slot probe so one hung device can't drop the whole receiver (#218)

buliwyf42:
## Context

Live re-testing #218 on the repro rig (macOS 27 beta, MX Master 4 + MX Master 3S + MX Mechanical Mini on one Bolt receiver) showed that #222 (`NodeLedger`) and #237 (one-shot retry) fix the *short transient* misses, but a **sustained** failure still drops the device list to "No devices" — details in [#218](https://github.com/AprilNEA/OpenLogi/issues/218). This PR fixes that remaining case.

## Root cause

A single paired **online** device's deep feature walk (`probe_features` / `Device::new`) hangs on every cycle, burning the whole 5 s `PROBE_BUDGET`. Each failing cycle still reads `pairing_count=Some(3)` and all three slot codenames — only the deep walk hangs — but because `probe_one` is the unit wrapped in `timeout(PROBE_BUDGET)`, when it fires the **entire node** is discarded, including the pairing-register reads that succeeded. The receiver yields nothing → GUI `inventory refreshed count=0`.

The tell: the **Unifying** slot probe already guards exactly this with `UNIFYING_SLOT_PROBE` (a per-slot cap, so a slow walk returns partial instead of starving the budget). The **Bolt** slot probe had no equivalent — `probe_or_reuse` was awaited bare in `probe_bolt_slot`, so one hung Bolt device blew the whole-receiver budget.

## Fix

Mirror the Unifying guard onto the Bolt path: a new `BOLT_SLOT_PROBE` (1.5 s) wraps the Bolt slot's `probe_or_reuse`; on timeout the slot falls back to its cached / identity-only data (codename + kind + online come from the pairing register, which reads fine every cycle) instead of letting the whole `probe_one` time out. A hung device keeps showing with its last-known capabilities while the rest of the receiver still enumerates.

1.5 s is well above a healthy BTLE walk (sub-second) yet small enough that **two** simultaneously-hung slots still fit `PROBE_BUDGET` after the 1.5 s arrival drain; the rare all-slots-hung case degrades to the existing per-node ledger replay (#222).

## Verification

```
cargo fmt --all -- --check                                   # clean
cargo clippy -p openlogi-hid --all-targets -- -D warnings     # clean
cargo test  -p openlogi-hid                                   # 61 pass
```

I couldn't add a unit test (the per-slot timeout needs a live/mock HID++ channel, same as the existing Unifying guard, which also has none). The change mirrors that proven pattern exactly.

**Manual hardware verification still recommended:** I have the repro rig but the wedge cleared on a receiver replug and I can't trigger the sustained hang on demand. The expected behaviour with this patch: when a device's deep probe hangs, the agent logs `Bolt slot probe timed out; using cached data if available` and the device list **stays populated** (the hung device keeps its last-known card) instead of dropping to "No devices".

Refs #218, #222.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5014,7 +5014,7 @@ version = "0.6.20"
 dependencies = [
  "async-hid",
  "bincode",
- "futures-lite 2.6.1",
+ "futures-lite",
  "interprocess",
  "openlogi-core",
  "openlogi-hid",
--- a/crates/openlogi-hid/src/inventory.rs
+++ b/crates/openlogi-hid/src/inventory.rs
@@ -57,6 +57,21 @@ const PROBE_BUDGET: Duration = Duration::from_secs(5);
 /// just lacks capabilities / battery until the next tick.
 const UNIFYING_SLOT_PROBE: Duration = Duration::from_millis(3500);
 
+/// Per-slot budget for the HID++ 2.0 feature walk on a Bolt paired device.
+///
+/// The whole receiver shares one [`PROBE_BUDGET`]; without a per-slot cap a
+/// single online device that stops answering its feature-walk reads (seen on a
+/// recent macOS IOHID stack with a new MX Master 4) burns the entire budget, so
+/// `probe_one` times out and the receiver yields *nothing* — every paired device
+/// drops to "No devices" even though its pairing-register identity read fine
+/// (#218). Capping each slot lets a hung device fall back to its cached /
+/// identity-only data while the rest of the receiver still enumerates, mirroring
+/// [`UNIFYING_SLOT_PROBE`]. Bolt BTLE round-trips are fast (a healthy walk is
+/// well under a second), so 1 s is generous headroom yet small enough that three
+/// online slots can hang at once and still fit `PROBE_BUDGET` after the 1.5 s
+/// arrival drain (1.5 + 3×1 = 4.5 s).
+const BOLT_SLOT_PROBE: Duration = Duration::from_secs(1);
+
 /// Errors raised while enumerating HID++ devices.
 #[derive(Debug, Error)]
 pub enum InventoryError {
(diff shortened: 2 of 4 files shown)


buliwyf42:
Live-tested this on the #218 rig (macOS 27 beta, MX Master 4 + MX Master 3S + MX Mechanical Mini on a Bolt receiver) and tightened it as a result.

The per-slot cap works: when a slot's deep walk hung, `Bolt slot probe timed out … slot=N` fired and that device kept its cached card instead of the whole receiver vanishing — confirmed live for 1–2 hung slots.

But I hit a budget bug at my first value (1.5 s). When the MX Master 3S woke so **all three** slots were online and hung at once, `1.5 s arrival drain + 3 × 1.5 s = 6 s` overran the 5 s `PROBE_BUDGET`, so the outer `timeout(PROBE_BUDGET, probe_one)` still fired and dropped the whole node. Tightened to `BOLT_SLOT_PROBE = 1 s` so all three online slots fit (`1.5 + 3×1 = 4.5 s`); pushed. (The all-three-at-1s case is budget-derived — the wedge is non-deterministic, so I couldn't re-trigger it on demand without more churn.)

One honest caveat: this keeps the *device list* populated (via the cached fallback), but it doesn't cure the underlying wedge — during it the device genuinely stops answering HID++ (writes like DPI/SmartShift would still time out), and only a physical receiver replug cleared it on my rig. So #251 is graceful degradation for the symptom; the root macOS-27 IOHID wedge stays the separate #218 track.

buliwyf42:
Thanks — already addressed. After live-testing on the rig I hit exactly this: with the MX Master 3S awake, all three slots online and hung at once overran the budget at 1.5 s/slot. The current head (`34314ba`) tightens `BOLT_SLOT_PROBE` to `Duration::from_secs(1)`, so `1.5 s drain + 3×1 s = 4.5 s` fits `PROBE_BUDGET` — your review picked up the earlier 1.5 s push. (Empty slots fail fast on `UnknownDevice` without a deep probe, so the worst realistic case is the 3 paired devices, not 6 slots.)

PR #252, opened 2026-06-13
fix(gui): don't show a multi-pid device twice in the carousel

buliwyf42:
## Problem

On 0.6.12, after Easy-Switching an **MX Master 3S** to another host, the device carousel showed **two** "MX Master 3S" cards.

## Root cause

`config_key` is `format!("{:x}{:04x}", extended_model_id, model_ids[0])` — it keys on the device's *primary reported pid*. A device with more than one model id reports a different primary pid per transport: the MX Master 3S is **b034 over BTLE** but **b043 via a Bolt receiver** (exactly #137). So the same physical mouse resolves to a different `config_key` across connections.

#224's "keep asleep devices" union (`append_offline_known`) re-adds a persisted-identity placeholder for every known device *absent from the live list*, matched **only by `config_key`**. So when the 3S was seen under both pids — a live/cached record under one, a stale persisted identity under the other — neither matched the other's key, and both showed:

```
config_key 0b043  →  live/cached "MX Master 3S"      (current transport)
config_key 0b034  →  offline placeholder "MX Master 3S" (stale, other pid)
```

Switching the mouse to another host is what surfaced it: the live record went offline/cached while the placeholder under the other pid stayed.

## Fix

Dedup the offline-placeholder union by **(display name, kind)** as well as `config_key` — both against the live records *and* among the placeholders themselves (a device persisted under both pids must still surface once). The asset-resolved display name is per-model, so it collapses the pid variants. No schema change, no asset resolver needed in this hot path (the function stays testable in isolation, as its doc comment requires).

The edge case — two physically distinct devices of the *same* model, one online and one offline-elsewhere — collapses to one card while the second is away (it carries `route: None`, so it's unconfigurable until it returns anyway, at which point it gets its own live record). That's a far rarer and milder outcome than a phantom duplicate of one device.

## Tests (`cargo test -p openlogi-gui state::devices`, 9 pass)

- `one_model_under_two_pids_is_not_duplicated` — live record under one pid + stale identity under the other → one card; a genuinely different known device is still added.
- `one_offline_model_persisted_under_two_pids_collapses_to_one_card` — both pids persisted, neither live → one card (placeholder-vs-placeholder dedup).
- existing `known_devices_are_appended_only_when_absent_from_live` still passes.

`cargo fmt --all -- --check` clean, `cargo clippy -p openlogi-gui --all-targets -- -D warnings` clean.

## Verification

Found and reproduced live on 0.6.12 (macOS 27 beta, MX Master 3S on a Bolt receiver). Refs #137, #224.

### DIFF
--- a/crates/openlogi-gui/src/state/devices.rs
+++ b/crates/openlogi-gui/src/state/devices.rs
@@ -111,20 +111,37 @@ pub(super) fn build_device_list(
 }
 
 /// Append an offline placeholder for every known device not already present in
-/// `list` (matched by `config_key`). Split out from [`build_device_list`] so
-/// the union rule is unit-testable without an [`AssetResolver`].
+/// `list`. Split out from [`build_device_list`] so the union rule is
+/// unit-testable without an [`AssetResolver`].
+///
+/// A device is "already present" when its `config_key` matches **or** a record
+/// with the same (display name, kind) is already in the list. The second match
+/// is what stops a one-physical-device-two-cards phantom: a device with more
+/// than one model id resolves to a *different* `config_key` per transport (the
+/// MX Master 3S is pid `b034` over BTLE but `b043` via a Bolt receiver — #137),
+/// so a stale identity persisted under the other pid would otherwise show as a
+/// second card. The resolved display name is per-model, so it collapses those
+/// pid variants — both against a live record and among the placeholders
+/// themselves (a device persisted under both pids must still surface once).
 fn append_offline_known<'a>(
     list: &mut Vec<DeviceRecord>,
     known: impl Iterator<Item = (&'a str, &'a DeviceIdentity)>,
 ) {
-    let present: HashSet<&str> = list.iter().map(|r| r.config_key.as_str()).collect();
-    // Collect before extending: `present` borrows `list`, so the phantoms must
-    // be materialized before we can mutate it.
+    let present_keys: HashSet<&str> = list.iter().map(|r| r.config_key.as_str()).collect();
+    let mut seen: HashSet<(String, DeviceKind)> = list
+        .iter()
+        .map(|r| (r.display_name.clone(), r.kind))
+        .collect();
+    // Collect before extending: `present_keys` borrows `list`, so the phantoms
+    // must be materialized before we can mutate it.
     let phantoms: Vec<DeviceRecord> = known
-        .filter(|(key, _)| !present.contains(key))
+        .filter(|(key, identity)| {
+            !present_keys.contains(key)
+                && seen.insert((identity.display_name.clone(), identity.kind))
+        })
         .map(|(key, identity)| offline_record(key, identity))
         .collect();
-    drop(present);
+    drop(present_keys);
     list.extend(phantoms);
 }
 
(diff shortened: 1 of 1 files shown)


buliwyf42:
Verified live on the rig (macOS 27 beta, MX Master 3S Easy-Switched to another host, on a Bolt receiver).

A note on the exact trigger I reproduced — it's slightly broader than the b034/b043 framing, and this fix covers both. With the 3S offline (on another host) it returns **no HID++ 2.0 model info**, so `build_device_list` surfaces it under the **fallback `config_key`** (`slot{N}` / `wpid…`) with its pairing-register codename, while the persisted identity sits under the model-derived key (`0b034`). Two different keys, same resolved name → two "MX Master 3S" cards. So any Easy-Switched / asleep device whose live key falls back hits this, not just multi-pid ones — and it recurs on every online→offline cycle.

Running a build of this branch against that exact state, the dedupe converges to one card:

```
inventory refreshed count=4 connected_keys=["0b367", "slot3", "3b042", "0b034"]   # both 3S
inventory refreshed count=4 connected_keys=[...]                                   # merge grace holds the stale placeholder 2 ticks
inventory refreshed count=3 connected_keys=["0b367", "slot3", "3b042"]             # 0b034 evicted → one 3S
```

The `(display_name, kind)` dedupe drops the `0b034` placeholder against the live `slot3` record; the ~6 s settle is just `INVENTORY_MISS_GRACE` aging out the initial-build placeholder. After that: a single 3S card, every cycle.

buliwyf42:
Good flag, but it doesn't apply here: `known_identities()` iterates `Config::devices`, which is a `BTreeMap<String, DeviceConfig>` — so iteration is sorted by `config_key` and fully deterministic across runs (the lexicographically-smallest key wins, e.g. `0b034` over `0b043`). The HashMap caveat you describe would be the concern; with the BTreeMap it isn't one.

buliwyf42:
Flagging a direct collision with #282 — these two PRs can't both land, and it's a design decision rather than a merge-conflict one.

This PR fixes a bug **inside** the persisted-identity / offline-placeholder subsystem (#224, shipped in v0.6.12): a multi-pid device showing twice in the carousel. #282 **removes that subsystem entirely** — `append_offline_known` and `offline_record` (the exact function this PR patches), `persist_identities` in `state.rs`, and `DeviceIdentity` / `device_identity()` / `set_device_identity()` / `known_identities()` in `config.rs` — because #280 wants a fully-disconnected device to drop out of the carousel instead of lingering as an offline card.

So the two encode opposite product decisions:

- **Keep asleep / Easy-Switched devices in the carousel** (the #224 behaviour, with this PR fixing its double-card bug), or
- **Drop disconnected devices out** (#280 / #282).

GitHub reports both as mergeable against `master` independently, but whichever lands first breaks the other: if #282 merges, this PR is moot — the duplicate-card path can't occur once there are no offline placeholders at all, and there's nothing left to rebase onto. If this merges first, #282 still deletes the whole thing.

Worth noting #282 also reverses #224, which is already released — so this is really a call on which direction the carousel should take. @AprilNEA, do you want sleeping/Easy-Switched devices to persist their cards, or to disappear when disconnected? Happy to close this if #280/#282 is the intended direction; if #224's behaviour stays, this fixes a real duplicate that surfaces on every online→offline cycle (verified live, details above).

buliwyf42:
Fresh repro on the **released v0.6.15** (not a dev build), so this ships in the current version. The carousel shows two `MX Master 3S` cards, both offline, under different keys:
SCR-20260621-sfyj

| Card | | Status |
|---|---|---|
| MX Mechanical Mini | Keyboard · slot 1 | 🟢 online · 100% |
| **MX Master 3S** | Mouse · **slot 3** | ⚪️ offline |
| MX Master 4 | Mouse · slot 4 | 🟢 online · 10% |
| **MX Master 3S** | Mouse · **slot 0** | ⚪️ offline (selected) |

Same physical mouse surfacing under two `config_key`s (`slot 3` vs `slot 0`), both offline — the `one_offline_model_persisted_under_two_pids_collapses_to_one_card` path this PR adds a test for. With the `(display_name, kind)` dedup these collapse to a single card.

(Screenshot from v0.6.15 to follow.)

davidbudnick:
> Fresh repro on the **released v0.6.15** (not a dev build), so this ships in the current version. The carousel shows two `MX Master 3S` cards, both offline, under different keys: SCR-20260621-sfyj
> 
> Card		Status
> MX Mechanical Mini	Keyboard · slot 1	🟢 online · 100%
> **MX Master 3S**	Mouse · **slot 3**	⚪️ offline
> MX Master 4	Mouse · slot 4	🟢 online · 10%
> **MX Master 3S**	Mouse · **slot 0**	⚪️ offline (selected)
> Same physical mouse surfacing under two `config_key`s (`slot 3` vs `slot 0`), both offline — the `one_offline_model_persisted_under_two_pids_collapses_to_one_card` path this PR adds a test for. With the `(display_name, kind)` dedup these collapse to a single card.
> 
> (Screenshot from v0.6.15 to follow.)

Thanks for flagging the issue, it mainly happens to me when moving from work to home, or quickly disconnecting the devices.

I'll update my branch as well.

buliwyf42:
Closing — superseded by v0.6.16.

`cf4033d` (`refactor(config): key settings by physical device`, shipped in v0.6.16) fixes this at the root: `config_key` is now derived from a stable physical identity (route + serial/unit) instead of the model-derived pid, and `append_offline_known` now blocks offline placeholders on **both** `config_key` and the new `model_key`. That's the same outcome this PR was after — a live record now suppresses the stale placeholder of the same physical device even when their pids differ — and the `model_key` approach is actually more precise than my `(display_name, kind)` dedup: it won't collapse two physically distinct devices of the same model, the trade-off I'd flagged as acceptable here.

Verified on v0.6.16: the carousel now shows a single card per device (the duplicate MX Master 3S is gone). No rebase needed — the bug is already gone upstream.

Thanks @AprilNEA. Closing as resolved.

PR #287, opened 2026-06-18
fix: make one-shot enumerate() retry transport-agnostic so Unifying partial drains recover

### LINKED ISSUE #277: One-shot enumerate() retry is a no-op for Unifying partial arrival-drain misses
## Summary

The one-shot `openlogi_hid::enumerate()` free fn — used only by the CLI (`openlogi list`, `openlogi diag *`) — retries through a transient probe miss (added in #237), but its early-exit gate never trips for a **Unifying** receiver. A Unifying probe that drops a device because its arrival event landed after the drain window reports `healthy = true`, so the retry loop returns the partial list immediately instead of retrying. #237's retry therefore covers the Bolt path (the #218 reporter's case) but is effectively a no-op on Unifying.

The continuous consumer (the GUI/agent inventory watcher) is **not** affected: it re-enumerates every ~2 s forever, so a single partial drain self-corrects on a later tick. This gap is specific to the single-shot CLI path.

## Mechanism

#237 made the one-shot `enumerate()` retry while any probed node is unhealthy:

https://github.com/AprilNEA/OpenLogi/blob/a82398665f9d7d27ce545d59b87df22c4d1c8205/crates/openlogi-hid/src/inventory.rs#L257-L290

The exit condition is `all_healthy || attempt >= ONESHOT_ATTEMPTS` ([`inventory.rs:271`](https://github.com/AprilNEA/OpenLogi/blob/a82398665f9d7d27ce545d59b87df22c4d1c8205/crates/openlogi-hid/src/inventory.rs#L271)), where `all_healthy` is the logical AND of each probed node's `probe.healthy`. The two receiver kinds define `healthy` very differently:

- **Bolt** ([`inventory.rs:537`](https://github.com/AprilNEA/OpenLogi/blob/a82398665f9d7d27ce545d59b87df22c4d1c8205/crates/openlogi-hid/src/inventory.rs#L537), `probe_bolt_receiver`): `healthy = pairing_count.is_some_and(|count| paired.len() == count)`. A device missing from the arrival drain lowers `paired.len()` below `count` → `healthy = false` → **the retry fires.** ✅
- **Unifying** ([`inventory.rs:622`](https://github.com/AprilNEA/OpenLogi/blob/a82398665f9d7d27ce545d59b87df22c4d1c8205/crates/openlogi-hid/src/inventory.rs#L622), `probe_unifying_receiver`): `healthy = pairing_count.is_some()`. Offline Unifying devices aren't enumerable yet (the `0xB5/0x5N` pairing-info sub-register format isn't resolved), so `paired.len() != count` is *expected* and cannot be the health signal. The arrival drain is the only device source, so a device whose arrival event arrives after `ARRIVAL_DRAIN` (1.5 s) is silently dropped — yet `pairing_count.is_some()` is still true → `healthy = true` → **the retry never fires.** ❌

So on Unifying, a single timing-induced partial drain in `openlogi list` / `diag` returns a short device set with no retry, and (because that partial list is written to the per-node ledger's `last_good`) it is returned as-is.

Originally flagged by Greptile as a P2 on #237 and acknowledged as a known limitation in the `probe_unifying_receiver` code comment.

## Impact

- Low. CLI-only; the GUI/agent path is unaffected. #218's reporter uses a Bolt receiver, so that issue's fix stands.
- Symptom: `openlogi list` / `openlogi diag *` occasionally under-reports devices on a Unifying receiver, with no automatic retry to recover within the single invocation.

## Proposed fix

Make the retry's early-exit transport-agnostic: stop when `all_healthy` **OR** the returned inventory is unchanged from the previous attempt.

- A Unifying partial drain → the missing device appears on a later attempt → the inventory *changes* → keep retrying until it stabilizes.
- Also removes a smaller wart: a chronically-unhealthy node (e.g. an asleep BT-direct device that never answers) keeps `all_healthy = false` forever, forcing every one-shot call to burn all `ONESHOT_ATTEMPTS` even when the receiver's devices read cleanly on attempt 1. A "stable inventory" stop returns as soon as the set stops growing.

Cost: needs `DeviceInventory` / `PairedDevice` equality + a stable ordering for the comparison, and (if `all_healthy` is dropped as the fast path) a minimum of two passes on the happy path. Keeping `all_healthy` as the fast-exit and adding "unchanged inventory" as the fallback stop preserves the one-pass happy path while closing the Unifying gap.

Rejected alternative: a Unifying-specific "online-device count vs a known-good baseline" health signal — stateful, needs persistence, and has its own flapping failure modes.

Refs #237, #222, #218.

mvanhorn:
## Summary

The one-shot `enumerate()` free fn (used by the CLI `list` / `diag` paths) retries through a transient probe miss, but its early-exit never tripped for a Unifying receiver, so a late-arriving device was silently dropped and the CLI returned a short device set.

The retry loop stopped on `all_healthy`, the AND of each probed node's `probe.healthy`. The two receiver kinds define `healthy` differently: a Bolt receiver sets `healthy = paired.len() == pairing_count` (a missing device makes it false and the retry fires), while a Unifying receiver sets `healthy = pairing_count.is_some()` because offline Unifying devices aren't enumerable yet. A Unifying device whose arrival event lands just after the ~1.5s drain window is dropped, yet `pairing_count.is_some()` stays true, so `all_healthy` stays true and the retry never fires. The continuous GUI/agent watcher is unaffected because it re-enumerates every ~2s.

Fixes #277.

## Why this matters

`openlogi list` and `openlogi diag` are the CLI's primary read paths, and a one-shot caller builds a fresh `Enumerator` with an empty ledger, so it has no prior snapshot to replay a transient miss against. The retry loop was added precisely to recover from that, but for a Unifying receiver it was a no-op: a device whose arrival event lands after the drain window is dropped while `pairing_count.is_some()` keeps `all_healthy` true, so the loop returns a short device set on the first pass. Users on Unifying hardware intermittently see fewer devices than are actually paired, with no recovery, while Bolt users (whose `healthy` rides on the count) are unaffected.

## Approach

Make the one-shot early-exit transport-agnostic by separating the retry signal from per-node ledger health:

- Add a `complete` flag to `NodeProbe`, reported alongside the existing `healthy`. For Unifying, `complete` requires `paired.len() == pairing_count`, so a partial drain now reports `complete = false` and the retry fires. `healthy` (which drives the ledger's per-node replay) is unchanged, so the watcher and the `#218` replay behavior are preserved.
- The retry loop stops on `all_complete`, on an unchanged inventory between two consecutive **healthy** passes (the expected stable Unifying offline-drain case, so a chronically-short set stabilizes instead of burning every attempt), or on the attempt cap. The unchanged-inventory shortcut is gated on `all_healthy` and the previous snapshot is only retained from a healthy pass, so a failed/timed-out/open-failed probe keeps using the full retry budget to reopen its channel and recover.
- Add `PartialEq` / `Eq` derives to the inventory device structs (`BatteryInfo`, `ReceiverInfo`, `DeviceModelInfo`, `DeviceTransports`, `PairedDevice`, `DeviceInventory`) so successive attempts can be diffed. These are additive derives only; field order is untouched, so the bincode wire format is unchanged (`openlogi-agent-core/tests/wire_format.rs` still passes).

As a side benefit, this also stops a chronically-unhealthy node (e.g. an asleep BT-direct device that never answers) from forcing every one-shot call to burn all attempts.

## Tests

Added unit tests in `openlogi-core` (structural equality across nested device fields) and `openlogi-hid`:

- Happy path: a complete inventory on attempt 1 returns after a single pass.
- Healthy partial drain stabilizes: the loop keeps retrying while the set grows, then stops once it stabilizes.
- Stabilization stop: a stable healthy-but-short set stops on "unchanged" rather than burning every attempt.
- Failed probe keeps retrying: an unchanged replay from an unhealthy pass does **not** stop early (regression guard).
- Bounded retries: a set that keeps changing still exits at the attempt cap.

`cargo fmt`, `cargo clippy`, and `cargo test` pass across the workspace.

### DIFF
--- a/crates/openlogi-core/src/device.rs
+++ b/crates/openlogi-core/src/device.rs
@@ -180,7 +180,7 @@ pub enum BatteryStatus {
 }
 
 /// Battery snapshot for one paired device, as last polled over HID++.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct BatteryInfo {
     /// Reported charge percentage (`0..=100`).
     pub percentage: u8,
@@ -193,7 +193,7 @@ pub struct BatteryInfo {
 /// Identity of an enumerated receiver — no paired-device state (that lives
 /// in [`DeviceInventory::paired`]). For a direct (Bluetooth/wired) device,
 /// a synthetic entry mirroring the device's own HID identity fills this role.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ReceiverInfo {
     /// Product string from the HID enumeration (e.g. `"Logi Bolt Receiver"`).
     pub name: String,
@@ -275,7 +275,7 @@ pub struct DeviceTransports {
 /// or a direct (Bluetooth/wired) attachment under its synthetic
 /// [`ReceiverInfo`]. Embedded in [`DeviceInventory`], so its field order is
 /// IPC wire format — see that type's contract.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PairedDevice {
     /// Receiver-assigned slot (1..=6 for Bolt).
     pub slot: u8,
@@ -310,7 +310,7 @@ pub struct PairedDevice {
 /// field and variant *order*, so reordering, retyping, or wrapping any field
 /// in this tree is a wire-format change and requires a `PROTOCOL_VERSION`
 /// bump (guarded by `openlogi-agent-core/tests/wire_format.rs`).
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct DeviceInventory {
     /// The receiver's identity — synthetic (mirroring the device itself)
     /// for a direct Bluetooth/wired attachment.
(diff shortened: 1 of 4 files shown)


mvanhorn:
Thanks again for #287 as well, @AprilNEA - transport-agnostic retry makes the Unifying partial-failure path much sturdier.

PR #294, opened 2026-06-20
feat(scroll): per-device inverted scrolling

### LINKED ISSUE #126: [Feature]: Inverted Scrolling Support
### Pre-flight checklist

### Problem / motivation

I would like the ability to set inverted scrolling on a per-device basis. The native MacOS setting (System Preferences > Mouse > Natural scrolling) applies to both trackpads and mice. However, I would like the ability to keep my trackpad with natural scrolling and use my mouse with "regular" scrolling.

### Proposed solution

Please add a "Natural scrolling" or "invert scroll behavior" configuration option for the mouse scroll wheel.

### Alternatives considered

_No response_

### Related area(s)

- [ ] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [x] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

### Additional context

Image

AprilNEA:
## Summary

Adds an opt-in, **per-device "Invert scroll direction"** toggle. A user can keep
macOS natural scrolling on the trackpad while running a traditional (reversed)
wheel on the mouse — which the native setting can't do.

Closes #126.

### Why the native option isn't enough

macOS scroll direction is a single global default (`com.apple.swipescrolldirection`);
the Mouse and Trackpad panes both reflect that *same* value. So "natural trackpad
+ reversed mouse" is impossible natively — exactly the case #126 describes, and the
reason tools like Scroll Reverser / LinearMouse exist. The toggle here inverts
*relative to* whatever the OS produces, so it composes correctly with the global
setting.

## Design — layered so each crate keeps its own job

- **openlogi-core** (data model): a per-device `invert_scroll` flag on `DeviceConfig`
  with accessors; omitted from `config.toml` when false.
- **openlogi-hook** (pure OS mechanism, no policy): `MouseEvent::Scroll` gains
  `is_continuous` (a trackpad/Magic-Mouse gesture vs. a discrete wheel detent), and
  a new `EventDisposition::ReplaceScroll` that re-emits the wheel with transformed
  deltas. Each platform's hook owns its own re-emission:
  - **macOS** mutates the `CGEvent` in place and scales its smooth-scroll companion
    fields (fixed-point + pixel deltas), so pixel-precise apps invert too — not just
    line-based ones.
  - **Linux** re-injects the flipped tick via uinput.
  - **Windows** re-injects via `SendInput`; the hook already drops `LLMHF_INJECTED`
    events, so the synthetic tick can't loop back.
- **openlogi-agent-core** (policy): the orchestrator publishes the active device's
  flag into a shared `AtomicBool`; the hook runtime inverts only a **discrete** wheel
  and passes continuous trackpad scrolling through untouched.
- **openlogi-gui**: a "Scrolling" card in the Pointer tab with an On/Off toggle
  (pure config — no hardware read). New strings translated across all 20 locales.

### Known limitation (documented in code)

The macOS `CGEventTap` reads the merged HID stream and can't attribute a scroll to a
physical device, so the flag applies to the **selected** device — the same
active-device compromise as `thumbwheel_sensitivity`. The trackpad-vs-wheel split is
done via `is_continuous`, which is what makes #126's actual scenario work. On Linux
(thread-per-device evdev) true per-device is structurally possible later.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean (macOS)
- `cargo clippy --target x86_64-pc-windows-gnu -p openlogi-hook -p openlogi-inject -p openlogi-agent-core --all-targets -- -D warnings` — clean (Windows code cross-linted)
- `cargo fmt --all --check` — clean
- Tests pass: openlogi-core, openlogi-hook, openlogi-agent-core (incl. `wire_format`
  — **no IPC protocol bump**, the flag travels via config + `ReloadConfig`), and
  openlogi-gui (incl. the i18n ordered-key parity test)

### DIFF
--- a/crates/openlogi-agent-core/src/hook_runtime.rs
+++ b/crates/openlogi-agent-core/src/hook_runtime.rs
@@ -7,6 +7,7 @@
 
 use std::cell::RefCell;
 use std::collections::BTreeMap;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, RwLock};
 
 use openlogi_core::binding::{
@@ -101,6 +102,7 @@ pub fn start(
     hooks: SharedHookMaps,
     dpi_cycle: Arc<RwLock<DpiCycleState>>,
     capture: CaptureChannel,
+    invert_scroll: Arc<AtomicBool>,
 ) -> Option<Hook> {
     if !Hook::has_accessibility() {
         warn!(
@@ -210,7 +212,26 @@ pub fn start(
             HOLD.with_borrow_mut(HoldState::cancel);
             EventDisposition::PassThrough
         }
-        MouseEvent::Scroll { .. } => EventDisposition::PassThrough,
+        MouseEvent::Scroll {
+            delta_x,
+            delta_y,
+            is_continuous,
+        } => {
+            // Invert only a discrete mouse-wheel tick, and only when the active
+            // device opted in (#126). A continuous trackpad / Magic Mouse gesture
+            // passes through so its native (e.g. natural) scrolling is left alone.
+            // The tap reads the merged HID stream and can't attribute a scroll to
+            // a physical device, so `invert_scroll` reflects the *selected*
+            // device — the same active-device compromise as `thumbwheel_sensitivity`.
+            if !is_continuous && delta_y != 0.0 && invert_scroll.load(Ordering::Relaxed) {
+                EventDisposition::ReplaceScroll {
+                    delta_x,
+                    delta_y: -delta_y,
+                }
+            } else {
+                EventDisposition::PassThrough
+            }
+        }
     });
 
     match result {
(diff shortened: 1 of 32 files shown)


AprilNEA:
Addressed the P1 in fdd916e — Windows now actually inverts instead of silently no-op'ing.

`translate_event` already drops `LLMHF_INJECTED` events, so re-injecting the flipped tick via `SendInput` can't loop back through the WH_MOUSE_LL hook. The dispatch now suppresses the original wheel and re-injects a transformed one, making Windows symmetric with Linux (uinput) and macOS (in-place CGEvent mutation). Verified with an `x86_64-pc-windows-gnu` cross-lint (`-D warnings`) plus a unit test on the `mouseData` reconstruction (one notch → 120, inverted → −120, exact negation).

PR #296, opened 2026-06-20
feat: add Capture Region to Clipboard button action

### LINKED ISSUE #288: [Feature]: Capture Region to Clipboard
### Pre-flight checklist

### Problem / motivation

Currently, you can configure the screenshot action, but it takes a full screenshot.

### Proposed solution

Add a new action to select "capture region to clipboard"

### Alternatives considered

_No response_

### Related area(s)

- [ ] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [ ] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

### Additional context

_No response_

mvanhorn:
## Summary

Adds a "Capture Region" button action that copies a selected screen region to the clipboard. It lives next to the existing full-screen "Screenshot" action in the System category and appears in the action picker.

## Why this matters

Issue #288 asks for a region-capture-to-clipboard action in addition to the full-screen screenshot. The existing `Screenshot` action only fires the full-screen capture, so there was no way to bind a button to grab a region. This adds that as a first-class action so users can wire a side button straight to a region selection that lands on the clipboard. Fixes #288.

## What changed and why

The action follows the same shape as `Screenshot`: a new `Action::CaptureRegion` variant in the System group, wired through `label()`, `category()`, `catalog()`, and the action-icon mapping (reusing the existing camera glyph, so no new asset is needed). A round-trip unit test documents that it serializes to the bare string `"CaptureRegion"`.

The platform behavior differs per OS because the region-to-clipboard chord differs. On macOS it sends Cmd+Shift+Ctrl+4, which is the interactive region selection routed to the clipboard rather than to disk. On Windows it sends Win+Shift+S, which opens Snip in region mode and copies the result. On Linux there is no universal region keysym, so it delegates to the desktop environment's screenshot handler the same way `Screenshot` does, and a code comment records that region versus full-screen depends on that handler.

The label is added to every locale file. The GUI i18n test requires each catalog action to have a translation in zh-CN, so leaving the other languages to fall back would break that test; a translation is provided for all 20 locales.

## Testing

`cargo test -p openlogi-core binding` passes 37 tests, including the new `binding_capture_region_roundtrips_as_single_string`. `cargo test -p openlogi-gui locale_file_resolves_keys` passes and confirms the locale coverage. `cargo fmt --check` is clean.

For manual verification, bind the action to a side button and fire it: the OS region picker should open and the capture should land on the clipboard. The macOS Cmd+Shift+Ctrl+4 path is the primary target of the issue.


AprilNEA (approved):

mvanhorn:
Appreciate this addition, @AprilNEA. A Capture Region to Clipboard button is a nice complement to the existing capture actions.

PR #300, opened 2026-06-21
feat(ipc): poll agent snapshots

AprilNEA:
## Summary
- add AgentSnapshot and append a v4 snapshot RPC to the tarpc Agent service
- return inventory health and inventory from one orchestrator-state read on the agent
- update the GUI IPC poll loop to use snapshot() instead of separate status()+inventory() RPCs
- pin the new protocol version and request variant in wire-format tests

## Validation
- cargo fmt --check
- cargo test -p openlogi-agent-core -p openlogi-agent -p openlogi-gui
- cargo clippy -p openlogi-agent-core -p openlogi-agent -p openlogi-gui --all-targets -- -D warnings
- pre-push cargo clippy --workspace --all-targets -- -D warnings

Stacked on #299.

### DIFF
--- a/crates/openlogi-agent-core/src/ipc.rs
+++ b/crates/openlogi-agent-core/src/ipc.rs
@@ -2,9 +2,9 @@
 //!
 //! tarpc generates the `AgentClient` and the `serve` glue from this trait. tarpc
 //! is strict request/response — no server push — so the streaming needs become
-//! polling: the GUI polls [`Agent::inventory`]/[`Agent::status`] on a timer, and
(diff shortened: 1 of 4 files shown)


AprilNEA:
Review follow-up: wrapped the IPC client module comment. I checked the suggested `PartialEq/Eq` derive for `AgentSnapshot`, but it does not compile because `DeviceInventory` intentionally does not currently implement equality; adding equality across the full device model would be a broader API change, so I left the snapshot derives unchanged.

PR #309, opened 2026-06-22
fix(unifying): enable wireless notifications so paired devices enumerate

laofun:
## Summary
A device on a Unifying receiver wouldn't show up even while connected. The
receiver only re-broadcasts arrival events while wireless notifications
(register 0x00) are on, and we never enabled it — so the trigger got ACK'd
but emitted nothing. Solaar does this before listing.

## Changes
- Enable wireless notifications before draining arrivals.
- Derive online from the feature walk succeeding (event.online was always
  "offline" here).
- Read device names at Unifying's 0x40 base instead of Bolt's 0x60, and
  clamp the reported length.
- Dedup repeated arrivals per slot.

## Testing
Tested on an MX Master 2S over a Unifying receiver. Added unit tests for the
name parser (normal, over-long length, short response).

### DIFF
--- a/crates/openlogi-hid/src/inventory/probe.rs
+++ b/crates/openlogi-hid/src/inventory/probe.rs
@@ -158,6 +158,19 @@ async fn probe_unifying_receiver(
     };
     debug!(events = connections.len(), "drained device-arrival events");
 
+    // The receiver can re-broadcast the same 0x41 for a slot more than once per
+    // trigger, so keep one connection per slot — otherwise the device is listed
+    // twice. Last write wins: a later event carries the freshest online flag.
+    let mut connections: Vec<_> = connections
+        .into_iter()
+        .map(|c| (c.index, c))
+        .collect::<HashMap<_, _>>()
+        .into_values()
+        .collect();
+    // HashMap iteration is unordered; sort by slot so the device list is stable
+    // across probe cycles instead of jittering.
+    connections.sort_by_key(|c| c.index);
+
     // Probe all online slots concurrently so a slow HID++ 2.0 feature walk on
     // one device doesn't push the next slot past the PROBE_BUDGET deadline.
     // Pass the receiver UID so each slot's cache key is scoped to this specific
@@ -427,6 +440,12 @@ async fn drain_device_arrival(bolt: &BoltReceiver) -> Vec<BoltDeviceConnection> 
 async fn drain_device_arrival_unifying(
     unifying: &UnifyingReceiver,
 ) -> Option<Vec<UnifyingDeviceConnection>> {
+    // The receiver only re-broadcasts 0x41 arrival events while wireless
+    // notifications are on; without this the trigger below is ACK'd but emits
+    // nothing, so a paired online device never surfaces.
+    if let Err(e) = unifying.set_wireless_notifications(true).await {
+        debug!(error = ?e, "enable wireless notifications failed");
+    }
     let rx = unifying.listen();
     if let Err(e) = unifying.trigger_device_arrival().await {
         debug!(error = ?e, "trigger_device_arrival failed; receiver may report no devices");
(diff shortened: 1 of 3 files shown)


laofun:
Addressed both review notes: the deduped connection list is now sorted by slot so device order is stable across probe cycles, and added a note on `InfoSubRegister::DeviceCodename` clarifying that `0x60` is the Bolt base while Unifying name reads go through the wire-verified `0x40` path in `inventory.rs`.

AprilNEA (commented):
This looks like the right direction overall — enabling the Unifying notification stream before `trigger_device_arrival`, correcting the 0x40 name base, and deduping repeated arrivals are all well-scoped, and CI is green.

One edge I’d tighten: `set_wireless_notifications(true)` currently writes the whole notifications register as `[0, 1, 0]`. That sets the wireless bit, but it also clears any other notification flags already enabled on register `0x00`. In this repo, the pairing flow writes `[0x00, 0x09, 0x00]` (`WIRELESS | SOFTWARE_PRESENT`) before streaming pairing events, so an inventory poll during/near pairing could accidentally drop the software-present bit or any future notification flag.

Could we make this read-modify-write, or at least use the same named notification flag constants as `pairing.rs` so the intended bitmask is explicit? The rest of the PR looks good to me.

laofun:
Good catch — switched `set_wireless_notifications` to read-modify-write: it now reads the notifications register first and only flips the WIRELESS bit, so `SOFTWARE_PRESENT` and any other flags are preserved.

chrematt:
Can confirm this fix works. Setup: MX Master 3 (wpid `4082`) on a Unifying receiver (`046d:c52b`), macOS 26 (Apple Silicon).

On stock v0.6.19 the receiver enumerates (`pairing_count=Some(1)`) but every probe cycle ends with `online devices differ from pairing count; … expected=1 found=0`, so the device never surfaces in the GUI (or flashes in briefly after pairing and disappears again). A direct HID++ probe against the receiver (enable wireless notifications on register `0x00`, then trigger a fake arrival via `0x02`) immediately yields a `0x41` arrival for slot 1 with the link established — the device was online the whole time; the agent just never turned notifications on. Exactly what this PR describes.

Note the branch no longer applies cleanly onto v0.6.19 (`inventory.rs` has since been split into `inventory/probe.rs` etc.), so I ported the changes onto the v0.6.19 tag to test. With the port applied, the slot enumerates stably across ticks as `online=true … codename=Some("MX Master 3")`, the name reads correctly via the Unifying `0x40` base, and the three parse tests pass. Happy to share the rebased version if that helps get this merged.

(Unrelated: occasional `device probe timed out — treating as a failed probe budget=5s` warnings remain, but that looks like #251 and the node ledger bridges them.)

rafaelje:
Confirmed that this fix works on my setup, GPT-5.6 SOL make it work:

  - macOS 26.5.1, Apple Silicon
  - OpenLogi v0.6.19
  - Logitech MX Master 3
  - Unifying receiver `046d:c52b`
  - Mouse WPID `4082`, slot 1

  With the stock v0.6.19 agent, every inventory cycle reported:

  receiver reports pairing count pairing_count=Some(1)
  drained device-arrival events events=0
  online devices differ from pairing count; offline devices not yet surfaced for Unifying expected=1 found=0

  The receiver knew that one device was paired, but trigger_device_arrival() produced no 0x41 events because wireless notifications had not been
  enabled. Consequently, the GUI briefly detected the mouse and then marked it offline.

  I ported this PR to the split inventory layout in the v0.6.19 tag. After enabling the WIRELESS bit in notification register 0x00 before triggering
  arrivals, every inventory cycle consistently reports:

  drained device-arrival events events=1
  unifying paired slot slot=1 online=true wpid=4082 kind=Mouse codename=Some("MX Master 3")

  The corrected Unifying codename register (0x40 + slot - 1) also returns MX Master 3 correctly. Control capture, gesture buttons, DPI buttons,
  thumbwheel handling, and the macOS event hook are active.

  The agent is running continuously through launchd, and the mouse no longer disappears from the GUI. The OpenLogi HID and HID++ test suites also
  pass: 212 tests passed, 0 failed.

  This confirms that the missing wireless-notification initialization is the root cause for this MX Master 3 / Unifying receiver combination.

PR #310, opened 2026-06-22
feat(gui): show connection-type icon on device cards

laofun:
## Summary
Hard to tell which device is Bolt, Unifying, or plain Bluetooth when you've
got a few in the list — they look the same. This adds a small glyph next to
the battery, picked from the device's route.

## Changes
- Pick an icon from `DeviceRoute`: bolt, unifying, or bluetooth.
- Three SVGs embedded alongside the existing action icons.

The bluetooth icon is lucide (already vendored). The bolt and unifying ones
are Logitech's marks — happy to swap for generic if you'd rather not.

## Testing
Added a unit test mapping each route to its icon. Verified in the gallery
with devices on both a receiver and direct Bluetooth.


laofun:
Made the `connection_icon_path` match exhaustive (`Some(Direct { .. }) | None`) so a new route variant trips the compiler, matching `route_label`. Left the bluetooth icon as a stroke glyph for now — swapping it to a filled variant is cosmetic and can follow if you prefer visual parity.

AprilNEA (commented):
Thanks — the code path itself looks clean to me: the route mapping is exhaustive, the assets are embedded in the same pipeline as the existing action icons, and CI is green.

One product/licensing note before merge: the new `bolt.svg` and `unifying.svg` are Logitech marks, while `app_assets.rs` currently documents `ACTION_ICONS` as vendored lucide icons under ISC. Since these SVGs would ship in the binary, I’d either:

- swap them for generic receiver / dongle glyphs, or
- keep the marks but make the asset licensing/source boundary explicit so they are not implicitly covered by the existing lucide/OpenLogi asset comments.

The contributor already called this out in the PR body, so this is more an owner decision than a code issue. If we’re comfortable shipping the marks, the implementation looks good.

AprilNEA:
Thanks for the clean implementation and for calling out the icon-source question. I’m going to merge this first so the route/icon UI work lands, then I’ll handle the asset/licensing cleanup separately on our side.

PR #312, opened 2026-06-22
feat(battery): support legacy 0x1000 BatteryStatus and its charging quirk

laofun:
Summary

Older mice like a Bluetooth-direct MX Master 2S only expose the legacy BatteryStatus feature (0x1000), never the unified 0x1004, so they showed no battery at all. This adds 0x1000 support and handles its quirk: the firmware reports 0% while charging.

Changes

- Implement 0x1000, mirroring the SmartShift 0x2110/0x2111 fallback. The inventory probe prefers 0x1004 and falls back to 0x1000. Coarse BatteryLevel from fixed buckets (0x1000 has no level bitmask).
- Hold the last-known % through a charge session so the reading does not flip to "Charging · 0%" once one discharge read exists.
- Cold start (charger plugged before the app opens, no cached %): show "Charging" instead of a bogus 0% across the card, summary, and tray menu.
- Add `openlogi diag battery` to print the raw 0x1004/0x1000 report.

Testing

- `cargo fmt --all -- --check`, `cargo clippy --workspace`, `cargo test` (bucket mapping, hold-% across all four branches, GUI label).
- On hardware: MX2S shows 50% discharging over BT-direct, and 0% status=Recharging while charging.


laofun:
Thanks @greptile-apps — applied all three:

1. `write.rs` fallthrough now reports `feature_hex: 0x1004` (the preferred feature) instead of implying `0x1000` was specifically absent.
2. `battery_feature_index` now `break`s past `u8::MAX` instead of `?`-returning `None` and discarding an already-found legacy probe.
3. Added `LegacyBatteryStatus::Other = 7` so a device reporting Solaar's value 7 surfaces as `BatteryStatus::Unknown` rather than failing the parse and making the battery vanish — covered by a new unit test.

AprilNEA (commented):
Thanks for the careful follow-up here — the 0x1000 fallback, diagnostic command, and regression tests all look well-scoped, and CI is green.

I found two edge cases worth tightening before merge:

1. `hold_percentage_while_charging` should probably be scoped to `BatteryProbe::Legacy`. The helper and comments describe the MX2S / `0x1000` charging quirk, but it is currently applied after `probe_features` and on cache refresh regardless of whether the selected probe is `Unified` or `Legacy` (`crates/openlogi-hid/src/inventory.rs`, around the calls after fresh reads). On a `0x1004` device that legitimately reports `charging_percentage = 0` while charging, or reports no percentage support, a previous non-zero cached value could be carried forward even though the legacy quirk is not involved. Matching on `BatteryProbe::Legacy` at the call site would keep the workaround tied to the feature that needs it.

2. The cold-start GUI suppression hides the `0%` text, but the detail summary progress bar still renders from `battery.percentage` (`relative_percent(0)` and `battery_color(0)`). That means the summary can still show an empty/critical-looking bar for the same “Charging, no reliable reading” case. It may be worth hiding the bar, using a neutral/charging style, or otherwise routing `battery_charging_no_reading(battery)` through the progress-bar rendering too.

Otherwise this is a nice, focused addition — especially the raw `openlogi diag battery` path for verifying the firmware behavior on real devices.

laofun:
Both addressed:

1. `hold_percentage_while_charging` now takes the `BatteryProbe` and early-returns unless it's `Legacy`, so the hold is tied to the `0x1000` quirk — a `0x1004` device reporting a genuine 0% while charging surfaces it. Added a `unified_charging_zero_is_not_held` test.
2. The summary progress bar now routes through `battery_charging_no_reading()` too: in the "charging, no reliable reading" case it leaves the track empty instead of drawing the 1%-wide critical sliver.

PR #341, opened 2026-07-01
feat(thumbwheel-tap): expose thumbwheel tap separately

### LINKED ISSUE #340: Thumbwheel rotation can trigger hidden App Exposé tap binding
## Problem

On an MX Master 3S, customizing thumbwheel sensitivity or rotation bindings diverts the wheel through HID++ `0x2150`. Setting both rotation directions to **Do Nothing** suppresses scrolling, but rolling and releasing the wheel can still trigger App Exposé.

## Reproduction

1. Select an MX Master 3S.
2. Set thumbwheel sensitivity away from the default, or customize either rotation direction.
3. Set **Thumb Wheel Up** and **Thumb Wheel Down** to **Do Nothing**.
4. Roll the thumbwheel and release it.

## Expected

Rotation is suppressed and no unrelated action fires.

## Actual

App Exposé can fire because the wheel reports a capacitive single-tap event separately from rotation.

## Root cause

- `ButtonId::Thumbwheel` represents the capacitive tap and defaults to `AppExpose`.
- The mouse model replaces the thumbwheel hotspot with Up/Down controls, so the tap binding is hidden and cannot be changed in the GUI.
- The HID capture path receives tap and rotation independently, but the hidden tap action still dispatches.
- Rotation diversion is incorrectly gated on the device advertising single-tap capability, preventing customized rotation on devices without tap support.

## Proposed fix

- Default thumbwheel tap to **Do Nothing**.
- Expose Up, Tap, and Down as separate non-overlapping GUI controls.
- Preserve explicit existing tap mappings.
- Divert rotation whenever thumbwheel capture is requested; use single-tap capability only to gate tap dispatch.
- Add core, projection, HID decode, capture-arming, geometry, and localization tests.

MoosaTae:
## What changed

- default the capacitive thumbwheel tap to **Do Nothing**
- expose separate Thumb Wheel Up, Tap, and Down controls in the mouse model
- migrate the historical pre-v4 `Thumbwheel = AppExpose` default away while preserving tap actions configured with the new schema
- keep rotation capture active on devices that do not advertise single-tap capability
- add translations, documentation, and focused regression tests

## Why

The MX Master 3S reports capacitive tap separately from rotation while the thumbwheel is diverted. OpenLogi hid that tap binding while defaulting it to App Exposé, so rolling and releasing the wheel could trigger App Exposé even when both rotation directions were configured as Do Nothing.

The capture setup also used single-tap support as a prerequisite for all thumbwheel diversion, unintentionally disabling customized rotation on devices without tap support.

## How to reproduce the original issue

1. Run an OpenLogi build before this MR with an MX Master mouse that exposes HID++ feature `0x2150` and supports capacitive single-tap.
2. Leave the hidden `Thumbwheel` binding at its historical default, **App Exposé**.
3. Set **Thumb Wheel Up** and **Thumb Wheel Down** to **Do Nothing**.
4. Roll the side thumbwheel and release/lift your thumb without intentionally tapping it.
5. Observe that App Exposé can open even though rotation is configured as Do Nothing. The GUI provides no separate tap control with which to discover or disable the triggering input.

After this MR, the mouse model shows **Thumb Wheel Up / Thumb Wheel Tap / Thumb Wheel Down** as separate controls. A fresh or migrated config leaves Tap at **Do Nothing**, so step 4 does not arm capture or dispatch App Exposé. Assigning an action to Tap intentionally enables capacitive-tap capture.

## Representative app log

Run the agent with debug tracing enabled (`OPENLOGI_LOG=debug`). Timestamps and module prefixes are omitted below; the messages and fields match the application tracing calls.

Before the fix, the hidden default could dispatch App Exposé:

```text
INFO  control capture active index=1 gesture=true dpi_buttons=0 thumbwheel=true
DEBUG HID++ button → action button=Thumbwheel action="App Exposé"
```

After the fix with Tap left at **Do Nothing**, the thumbwheel stays in native rotation mode when no sensitivity or rotation customization requires capture:

```text
INFO  control capture active index=1 gesture=true dpi_buttons=0 thumbwheel=false
```

After explicitly assigning **App Exposé** to **Thumb Wheel Tap**, capture and dispatch are expected:

```text
INFO  control capture active index=1 gesture=true dpi_buttons=0 thumbwheel=true
DEBUG HID++ button → action button=Thumbwheel action="App Exposé"
```

The exact `index`, `gesture`, and `dpi_buttons` values depend on the connected device and its configured controls.

## User impact

Users can configure rotation and tap independently. Tap is inert by default, while horizontal scrolling remains the default for rotation. The v4 migration removes the historical implicit App Exposé tap from older configs; tap actions intentionally selected after migration remain intact and functional.

## Validation

- `cargo test -p openlogi-core -p openlogi-hid -p openlogi-agent-core --lib` — 182 passed
- `cargo +1.94.0 test -p openlogi-gui` — 40 passed
- `cargo fmt --all --check`
- `git diff --check`
- release macOS app bundle built and codesign verification passed

Rust 1.94 was used for the GUI test and bundle because the current locked `smol_str 0.3.6` requires Rust 1.89+, while the local default toolchain is Rust 1.88.

Fixes #340

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- *(thumbwheel)* expose capacitive tap separately, default it to Do Nothing, and keep rotation capture available on devices without tap support
+
 ## [0.6.22](https://github.com/AprilNEA/OpenLogi/compare/openlogi-core-v0.6.21...openlogi-core-v0.6.22) - 2026-07-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- *(thumbwheel)* expose capacitive tap separately, default it to Do Nothing, and keep rotation capture available on devices without tap support
+- *(thumbwheel)* expose firmware-supported capacitive tap separately, default it to Do Nothing, hide it on unsupported devices, and keep rotation capture available without tap support
 
 ## [0.6.22](https://github.com/AprilNEA/OpenLogi/compare/openlogi-core-v0.6.21...openlogi-core-v0.6.22) - 2026-07-21
 
--- a/README.md
+++ b/README.md
@@ -92,9 +92,8 @@ Things OpenLogi does that Options+ won't:
 | Interface localization (20 languages: da, de, el, en, es, fi, fr, it, ja, ko, nb, nl, pl, pt-BR, pt-PT, ru, sv, zh-CN, zh-HK, zh-TW) | ✅ |
 | Linux packaging: udev rules, systemd unit, `.deb` / `.rpm` / `.pkg.tar.zst` | ✅ Linux |
 | Gesture-button per-direction bindings + live capture | ✅ (device capability dependent) |
-| Middle / mode-shift / thumbwheel button capture | ✅ middle on all platforms; mode-shift / thumbwheel device dependent |
+| Middle / mode-shift / thumbwheel tap + rotation capture | ✅ middle on all platforms; mode-shift / thumbwheel device dependent |
 | Windows (agent, GUI, event hook, installer) | ✅ Windows 11 hardware validated; newer port with ongoing compatibility polish |
-
 ¹ Media key actions use D-Bus MPRIS on Linux; a handful of macOS-specific actions have no universal Linux equivalent and are no-ops. Windows maps platform actions to native equivalents where available.
 
 ## Install
--- a/crates/openlogi-agent-core/src/bindings.rs
+++ b/crates/openlogi-agent-core/src/bindings.rs
@@ -161,6 +161,24 @@ mod tests {
     }
 
     #[test]
+    fn thumbwheel_tap_is_off_by_default_but_preserves_explicit_actions() {
+        let mut cfg = Config::default();
+        let projected = bindings_for(&cfg, Some("2b042"), None);
+        assert_eq!(projected.get(&ButtonId::Thumbwheel), Some(&Action::None));
+
+        cfg.set_binding(
+            "2b042",
+            ButtonId::Thumbwheel,
+            Binding::Single(Action::AppExpose),
+        );
+        let projected = bindings_for(&cfg, Some("2b042"), None);
+        assert_eq!(
+            projected.get(&ButtonId::Thumbwheel),
+            Some(&Action::AppExpose)
+        );
+    }
+
+    #[test]
     fn oshook_gestures_collects_only_os_hook_gesture_buttons() {
         let mut cfg = Config::default();
         // A gesture-mode Back (an OS-hook button) — included, raw map preserved.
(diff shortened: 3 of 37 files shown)


rajatkb:
Is this getting worked on ? Honestly this is the first time I come to know my mouse even supports this ! Does the MR keeps adds confguring this capacitive input ?

MoosaTae:
> **Legacy Tap Still Arms Capture**
> Older configs can contain an explicit `Thumbwheel = AppExpose` tap mapping while rotation bindings and sensitivity are still default.

Fixed in `5f18315` with a v4 config migration. When loading a pre-v4 config, OpenLogi removes only the historical `Thumbwheel = AppExpose` default, so that legacy value no longer arms diversion or changes native rotation handling. Other bindings are preserved.

The migration is intentionally version-gated: a user who explicitly configures `AppExpose` (or another tap action) in a v4 config still gets functional capacitive-tap capture. Changing `thumbwheel_armed` to ignore tap-only bindings globally would make the new tap control nonfunctional.

Regression coverage verifies both sides: legacy v3 defaults are removed, while intentional v4 tap bindings are preserved. `cargo test -p openlogi-core -p openlogi-hid -p openlogi-agent-core --lib` passes (182 tests), along with formatting and diff checks.

MoosaTae:
> **Does every thumbwheel really expose this tap input?**
> The UI must not present Thumb Wheel Tap merely because the asset has a thumbwheel marker; support has to come from the device-reported HID++ `0x2150` single-tap capability.

Fixed in `c6beeeb`. Inventory now reads `getThumbwheelInfo`, records `capabilities.single_tap` as `Capabilities::thumbwheel_tap`, persists it with device identity, and carries it over IPC (protocol v9). The UI always keeps thumbwheel Up/Down when a thumbwheel exists, but only adds Tap when the measured capability is true. The existing HID dispatch guard remains in place as defense in depth.

Debug builds now provide direct hardware evidence:

```text
DEBUG thumb wheel single-tap capability probed slot=1 supported=true
```

Unsupported devices report `supported=false` and do not show a Tap control. Capability probe failures also fail closed and are logged. Added geometry regression coverage for both supported three-control and unsupported rotation-only layouts. Core/HID/agent tests pass (189 tests, including IPC wire-format coverage), as do formatting and diff checks. The local GUI build remains blocked before crate compilation by the missing Xcode Metal Toolchain.

ron159:
This is a very annoying issue. I wonder when this PR will be merged.

nonstdout:
@ron159

You can temporarily disable until the fix lands, with:

`.config/openlogi/config.toml`

```
[devices."receiver:9f237ab95ad04c4d:slot:2".bindings]
Thumbwheel = "None"
ThumbwheelScrollUp = "VolumeDown"
ThumbwheelScrollDown = "VolumeUp"
```

HTH

AprilNEA:
Superseded by #857 — thanks for this, and for the write-up in #340. That issue is what made the whole cluster tractable: it named the three-way interaction (invisible binding + no GUI hotspot + divert armed by unrelated settings) that
seven separate reports had only described from the outside.

What landed from here: the tap seeds to `Action::None`, along with your point about preserving explicitly configured taps — which turned out to need no migration at all, since defaults are never persisted and the only `Thumbwheel` entries on disk are hand-written ones. 

You're credited as co-author. Thanks!

Issue #336, opened 2026-06-30
Bluetooth-direct devices don't appear + no diagnostic output: missing Input Monitoring grant + launchd agent logs nowhere


MichaelDanCurtis:
## Summary

A fresh install with a normal macOS setup shows no devices in the GUI, with
**zero diagnostic output** explaining why. Root cause turned out to be a
missing **Input Monitoring** privacy grant to the background agent — a
permission that is (a) separate from Accessibility, (b) never requested or
prompted, and (c) undocumented. Compounding it, the launchd-managed agent
runs with **no log redirection**, so its `failed to open HID++ channel`
warnings go nowhere — the user has no way to discover the cause without
running the binary by hand with `OPENLOGI_LOG=debug`.

Filing this so the next person doesn't spend an hour guessing.

## Environment

- macOS 15 (darwin 25.5.0, arm64)
- Devices: **MX Master 3S** + **MX KEYS S**, both paired **Bluetooth-direct
  (BLE)** — no Logi Bolt/Unifying receiver, nothing wired
- Logi Options+: **not installed** (so receiver contention is ruled out)
- OpenLogi 0.6.18 via Homebrew cask
- Accessibility: granted to OpenLogi Agent

## Symptom

GUI opens, but the device carousel is empty — neither device appears, so no
buttons can be configured. Both devices are connected to the Mac over
Bluetooth (confirmed in System Settings → Bluetooth). No error, no prompt,
nothing in any log.

### 1. Missing Input Monitoring grant (the actual blocker)

OpenLogi's device discovery (`openlogi_hid::transport::enumerate_hidpp_devices`)
**does** see BLE-direct devices — the debug log shows them matching the
`0xFF43 / 0x0202` HID++ long-report collection:

```
DEBUG logitech HID node name=MX Master 3S  pid=b034 usage_page=0xff43 usage_id=0x0202 matched=true
DEBUG logitech HID node name=MX KEYS S MAC pid=b37c usage_page=0xff43 usage_id=0x0202 matched=true
DEBUG HID++ candidate interfaces count=2
WARN  failed to open HID++ channel — retrying next tick  error=Message("Failed to open device")
WARN  failed to open HID++ channel — retrying next tick  error=Message("Failed to open device")
```

Enumeration works; the `IOHIDDeviceOpen` call fails for both, every 2s tick,
with `"Failed to open device"`. On macOS under the Hardened Runtime, that
maps to a denied TCC grant — specifically **Input Monitoring**
[...]

PR #342, opened 2026-07-01
fix(openlogi): stabilize bluetooth direct devices

toby1991:
Filter non-identifying direct HID++ model info before it becomes a persistent device identity, and keep Bluetooth-direct offline placeholders on the HID++ self slot instead of slot 0.

Treat no-evidence direct refreshes for cached devices as transient misses so a slow or sleeping BLE device replays last-good inventory instead of being cleared.

Required macOS privacy grants for packaged installs:

- Input Monitoring: OpenLogi GUI app bundle at /Applications/OpenLogi.app (binary: /Applications/OpenLogi.app/Contents/MacOS/openlogi-gui).

- Input Monitoring: OpenLogi Agent helper bundle at /Applications/OpenLogi.app/Contents/Library/LoginItems/OpenLogiAgent.app (binary: /Applications/OpenLogi.app/Contents/Library/LoginItems/OpenLogiAgent.app/Contents/MacOS/openlogi-agent).

- Accessibility: OpenLogi Agent helper bundle at /Applications/OpenLogi.app/Contents/Library/LoginItems/OpenLogiAgent.app (binary: /Applications/OpenLogi.app/Contents/Library/LoginItems/OpenLogiAgent.app/Contents/MacOS/openlogi-agent), because the agent owns the CGEventTap mouse hook.

Remove the temporary async-hid path patch/vendor tree and raw HID transport trace from the final change; the fix now stays inside OpenLogi's probe and GUI state logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- *(hid)* keep macOS Bluetooth-direct Logitech devices discoverable when HID++ optional reads are slow.
+- *(gui)* avoid stale Bluetooth-direct `unit:*` device cards and render direct offline placeholders as slot 255.
+
 ## [0.6.22](https://github.com/AprilNEA/OpenLogi/compare/openlogi-core-v0.6.21...openlogi-core-v0.6.22) - 2026-07-21
 
 ### Added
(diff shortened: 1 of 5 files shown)


toby1991:
image
image
https://github.com/AprilNEA/OpenLogi/issues/336

AprilNEA:
Thanks for the investigation. I’m closing this version because it is stale against current `master`, its current Clippy checks are failing, and the broad cache/identity workaround does not address the transport and retry failures now tracked in #798 and #781. The proposed implementation also still drops kind-only direct-device evidence and can conflate two identical physical devices during stale-card suppression.

The useful findings are worth carrying forward: reject semantically empty model-information payloads, bound optional BLE reads, preserve known identity across transient no-evidence refreshes, and key offline deduplication by stable physical identity. These should be acceptance criteria for focused fixes around #798/#781; #880 tracks a current affected device.

PR #344, opened 2026-07-01
feat: add function key remapper

MichaelDanCurtis:
## Summary
- add a keyboard Keys tab with staggered Esc/F1-F19 callout bubbles, centered key markers, hover/selection highlighting, and a scrollable action picker panel
- hide the legacy Buttons tab for keyboards; function-key remapping now lives only under Keys while mouse/trackball devices keep Buttons
- persist global keyboard function-key bindings in config and wire them through the agent hook runtime so bound keys dispatch actions and suppress the original key event
- add power-user picker/editor support for Type Text, Run AppleScript, Run Shell Command, and Workflow actions, reusing the mouse action picker style
- bump the GUI/agent IPC protocol so the new GUI cannot silently drive an older installed agent without function-key dispatch
- isolate local macOS bundle builds under `org.openlogi.openlogi.dev` / `org.openlogi.agent.dev` before signing, and route those builds through the `openlogi-dev` XDG profile so dev agents cannot occupy the production socket, lock, config, or asset cache
- remove the stale cargo-bundle DMG generated before helper embedding/signing; `macos package` remains the production-ID DMG path

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p xtask`
- `cargo test -p openlogi-agent-core --test wire_format`
- `cargo test -p openlogi-gui`
- `cargo test --workspace`
- `cargo run -p xtask -- macos bundle`
- verified keyboard tab regression: asset-backed keyboards with remappable keys show `Keys`, `Lighting`, and `Device`, but not `Buttons`
- verified local bundle Info.plists are `OpenLogi Dev` / `OpenLogi Agent Dev` with `.dev` bundle identifiers
- verified rebuilt local app/helper are signed with `Apple Development: michaeldancurtis@gmail.com (P7R2A675CN)` and pass `codesign --verify --strict`
- verified `target/release/bundle/dmg/OpenLogi.dmg` is removed after `macos bundle`, avoiding the pre-helper cargo-bundle artifact
- killed the mixed `/Applications` GUI + local helper state, launched the exact rebuilt dev app/helper path, and verified both running processes come from `target/release/bundle/osx/OpenLogi.app`
- verified the running dev helper listens on `~/.config/openlogi-dev/agent.sock`, the GUI process reports `org.openlogi.openlogi.dev`, TCC has an allowed `org.openlogi.agent.dev` row, and the visible app shows Accessibility granted instead of the background-service failure


recchia:
Heads up @MichaelDanCurtis — #191 also touches `crates/openlogi-gui/src/main.rs` (adds Wayland `app_id` handling + a couple of new `mod` declarations near the top of the file). The overlap looks minor (your change just adds `mod keyboard_model;` to the same block), so it should merge cleanly either way, but flagging it in case ordering matters when one of us rebases.

Issue #100, opened 2026-06-03
M500S wheel tilt (left/right) buttons not detected


cedric-anamorphik:
## Bug: Wheel tilt buttons not detected (M500S)

The horizontal scroll / wheel tilt buttons (wheel left and wheel right) on the **Logitech M500S** are not detected by OpenLogi. 
They do not appear in the button list and cannot be assigned any action.

### Steps to reproduce
1. Connect a Logitech M500S
2. Open OpenLogi
3. Tilt the scroll wheel left or right
4. No button event is registered

### Expected behavior
Wheel tilt left and wheel tilt right should be detected as distinct buttons and be configurable like any other mouse button.

### Environment
- Mouse: Logitech M500S
- OS: Mac OS 15.7.3 (24G419) (Intel)
- OpenLogi version: 0.4.0

g30r93g:
Same for MX Ergo

PR #359, opened 2026-07-06
Add remappable support for horizontal scroll wheel tilt actions

srako:
…buttons

Add WheelLeft and WheelRight to the ButtonId enum so the main scroll wheel's horizontal tilt — which arrives as a horizontal scroll event at the OS hook, not as a button press — can be intercepted and remapped to any Action, just like a physical button.

Core changes:
- binding.rs: WheelLeft/WheelRight variants, ALL array, labels, and default bindings (native horizontal scroll, pass-through)
- hook_runtime.rs: intercept MouseEvent::Scroll with non-zero delta_x from a non-trackpad source; look up the bound action and dispatch it with a 250ms per-direction cooldown so one deliberate tilt fires once
- mouse_buttons.rs: GUI hotspots flanking the middle-click area
- geometry.rs: with_wheel_tilt() auto-derives tilt hotspots from the MiddleClick position; default_labels() updated for the new buttons

Default behavior is unchanged: unbound or default-bound tilt events pass through as native horizontal scroll. Only when the user maps WheelLeft/WheelRight to a non-default action does the hook suppress the scroll event and fire the bound action instead.

Related to #100

### DIFF
--- a/crates/openlogi-agent-core/src/hook_runtime.rs
+++ b/crates/openlogi-agent-core/src/hook_runtime.rs
@@ -8,6 +8,7 @@
 use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::sync::{Arc, RwLock};
+use std::time::{Duration, Instant};
 
 use openlogi_core::binding::{
     Action, ButtonId, GestureDirection, SwipeAccumulator, default_binding,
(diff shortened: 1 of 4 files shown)


tomByrer:
Any progress on this?

kdeldycke:
Data relevant to the layer choice, since this and #357 intercept the same two controls at different points. Source: an MX Ergo's `0x1B04` getCidInfo, via Logi Options' HID++ response cache.

```
idx  CID   TID   flags  decoded
6    0x5B  0x3F  0x71   mouse|reprog|divertable|persist
7    0x5D  0x40  0x71   mouse|reprog|divertable|persist
```

Tilt is a pair of discrete divertable controls on this device, with the same flags as the already-bindable `0x52`, `0x53`, `0x56` and `0xED`.

The `delta_x` observation in this PR and the CID above are consistent: the factory assignment of `0x5B`/`0x5D` is AC Pan, the HID horizontal-scroll usage. Logi Options names them `LeftScrollAsAcPan` / `RightScrollAsAcPan`. Tilt therefore reaches the OS hook as horizontal scroll only while the CID is undiverted; diverting it suppresses the scroll event.

Two consequences for the hook approach:

1. Diversion yields direction exactly; `delta_x` interception infers it, which is what the 250ms per-direction cooldown compensates for. Devices with both a tilting main wheel and a thumbwheel emit horizontal scroll from both.
2. The non-trackpad source test is worth checking against a Bluetooth-direct connection, where the same unit presents as a different HID source.

A hook path still covers devices whose tilt exposes no CID, so both layers may be wanted, preferring diversion when `0x1B04` advertises the control.

PR #362, opened 2026-07-07
fix: embed and sign the openlogi CLI in the macOS app bundle

### LINKED ISSUE #328: [Bug]: CLI missing from MacOS release
### Pre-flight checklist

### Which part of OpenLogi?

CLI (the `openlogi` binary)

### OpenLogi version

0.6.17

### Operating system

macOS

### OS version & architecture

macOS 15.7.8 (arm64)

### Device model

MX Master 3S

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

My MX Master isn't showing up, so I went to try and run 'openlogi list' but can't find the binary anywhere. I tried both via the direct download, as well as the homebrew cask, but it doesn't appear that the CLI is included in the app bundle :\

### Steps to reproduce

1. `brew install --cask openlogi`
2. In new terminal session:
    ```bash
    bash-5.3$ openlogi
    bash: openlogi: command not found
    ```
    ```bash
    find /Applications/OpenLogi.app -name "openlogi*"
    /Applications/OpenLogi.app/Contents/MacOS/openlogi-gui
    /Applications/OpenLogi.app/Contents/Library/LoginItems/OpenLogiAgent.app/Contents/MacOS/openlogi-agent
    ```

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.6.17 (release)
- Agent: v0.6.17 (connected)
- IPC protocol: GUI 8 / agent 8
- Inventory: ready
- OS: macOS 15.7.8 (arm64)
- Launch at login: yes · Menu bar: yes · Update check: on
- Config: schema 3 · 0 configured device(s) · thumbwheel 14

**Assets**
- Cache path: ~/.local/share/openlogi/assets · Bundle assets: absent

**Devices (0)**
- No devices detected.
```

### `openlogi list` output

```text

```

### Logs

```shell

```

mvanhorn:
## Summary

Fixes #328. The macOS `.app` bundle now includes the `openlogi` CLI binary, so it is present in the release the same way it is in the Linux package and where the Homebrew cask's `binary` stanza expects it (`Contents/MacOS/openlogi`).

## Why this matters

The macOS bundle step only embedded `openlogi-gui`, so the CLI was absent from the shipped `.app`. Users installing the macOS release had no `openlogi` command, while Linux users did. The fix mirrors the existing agent-helper embedding so the CLI ships alongside the GUI.

## Changes

- `xtask/.../macos/bundle.rs`: add `embed_cli`, which builds `openlogi --release` and copies it into `Contents/MacOS/openlogi`, mirroring `embed_agent_helper`. Add `verify_bundle_binaries`, called at the end of `run()`, asserting both `openlogi` and `openlogi-gui` are present so a future regression fails the build loudly (unit-tested against a temp dir, no Xcode toolchain required).
- Signing: the embedded CLI is a second Mach-O under `Contents/MacOS`, so it needs its own hardened-runtime signature before the outer app or notarization rejects its as-built ad-hoc signature. `sign_app` now signs (and verifies) the CLI in the same inside-out order it uses for the agent helper, and the release workflow's signing step does the same, so tagged notarized builds stay valid.

## Testing

`cargo build -p xtask`, `cargo test -p xtask` (5 passing, including the new `verify_bundle_binaries` cases), and `cargo clippy -p xtask` all pass. Running `xtask macos bundle` locally produces an `OpenLogi.app` containing both `openlogi` and `openlogi-gui`. The codesign/notarization steps run only in the tagged release workflow (they need the signing certificate), so those were reviewed against the existing agent-helper signing pattern rather than executed here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,6 +115,7 @@ jobs:
           set -euo pipefail
           app="target/release/bundle/osx/OpenLogi.app"
           helper="$app/Contents/Library/LoginItems/OpenLogiAgent.app"
+          cli="$app/Contents/MacOS/openlogi"
           # Inside-out signing: seal the nested agent helper with its own
           # signature first, then the outer app (which seals the signed helper).
           # --deep is deprecated and can't give the helper an independent
@@ -124,10 +125,18 @@ jobs:
             codesign --force --options runtime --timestamp \
               --sign "$APPLE_SIGNING_IDENTITY" "$helper"
           fi
+          # The embedded CLI is a second Mach-O under Contents/MacOS; sign it
+          # with the hardened runtime before the outer app or notarization
+          # rejects its as-built ad-hoc signature.
+          if [ -f "$cli" ]; then
+            codesign --force --options runtime --timestamp \
+              --sign "$APPLE_SIGNING_IDENTITY" "$cli"
+          fi
           codesign --force --options runtime --timestamp \
             --sign "$APPLE_SIGNING_IDENTITY" "$app"
           codesign --verify --strict --verbose=2 "$app"
           [ -d "$helper" ] && codesign --verify --strict --verbose=2 "$helper"
+          [ -f "$cli" ] && codesign --verify --strict --verbose=2 "$cli"
 
       - name: Create and sign DMG
         env:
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for taking both of mine, @AprilNEA - the signed CLI in the app bundle closes out the macOS story.

PR #363, opened 2026-07-07
Back and Forward mouse buttons now work in Safari and Chrome

b0x42:
## What this fixes

If you have a Logitech MX Vertical (or any Logitech mouse with dedicated Back/Forward buttons), those buttons previously did nothing in Safari and other browsers on macOS — even though Chrome eventually worked. This PR fixes that: pressing the Back button navigates back in both Safari and Chrome, and Forward navigates forward.

---

## Why it didn't work

Logitech mice don't send Back/Forward as ordinary mouse button clicks. They use a proprietary HID++ firmware protocol to report those buttons, and macOS never translates them into the standard mouse events that apps like Safari listen for. OpenLogi's existing input hook only captured standard mouse events, so it never saw Back/Forward presses at all.

On top of that, **Safari has a special restriction**: even when you synthesize the correct mouse button events or keyboard shortcuts in software, Safari's web engine (WKWebView) runs in a sandboxed subprocess and ignores them entirely. Chrome doesn't have this restriction, which is why Chrome worked once the buttons were being captured — but Safari didn't.

There was also a subtle timing problem: the MX Vertical's buttons are visible at two levels simultaneously — as standard OS mouse button events (intercepted by OpenLogi's CGEventTap hook) and as HID++ firmware events. The hook path was firing first and sending `Cmd+[` to Safari (which Safari ignores), while the correct AX path was running too late.

---

## How it's fixed

**Capturing the buttons (works for all apps)**

The fix adds the Back and Forward button IDs to OpenLogi's HID++ capture session — the same mechanism already used for the DPI button and gesture controls. When you press Back or Forward, the firmware event is now intercepted, identified, and forwarded into OpenLogi's remapping pipeline.

A 350ms debounce prevents the MX Vertical's firmware from registering a single physical click as multiple rapid presses.

**Making Safari work**

Since Safari ignores synthetic input events, the fix uses macOS's Accessibility API instead. Rather than injecting a fake button press, OpenLogi finds Safari's actual Back/Forward toolbar button and programmatically clicks it — the same way a user would with a mouse cursor. Safari responds to this correctly.

For Chrome and other browsers, the fix continues to use keyboard shortcuts (`Cmd+[` / `Cmd+]`), which those apps handle natively.

**Fixing the timing race**

The frontmost app's process ID is captured at the exact moment you press the physical button — on the CGEventTap callback thread, where macOS guarantees the app context is current. This ensures OpenLogi always acts on the right browser window, whether the event comes through the hook path or the HID++ path.

---

## Files changed

| Crate | What changed |
|-------|-------------|
| `openlogi-hid` | Added Back/Forward CID constants; capture and debounce button presses; embed frontmost app PID in button events |
| `openlogi-inject` | New `ax_navigate_browser()` — finds and presses Safari's toolbar button via Accessibility API; falls back to keyboard shortcut for other browsers |
| `openlogi-agent-core` | Route Back/Forward through AX path first (in both the CGEventTap hook and the HID++ gesture watcher), keyboard shortcut second |

---

## Technical details

<details>
<summary>HID++ CIDs for Back/Forward</summary>

The MX Vertical uses CIDs `0x0053` (Back) and `0x0056` (Forward) — below the documented `0xB8` range in Logitech's control-ID list. This PR also adds the higher-range MultiPlatform CIDs (`0x00BD`, `0x00CE`, `0x00DB` for Back; `0x00CF` for Forward) used on other Logitech models.

</details>

<details>
<summary>Why the Accessibility approach for Safari</summary>

Safari's WKWebView runs in a separate sandboxed process. CGEvents posted at the HID or Session tap level don't reach it — this is by design. Tools like Karabiner-Elements and Mac Mouse Fix that support Safari use a DriverKit virtual HID device (kernel-level injection), which requires a special Apple entitlement. The AX approach achieves the same result without requiring a kernel extension.

The AX tree search skips `AXSplitGroup`, `AXTabGroup`, `AXOpaqueProviderGroup`, and `AXRadioButton` subtrees to avoid exhausting search depth on Safari's tab bar.

</details>

<details>
<summary>The dual-path timing issue</summary>

The MX Vertical's buttons are visible at two layers simultaneously: as `OtherMouseDown` OS events (CGEventTap hook) and as HID++ DivertedButtons CID events (gesture capture session). The hook path was firing first and sending `Cmd+[` to Safari (ignored). Now both paths capture `NSWorkspace.frontmostApplication.processIdentifier` on the tap callback thread and call `AXPress` on the app's toolbar button directly.

</details>

<details>
<summary>Notes for reviewers</summary>

- `CapturedInput::ButtonPressed` now carries `#[serde(skip)] Option<i32>` (frontmost PID). Skipped in serialization — dispatch hint only, not part of the wire format.
- The 350ms debounce is tuned for the MX Vertical. Other devices may tolerate a tighter window.
- Long-term, a DriverKit virtual HID device would be more robust. The AX path is pragmatic and avoids requiring Apple entitlements.

</details>

---

## Testing

Tested on MX Vertical (046d:b020) over Bluetooth on macOS:
- **Safari** ✅ — Back/Forward navigate browser history
- **Chrome** ✅ — Back/Forward navigate browser history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5132,6 +5132,8 @@ dependencies = [
  "futures-concurrency",
  "futures-lite",
  "num_enum",
+ "objc2 0.6.4",
+ "objc2-app-kit 0.3.2",
  "openlogi-core",
  "openlogi-hidpp",
  "serde",
--- a/crates/openlogi-agent-core/Cargo.toml
+++ b/crates/openlogi-agent-core/Cargo.toml
@@ -29,3 +29,4 @@ workspace = true
 bincode = "1.3"
 # Only to construct an `InventoryError::Hid` in the watcher's classify tests.
 async-hid = { workspace = true }
+
--- a/crates/openlogi-agent-core/src/hook_runtime.rs
+++ b/crates/openlogi-agent-core/src/hook_runtime.rs
@@ -7,7 +7,8 @@
 
 use std::cell::RefCell;
 use std::collections::BTreeMap;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, PoisonError, RwLock};
+use std::time::{Duration, Instant};
 
 use openlogi_core::binding::{
     Action, ButtonId, GestureDirection, SwipeAccumulator, default_binding,
@@ -265,6 +266,40 @@ fn is_native_click(id: ButtonId, action: &Action) -> bool {
     )
 }
 
+/// Minimum time between two BrowserBack (or two BrowserForward) keyboard
+/// dispatches, shared across the CGEventTap hook and the HID++ gesture
+/// watcher — both call [`dispatch_action`] independently, and on devices
+/// where one physical press is visible through both paths, a naive dispatch
+/// would fire the keyboard shortcut twice for one click. Same window as the
+/// HID++ path's own intra-press debounce (`BACK_FORWARD_DEBOUNCE` in
+/// `openlogi-hid`), for consistency.
+const BROWSER_NAV_DEBOUNCE: Duration = Duration::from_millis(150);
+
+/// Per-direction last-dispatch timestamps backing [`browser_nav_debounce_ok`].
+/// `(last_back, last_forward)`.
+static BROWSER_NAV_LAST: Mutex<(Option<Instant>, Option<Instant>)> = Mutex::new((None, None));
+
+/// Whether a BrowserBack/BrowserForward keyboard dispatch for `action` should
+/// proceed, or be suppressed as a duplicate of one already sent (from either
+/// dispatch path) within [`BROWSER_NAV_DEBOUNCE`]. Records the dispatch time
+/// on every `true` return so the *next* call — from either path — sees it.
+fn browser_nav_debounce_ok(action: &Action) -> bool {
+    let mut last = BROWSER_NAV_LAST
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner);
+    let slot = if matches!(action, Action::BrowserForward) {
+        &mut last.1
+    } else {
+        &mut last.0
+    };
+    let now = Instant::now();
+    let fire = slot.is_none_or(|t| now.duration_since(t) >= BROWSER_NAV_DEBOUNCE);
+    if fire {
+        *slot = Some(now);
+    }
+    fire
+}
+
 /// Route a bound action either to OS-level event synthesis
 /// ([`Action::execute`]) or to one of OpenLogi's hardware-side handlers.
 ///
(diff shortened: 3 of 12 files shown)


b0x42:
Thanks for the review! All 5 issues addressed in the latest commit:

1. **Localized AX descriptions** — now checks `AXIdentifier` first (`BackForwardToolbarButton_Back`/`BackForwardToolbarButton_Forward`, locale-independent), falls back to English description for older Safari versions.

2. **Partial arm leaves diversion enabled** — Back/Forward CID diversion loops now roll back any already-diverted CIDs before propagating an error.

3. **Dual input paths can repeat** — documented as a known non-issue: Safari only responds to AXPress (no double-navigation), Chrome only responds to keyboard shortcuts (idempotent). Added a TODO for proper cross-path suppression.

4. **Debounce swallows real clicks** — reduced from 350ms to 150ms. This covers the MX Vertical's intra-press bounce window (~50-100ms) while allowing intentional rapid double-clicks (typically ≥200ms apart).

5. **CFArray leak** — `find_button` now `CFRetain`s the found button before returning, then immediately `CFRelease`s each `children_val` after the loop. The caller receives a +1 retained pointer and releases it after `AXPress`.

b0x42:
Addressed the 3 remaining issues:

1. **Rollback misses DPI** — the error path now rolls back gesture, DPI, Back, and Forward CIDs (all previously-diverted controls) before propagating the error.

2. **Duplicate AX navigation** — removed the AX call from the hook path entirely. The hook now uses only `Cmd+[`/`Cmd+]` (keyboard shortcut for Chrome). Safari is handled exclusively by the HID++ gesture watcher, which calls `ax_navigate_browser` with the PID captured at press time. The two paths no longer both call `AXPress`.

3. **Localized fallback still fails** — the button search now tries three locale-independent signals in order before falling back to the English description:
   - `AXIdentifier` (`BackForwardToolbarButton_Back`/`_Forward`) — primary, locale-independent
   - `AXSubrole` (`AXBackForwardButtonBack`/`AXBackForwardButtonForward`) — secondary, locale-independent
   - `AXDescription` English text — last resort for older Safari/macOS versions

b0x42:
Fixed: **Later Arm Failure Leaks Diversions**

Changed `tw.set_reporting(true, false).await?` to a `match` that logs a warning and continues (skipping thumbwheel click capture) rather than propagating the error. This means a thumbwheel setup failure no longer aborts the session, so already-diverted Back/Forward and DPI controls are always returned inside `ArmedControls` where `disarm()` can restore them on teardown.

b0x42:
All 3 issues fixed:

**1. Recursive Retain Leak** — moved `CFRetain` to the leaf (inside the three button-match checks), removed the `map(|el| CFRetain(el))` in each parent frame's propagation path. The button is now retained exactly once regardless of tree depth.

**2. Earlier Diversion Leaks** — the gesture and DPI CID diversion loops now use explicit error handling with rollback (matching the Back/Forward pattern already in place). A DPI failure rolls back gesture; a gesture failure has nothing prior to roll back.

**3. Localized Fallback Fails** — added `find_nav_button_by_position()` as a structural fallback when identifier/subrole/description all fail. It walks `AXWindow → AXToolbar → AXGroup (nav group) → AXGroup → AXButton[0 or 1]` by role and position, which is locale-independent and stable across Safari versions.

davidbudnick:
@b0x42 Hello, thanks for taking the time to submit a change request, looks as you have some Github actions which are failing. Can you please fix them?

b0x42:
Fixed — rebased onto current master and pushed fixes for everything Greptile flagged (rollback/leak/use-after-free issues, plus a cross-path double-navigation bug found along the way). Build/CI still show `action_required` though — looks like they need a maintainer to approve running workflows on this PR, could you approve them?

davidbudnick:
> Fixed — rebased onto current master and pushed fixes for everything Greptile flagged (rollback/leak/use-after-free issues, plus a cross-path double-navigation bug found along the way). Build/CI still show `action_required` though — looks like they need a maintainer to approve running workflows on this PR, could you approve them?

Running them now!

PR #388, opened 2026-07-13
feat(hid): recognise Lightspeed nano receivers (G-series, e.g. G305)

kiwimaker:
## Summary

Adds recognition for the **Logitech Lightspeed nano receiver** (`046d:c53f`) — the receiver bundled with G-series wireless mice such as the **G305**. Before this change, `receiver::detect()` only matched Bolt (`c548`) and Unifying (`c52b`, `c532`) receivers; a Lightspeed receiver fell through to `None`, so any mouse paired to it was completely invisible (the mouse never appeared in `list` / the GUI, even though the keyboard on a Unifying receiver did).

A Lightspeed nano receiver is **not** a Unifying receiver, but it exposes the same HID++ 1.0 receiver registers (`0x02` connections, `0xB5/0x5N` pairing info) that this codebase already drives for Unifying. So the cleanest fix is to route it through the existing Unifying code path and differentiate it only where it's user-visible: the receiver's display name.

## Changes

- **`openlogi-hidpp` / `receiver::unifying::VPID_PAIRS`** — add `046d:c53f` so `detect()` builds the (Unifying-protocol) receiver for it. Documented that this is a protocol-compat Lightspeed receiver, not Unifying proper.
- **`openlogi-hid` / `route.rs`** — new `LIGHTSPEED_PIDS` list plus two small helpers: `speaks_unifying_protocol(pid)` (Unifying ∪ Lightspeed) and `receiver_display_name(pid)`. `device_route_for` now routes Lightspeed receivers as `DeviceRoute::Unifying` (they were previously defaulting to `Bolt`, which would address pairing slots with the wrong register flow on the write path).
- **`openlogi-hid` / `pairing.rs`** — `family_for` classifies Lightspeed as `ReceiverFamily::Unifying` via the shared helper.
- **`openlogi-hid` / `inventory/probe.rs`** — the enumerated receiver is now labelled via `receiver_display_name`, so a Lightspeed receiver surfaces as **"Lightspeed Receiver"** while Unifying stays "Unifying Receiver".
- Tests: added coverage that Lightspeed PIDs route as Unifying and that the display name resolves correctly.

## Testing

Verified on macOS with a real **Logitech G305** (paired device wpid `0x4074`) on its Lightspeed nano receiver:

```
# before
$ openlogi list        # G305 receiver not recognised — mouse invisible

# after
Lightspeed Receiver (E93B9602, vid=046d pid=c53f)
  └─ slot 1 ● mouse, wpid=4074 ... transports=equad
Unifying Receiver (E3C15593, vid=046d pid=c52b)
  └─ slot 1 ● keyboard, wpid=408a ...
```

- `cargo test -p openlogi-hid -p openlogi-hidpp` — all pass (incl. the new tests).
- `cargo clippy` / `cargo fmt` clean on the touched crates.

## Notes / scope

- I could only test `0xc53f` (the receiver that shipped with my G305). Other Lightspeed receiver PIDs (`0xc539`, `0xc541`, `0xc545`, `0xc547`, …) almost certainly work through the same path, but I've deliberately **not** added PIDs I couldn't verify — they can be appended to `LIGHTSPEED_PIDS` as people confirm them.
- Naming lives in the app layer (`openlogi-hid`) rather than the vendored `hidpp` fork, keeping the protocol layer's notion of "this is the Unifying register protocol" intact.
- The GUI's per-device "connected via … receiver" label (derived from `DeviceRoute::Unifying`) still reads "Unifying"; the primary receiver name shown in the inventory/carousel is correct. Happy to extend the GUI label too if you'd prefer a follow-up.

### DIFF
--- a/crates/openlogi-hid/src/inventory/probe.rs
+++ b/crates/openlogi-hid/src/inventory/probe.rs
@@ -254,7 +254,7 @@ async fn probe_unifying_receiver(
     NodeProbe {
         inventory: Some(DeviceInventory {
             receiver: ReceiverInfo {
-                name: "Unifying Receiver".to_string(),
+                name: crate::route::receiver_display_name(info.product_id).to_string(),
                 vendor_id: info.vendor_id,
                 product_id: info.product_id,
                 unique_id,
--- a/crates/openlogi-hid/src/lib.rs
+++ b/crates/openlogi-hid/src/lib.rs
@@ -59,7 +59,10 @@ pub use pairing::{
     Click, DiscoveredDevice, PairingCommand, PairingError, PairingEvent, PairingReceiver,
     PasskeyMethod, ReceiverFamily, ReceiverSelector, list_pairing_receivers, run_pairing, unpair,
 };
-pub use route::{BOLT_PIDS, DIRECT_DEVICE_INDEX, DeviceRoute, UNIFYING_PIDS};
+pub use route::{
+    BOLT_PIDS, DIRECT_DEVICE_INDEX, DeviceRoute, LIGHTSPEED_PIDS, UNIFYING_PIDS,
+    receiver_display_name, speaks_unifying_protocol,
+};
 pub use smartshift::{AUTO_DISENGAGE_PERMANENT, SmartShiftMode, SmartShiftStatus};
 pub use standalone::enumerate_standalone;
 pub use write::{
--- a/crates/openlogi-hid/src/pairing.rs
+++ b/crates/openlogi-hid/src/pairing.rs
@@ -78,7 +78,8 @@ impl From<ReceiverFamily> for PairingPhase {
 fn family_for(product_id: u16) -> Option<ReceiverFamily> {
     if crate::BOLT_PIDS.contains(&product_id) {
         Some(ReceiverFamily::Bolt)
-    } else if crate::UNIFYING_PIDS.contains(&product_id) {
+    } else if crate::speaks_unifying_protocol(product_id) {
+        // Unifying proper plus protocol-compatible Lightspeed receivers.
         Some(ReceiverFamily::Unifying)
     } else {
         None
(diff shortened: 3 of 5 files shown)


Stanley5249:
Tested this approach on Windows 11 with a **G502 X LIGHTSPEED** (receiver `046d:c547`, wpid `409f`): applied the patch onto current master (it needs a small rebase — the `route.rs` tests context drifted), added `0xc547` to `LIGHTSPEED_PIDS` and `VPID_PAIRS`, and everything works end to end:

```
Lightspeed Receiver (9A998722, vid=046d pid=c547)
  └─ slot 1 ● G502 X LS (mouse, wpid=409f, battery=63% good (discharging))
```

`diag features` reports 29 features (incl. `0x2201` AdjustableDpi, `0x8100`/`0x8110` onboard profiles), and `diag dpi` completes a read → write → read-back → restore round-trip through the `DeviceRoute::Unifying` path.

Two suggestions while you're in here:
1. Could you add `0xc547` to both lists? Happy to share my diff if useful.
2. `transport.rs`'s Linux receiver-child filter (`is_receiver_child_sysfs_path`) still only chains `BOLT_PIDS`/`UNIFYING_PIDS`, so `hid-logitech-dj` child nodes of a Lightspeed receiver wouldn't be filtered on Linux — probably worth chaining `LIGHTSPEED_PIDS` there too.

xianjianbo:
Hi, thanks for adding recognition support for the G304/G305 LIGHTSPEED receiver. I tested OpenLogi with a G304 (PID C53F, WPID 0x4074, firmware RQM68.02_B0021). The device is detected correctly and DPI control works through 0x2201, but scroll-direction inversion is shown as unavailable. Looking through the code and the device’s feature table, this appears to be because the mouse does not advertise [0x2121 (HiResWheel)](https://openlogi.org/zh/hidpp/features/x2121-hires-wheel); it also does not advertise 0x2120 or 0x2100. Are you aware of any other HID++ 1.0 register or HID++ 2.0 feature that the G304/G305 might expose for reversing the wheel direction?

Issue #313, opened 2026-06-23
[Bug]: MX Master 4 haptic thumb pad does not trigger gestures (CID 0x01A0 unsupported)


billwork-ops:
### Pre-flight checklist

- [x] I searched existing issues and did not find this specific MX Master 4 input-capture bug.
- [x] I am using the latest release, v0.6.16.
- [x] Logi Options+ is not running while testing OpenLogi.

### Which part of OpenLogi?

GUI (desktop app) / HID++ control capture

### OpenLogi version

0.6.16 (GUI and agent)

### Operating system

macOS 15.7.7 (arm64)

### Device model

MX Master 4 for Business (WPID `b048`)

### Connection

Logi Bolt receiver; the tested device is online in slot 1. The receiver identifier has been omitted from this report.

### Affected area

Button remapping / gesture-button capture

### What happened?

The GUI allows **Thumb pad** to be selected as the gesture owner and accepts Click/Up/Down/Left/Right bindings. Pressing or holding and swiping on the MX Master 4 haptic thumb pad does not trigger any bound action.

The device is online through Bolt, Accessibility is granted, and the macOS input hook is installed. This is therefore different from the Bluetooth-direct capture limitation.

Expected behavior: selecting **Thumb pad** should arm the physical haptic thumb pad / Action Ring button and dispatch its click and directional gestures.

### Steps to reproduce

1. Connect an MX Master 4 for Business through a Logi Bolt receiver.
2. Select the online device in OpenLogi.
3. Open the button configuration and choose **Thumb pad** as the gesture owner.
4. Bind Click and one or more directions to visible actions such as App Expose or Mission Control.
5. Press/release the haptic thumb pad, then hold and swipe in a configured direction.
6. Observe that no bound action runs.

### Diagnostics report (privacy-filtered)

```markdown
OpenLogi GUI: v0.6.16 (release)
Agent: v0.6.16 (connected)
IPC protocol: 8
Inventory: ready
OS: macOS 15.7.7 (arm64)
Accessibility: granted
Input hook: installed
Config schema: 3

Device: MX Master 4 for Business
Connection: Logi Bolt receiver
Online: yes
Battery: 90%
WPID: b048
Model IDs: b048/0000/0000
Capabilities: buttons=yes, pointer=yes, lighting=no
Transport: BTLE
Render: mx_master_4_for_business
Slot: 1
```

### Suspected root cause

[...]

PR #415, opened 2026-07-18
fix: divert thumbwheel for rotation rebinds even without single-tap capability

tagawa0525:
## Problem

On an MX Master 4, rebinding the thumbwheel rotation (`ThumbwheelScrollUp` / `ThumbwheelScrollDown`) or changing the thumbwheel sensitivity has no effect — silently. The GUI accepts and persists the config, `reload_config` goes through, the capture session restarts, but the wheel keeps scrolling natively in both directions.

## Root cause

`arm_controls` in `openlogi-hid/src/gesture.rs` only diverts the thumb wheel over `0x2150` when `getThumbwheelInfo` reports the single-tap capability:

```rust
if supports_single_tap {
    tw.set_reporting(true, false).await?;
    thumb = Some((tw, info.index));
} else {
    debug!("thumb wheel reports no single tap — click not capturable");
}
```

The MX Master 4 does not set that capability bit (verified on hardware — Bolt receiver, feature `0x2150` v0 present at index 19; getInfo byte 5 has `0x08` clear). Its thumb area was reworked into a haptic pad, which already behaves differently over HID++ elsewhere (see #313), so the missing bit is plausibly genuine rather than a firmware reporting bug.

But single-tap support only matters for capturing the wheel's *click*. Rotation rebinds and the sensitivity multiplier need the diverted event stream too — `thumbwheel_armed()` in `openlogi-agent-core` correctly requests capture for them — and the arming side then silently refuses. The user gets no error anywhere: just a debug-level line, and settings that appear to do nothing. This affects every platform equally; it just hasn't been visible because MX Master 3/3S report the bit and arm fine.

## Fix

Divert whenever capture was requested, regardless of the single-tap capability. A missing single tap now only means a bound click can never fire (the `single_tap` event bit simply never arrives), which is the accurate scope of the limitation. Doc comments updated to match.

Behavior is unchanged for wheels that do report single tap.

If the original gate was intentional (avoid suppressing native scroll on wheels where re-synthesis was considered untested), I'd argue the inconsistency should then be fixed on the `thumbwheel_armed()` side instead — requesting capture and silently not arming is the worst of both. Happy to rework in that direction if preferred.

## Verification

On real hardware (MX Master 4, slot 2 on a Bolt receiver, Linux/Wayland):

Before — with `ThumbwheelScrollUp = "HorizontalScrollLeft"` saved and reloaded:

```
DEBUG openlogi_hid::gesture: thumb wheel reports no single tap — click not capturable
INFO  openlogi_hid::gesture: control capture active index=2 gesture=false dpi_buttons=1 thumbwheel=false
```

After:

```
INFO  openlogi_hid::gesture: control capture active index=2 gesture=false dpi_buttons=1 thumbwheel=true
```

and the wheel behaves as configured (both rotation rebinds and the sensitivity multiplier now apply).

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` all pass.

https://claude.ai/code/session_01N76prZoUHMrHZajXq4jET1

### DIFF
--- a/crates/openlogi-hid/src/gesture.rs
+++ b/crates/openlogi-hid/src/gesture.rs
@@ -13,7 +13,8 @@
 //! it, mirroring how the CGEventTap hook handles the side buttons. The thumb
 //! wheel is special: diverting it stops native horizontal scroll, so the GUI
 //! re-synthesises scroll from the [`CapturedInput::Scroll`] deltas — the wheel
-//! is therefore only diverted when its click is actually bound.
+//! is therefore only diverted when the user's thumbwheel config leaves its
+//! defaults (click bound, rotation rebound, or sensitivity changed).
 
 use std::sync::{Arc, Mutex, PoisonError, RwLock};
 
@@ -44,8 +45,8 @@ pub enum CapturedInput {
     /// ([`ButtonId::Thumbwheel`]).
     ButtonPressed(ButtonId),
     /// Thumb-wheel rotation to re-synthesise as horizontal scroll, in the
-    /// wheel's `diverted_res` increments. Emitted only while the wheel is
-    /// diverted to capture its click.
+    /// wheel's `diverted_res` increments. Emitted while the wheel is diverted
+    /// (click bound, rotation rebound, or sensitivity changed).
     Scroll(i16),
 }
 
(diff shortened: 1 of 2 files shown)


tagawa0525:
Both stale doc comments flagged by the review are updated in 6304f4b — the module-level doc and the `CapturedInput::Scroll` variant doc now describe the broadened divert trigger (click bound, rotation rebound, or sensitivity changed).

PR #419, opened 2026-07-18
feat: per-device capture — every online device gets its own session, bindings, and settings

tagawa0525:
## Problem

The runtime manages only the GUI-selected device. With more than one mouse connected this produces two real defects:

1. **The non-selected device's settings are dead.** Its thumbwheel rebinds, sensitivity, and button captures silently stop applying the moment you switch the carousel to another device.
2. **The Linux evdev hook applies the selected device's profile to every mouse it grabs** — including other vendors' mice and virtual pointers (e.g. keyd's). Side-button remaps configured for an MX Master land on whatever other pointing device you plug in.

## Design

Capture moves to the HID++ layer per device, where device identity is structural (one channel per device) instead of inferred:

- **`capture_plan` (agent-core, new):** the orchestrator rebuilds one `DeviceCapturePlan` per *online* device — route, per-device binding maps, divert set, effective thumbwheel sensitivity — on every config / app / inventory change, including pure online-flag flips.
- **Plan-driven sessions (`openlogi-hid`):** `run_capture_session` takes a `CaptureSpec`. Rebindable standard buttons (middle/back/forward, `0x0052/0x0053/0x0056`) divert over `0x1b04` and dispatch as `ButtonPressed`. A button at its default binding is **never diverted**, so native behavior needs no re-synthesis; the OS-hook gesture owner's button is likewise excluded so hold+swipe keeps its events.
- **Multi-session watcher:** sessions live in a map keyed by config key; inputs arrive tagged with their device and dispatch against that device's own plan (per-device thumbwheel accumulators included). The capture-vs-pairing lease is shared across sessions via an `Arc` and frees itself when the last session exits — pairing semantics unchanged.
- **Per-device settings:** `DeviceConfig.thumbwheel_sensitivity` overrides the app-wide default (which remains the fallback and keeps its Settings → General slider); the SmartShift panel gains a per-device sensitivity slider (existing localized label, no new i18n keys). `DpiCycles` replaces the single cycle state: HID++ dispatch cycles the device an event arrived on, the OS hook (which cannot attribute events) keeps targeting the selection, and a config reload no longer snaps an unchanged device's cycle index back to `preset[0]`.
- **Per-device manage toggle:** `DeviceConfig.enabled` (default `true`). A disabled device is left fully native — no capture session, no volatile re-apply, no DPI-cycle entry, empty hook maps. The Home gallery ring now shows managed state (accent blue = managed, red = disabled) instead of marking the selection, since capture no longer depends on it.
- **Linux hook:** grabs only Logitech (VID `046d`) mice, mirroring the macOS hook's vendor awareness.

## Verification

On real hardware (MX Master 4 + MX Master 3S on two Bolt receivers, Linux/Wayland):

- Both devices hold concurrent capture sessions; side buttons, gesture swipes/click, DPI button, thumbwheel rotation and (3S) single tap all arrive correctly attributed with no OS hook involved, and every diverted control restores on teardown.
- Independent per-device behavior confirmed live: different thumbwheel directions and sensitivities on each mouse simultaneously, regardless of carousel selection.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` all pass.

HID++-layer changes are platform-neutral; runtime testing was on Linux (the OS-hook change is Linux-only).

## Notes

- The first two commits are #415 (thumbwheel divert fix) — this branch builds on it; if #415 merges first this PR rebases to just the five feature commits.
- Known limitation kept intentionally small: the OS hook still applies the selection's profile to the (now Logitech-only) devices it grabs; fully per-device hook dispatch would need evdev device attribution plumbed through the hook and is left for a follow-up.

https://claude.ai/code/session_01N76prZoUHMrHZajXq4jET1

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -1,0 +1,187 @@
+//! Per-device capture plans: what each online device's HID++ capture session
+//! should divert, plus the device's own binding maps for dispatch.
+//!
+//! The orchestrator rebuilds the shared plan list from config + inventory for
+//! *every* online device (not just the GUI's selection), and the capture
+//! watcher diffs it into running sessions. Keeping the binding maps inside the
+//! plan is what makes dispatch per-device: an input is resolved against the
+//! plan of the session it arrived on, never against a global selected-device
+//! map.
+
+use std::collections::BTreeMap;
+use std::sync::{Arc, RwLock};
+
+use openlogi_core::binding::{Action, ButtonId, GestureDirection, default_binding};
+use openlogi_core::config::Config;
+use openlogi_hid::DeviceRoute;
+use openlogi_hid::gesture::DIVERTABLE_STANDARD_BUTTONS;
+use openlogi_hid::reprog_controls::GESTURE_BUTTON_CID;
+
+use crate::bindings::{bindings_for, gesture_bindings_for, oshook_gestures_for};
+
+/// Everything the capture watcher needs to run one device's session and
+/// dispatch its events.
+#[derive(Debug, Clone, PartialEq)]
+pub struct DeviceCapturePlan {
+    /// Stable per-device config key (binding / preset lookup).
+    pub config_key: String,
+    /// HID++ route the session opens.
+    pub route: DeviceRoute,
+    /// Per-button single actions for this device (per-app effective).
+    pub bindings: BTreeMap<ButtonId, Action>,
+    /// Per-direction map when the dedicated HID++ gesture button owns the
+    /// gesture role on this device; empty otherwise.
+    pub gesture_bindings: BTreeMap<GestureDirection, Action>,
+    /// Standard buttons whose binding leaves the default — divert over
+    /// `0x1b04`. A button at its default keeps its native HID behavior, so no
+    /// re-synthesis is ever needed.
+    pub divert_buttons: Vec<(u16, ButtonId)>,
+    /// Whether any thumbwheel binding leaves its default. Combined with the
+    /// sensitivity to decide thumb-wheel diversion.
+    pub thumbwheel_bindings_nondefault: bool,
+    /// This device's effective thumb-wheel sensitivity (device override or the
+    /// app-wide default).
+    pub thumbwheel_sensitivity: i32,
+    /// Capture re-arm generation from the orchestrator. Bumps on reconnect /
+    /// system wake so sessions restart even when route and divert set match.
+    pub rearm_generation: u64,
+}
+
+/// Shared plan list, rewritten by the orchestrator and read by the watcher.
+pub type SharedCapturePlans = Arc<RwLock<Vec<DeviceCapturePlan>>>;
+
+/// Build one device's plan from the config (per-app effective for `app`).
+#[must_use]
+pub fn plan_for_device(
+    config: &Config,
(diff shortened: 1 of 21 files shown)


tagawa0525:
Both review findings addressed in 170cfbe:

- **DPI cycles on the inventory fast path**: the same-set/fresh-online-flags branch now goes through a single `publish_device_runtime()` that rewrites the capture plans *and* the per-device DPI-cycle map together, so a waking device gets both its session and its DPI slot — and the two views can't diverge again by construction.
- **ToggleSmartShift write lock**: `DpiCycles` gained a read-only `target_for()` (with a unit test covering the explicit-key and selection-fallback paths) and the toggle now takes only a read lock.

tagawa0525:
The remaining cross-tick window flagged in the re-review summary is closed in e277c5b: a deliberately stopped session now stays tracked (stop sender taken) until its task's completion report arrives — which is sent only after `run_capture_session` returns, restore writes included — and a tracked key is never re-armed. This covers both plan changes and the post-pairing recovery path, no matter how many ticks the teardown takes.

tagawa0525:
Fixed in b00d213: `set_current_app` now routes through `publish_device_runtime()`, so the capture plans (binding maps and divert sets, both per-app effective) are republished whenever the foreground app changes. Covered by `app_switch_republishes_capture_plans` (RED in 2985353). DPI cycle state is preserved across the republish since presets are not app-scoped.

tagawa0525:
On the two P2 notes from the latest review pass:

- **Sensitivity override not clearable to `None`**: fixed in 8065e4c. Committing the per-device slider at the app-wide default now clears the override (the slider is the device's only sensitivity control, so landing on the default is the "no override" gesture), and the no-op guard compares the stored override rather than the effective value so an existing pinned-at-default override actually clears.
- **Forwarding-task drain ordering**: keeping as is. Dispatch deliberately resolves every input against the plans in force at dispatch time (same event-time semantics as the OS hook on a config edit), so a buffered input from a stopping session is interpreted exactly as if it had arrived a moment later — reordering `done` behind the forwarding drain wouldn't change which plan those one or two events hit, and no hardware write is involved.

tagawa0525:
Fixed in b57b279: the teardown clear is now guarded by an `Arc::ptr_eq` identity check, so a stopping session only removes its own published channel and never evicts a sibling session's.

rafamadriz:
Out of curiosity, will this fix the issue of openlogi not working correctly when keyd has already remapped some keys?

I was giving it a tried the other day and I couldn't get the next/prev desktop action to work. Instead, Caps Lock was being toggle whenever I did the gesture. It all came down to keyd and openlogi not playing nice together, I use keyd it to remap ctrl to caps lock plus some other things.

I didn't find an existing issue but it seems like this addresses that.

davidbudnick:
> Out of curiosity, will this fix the issue of openlogi not working correctly when keyd has already remapped some keys?
> 
> I was giving it a tried the other day and I couldn't get the next/prev desktop action to work. Instead, Caps Lock was being toggle whenever I did the gesture. It all came down to keyd and openlogi not playing nice together, I use keyd it to remap ctrl to caps lock plus some other things.
> 
> I didn't find an existing issue but it seems like this addresses that.

👋🏻 Hey, if this is still causing issues for you feel free to open an issue and report it :)

PR #430, opened 2026-07-20
fix(agent): apply per-app gesture button clicks

devsaddle:
## Summary

Apply per-app single-action overrides to the Click arm of dedicated HID++ gesture sources while keeping their global swipe directions and gesture mode intact.

This revision is rebased onto current `master` at OpenLogi 0.8.3 and adapts the fix to the current multi-source, per-button gesture architecture, including long-press capture behavior. It supports the dedicated gesture button and other HID++ gesture sources such as the MX Master 4 haptic panel.

## Root cause

Capture plans already rebuild ordinary bindings for the foreground app, but `hidpp_gesture_maps_for` projected only the global gesture maps. A dedicated HID++ gesture press therefore kept dispatching its global Click action after the foreground app changed.

## Changes

- pass the foreground app into `hidpp_gesture_maps_for`
- build gesture mode and swipe arms from the global binding shape
- overlay an effective per-app `Binding::Single` onto only `GestureDirection::Click`
- preserve current long-press exclusion and side-gesture diversion behavior
- pass the app through each device capture plan while keeping the desktop editing view global
- cover click override, swipe preservation, app exit fallback, and foreground-app capture-plan republishing

## Testing

- `RUSTFLAGS=-D warnings cargo fmt --all -- --check`
- `RUSTFLAGS=-D warnings cargo clippy --workspace --all-targets --locked -- -D warnings`
- `RUSTFLAGS=-D warnings cargo test --workspace --locked`
- `RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent`
- `cargo xtask ci wasm`

Previously hardware-validated with an MX Master 2S using Safari and Chrome per-app navigation actions. The rebased implementation has full workspace coverage; hardware behavior will be revalidated in the combined local 0.8.3 patch build.

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -115,7 +115,7 @@ pub fn plan_for_device(
     // One direction map per HID++ source in gesture mode — several may
     // gesture at once, each armed with its own raw-XY divert (the capture
     // target below derives the CIDs to divert from this map's keys).
-    let gesture_bindings = hidpp_gesture_maps_for(config, Some(config_key));
+    let gesture_bindings = hidpp_gesture_maps_for(config, Some(config_key), app);
     let divert_gesture_buttons = if os_mouse_hook_available {
         DIVERTABLE_STANDARD_BUTTONS
             .into_iter()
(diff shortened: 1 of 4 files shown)


davidbudnick:
Fixes #878

PR #464, opened 2026-07-26
fix: prevent unintended thumbwheel tap actions

### LINKED ISSUE #435: [Bug]: Capacitive touch on the horizontal scroll whell
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.21

### Operating system

Windows

### OS version & architecture

Windows 11 Pro 25H2

### Device model

MX Master 3s

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

If i invert the directions of the horizontal scroll whell the directions are inverted, but the scroll whell starts to have a capacitive touch behavior, every time i touch it it opens App Exposé. If i just undo the change and keep the directions as default the behavior stops. 

### Steps to reproduce

1. invert the horizontal scroll direction

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

### LINKED ISSUE #477: [Bug]: Horizontal wheel volume control triggers Windows Task View unexpectedly
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.22

### Operating system

Windows

### OS version & architecture

Windows11 26H1 28120.2546, x64

### Device model

MX Master 3S 

Logi Bolt receiver

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

1. Open the software and navigate to button/wheel remapping settings.
2. Assign the horizontal wheel to “Volume Up” (for one direction) and “Volume Down” (for the opposite direction).
3. roll the horizontal wheel in either direction.

Expected behavior: The system volume should increase or decrease accordingly, with no other actions triggered.

Actual behavior: The volume does change, but Windows Task View (the virtual desktop overview) also opens immediately after each roll. It appears as if the horizontal wheel is sending the Win + Tab shortcut or triggering the Task View button.

### Steps to reproduce

1. Open the software and navigate to button/wheel remapping settings.
2. Assign the horizontal wheel to “Volume Up” (for one direction) and “Volume Down” (for the opposite direction).
3. roll the horizontal wheel in either direction.

Expected behavior: The system volume should increase or decrease accordingly, with no other actions triggered.

Actual behavior: The volume does change, but Windows Task View (the virtual desktop overview) also opens immediately after each roll. It appears as if the horizontal wheel is sending the Win + Tab shortcut or triggering the Task View button.

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.6.22 (release)
- Agent: v0.6.22 (connected)
- IPC protocol: GUI 10 / agent 10
- Inventory: ready
- OS: Windows (x86_64)
- Locale: zh-CN (UI: follow system)
- Launch at login: yes · Menu bar: yes · Update check: on
- Config: schema 3 · 2 configured device(s) · thumbwheel 100

**Assets**
- Cache path: C:\Users\21061\.local/share/openlogi\assets · Bundle assets: absent

**Devices (3)**
- MX Master 3S — mouse (codename: MX Master 3S)
  - Connection: Logi Bolt receiver · Online: no · Battery: 60% (discharging, full)
  - Model: receiver:39c9c141742607e9:slot:1 (wpid: b034, model-ids: b034/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_master_3s · Slot 1
- MX Master 3S — mouse (codename: MX Master 3S)
  - Connection: Logi Bolt receiver · Online: no · Battery: 60% (discharging, full)
  - Model: receiver:39c9c141742607e9:slot:2 (wpid: b034, model-ids: b034/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_master_3s · Slot 2
- MX Master 3S — mouse (codename: MX Master 3S)
  - Connection: Logi Bolt receiver · Online: yes · Battery: 60% (discharging, full)
  - DPI: 1000 dpi (range 200–8000, 157 steps)
  - Model: receiver:39c9c141742607e9:slot:3 (wpid: b034, model-ids: b034/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_master_3s · Slot 3

**Receivers (1)**
- Logi Bolt Receiver (VID 046d / PID c548)
```

### `openlogi list` output

```text

```

### Logs

```shell

```

mvanhorn:
## Summary

Replace the watcher's single thumb-wheel armed flag with a typed capture mode that distinguishes native reporting, diverted rotation only, and diverted rotation with tap delivery; derive it from sensitivity plus the effective click and rotation bindings. Thread that mode through `run_capture_session`, export it from `openlogi-hid`, and include it in the manager's desired/current session state so changing the thumb-wheel click binding restarts capture with the correct behavior. Keep rotation decoding and re-synthesis unchanged, but have the HID listener emit `ButtonPressed(Thumbwheel)` only when the mode explicitly includes taps, preventing rotation inversion from activating the seeded App Exposé default.

## Why

On an MX Master 3S, swapping the horizontal thumb-wheel directions causes touching the wheel to launch App Exposé even though only rotation was edited. The current watcher diverts HID++ Thumbwheel feature `0x2150` whenever either rotation binding differs from its default, and the diverted listener then forwards every reported single tap as `ButtonId::Thumbwheel`. The effective binding map always seeds that button with App Exposé, so rotation-only customization unintentionally activates a dormant default tap action. This path remains present on current `master`, and there are no open or previously closed cross-referenced PRs for the issue.

## Testing

- Happy path: swap `ThumbwheelScrollUp` and `ThumbwheelScrollDown` while the thumb-wheel click remains at its seeded default; the watcher selects rotation-only capture, wheel rotation scrolls in the inverted direction, and a tap/touch report emits no App Exposé action.
- Edge cases: a non-default thumb-wheel sensitivity still selects rotation-only capture, while explicitly rebinding `Thumbwheel` selects rotation-plus-tap and continues to deliver exactly one button event alongside any nonzero rotation in the same HID++ report.
- Error paths: at default sensitivity, an unavailable or poisoned effective binding map falls back to no diversion instead of enabling tap capture, and unrelated, response, or zero-motion HID++ messages without a tap continue to produce no captured input.

Fixes #435

AI was used for assistance.

Fixes #477

### DIFF
--- a/crates/openlogi-agent-core/src/watchers/gesture.rs
+++ b/crates/openlogi-agent-core/src/watchers/gesture.rs
@@ -3,7 +3,7 @@
 //! Runs [`openlogi_hid::run_capture_session`] concurrently for every device in
 //! the shared capture-plan list (not just the GUI's selection), restarts a
 //! session when its device's plan — route, diverted controls, thumb-wheel
-//! arming — changes, and dispatches each captured input against the binding
+//! capture mode — changes, and dispatches each captured input against the binding
 //! maps of the device it arrived on:
 //!
 //! - a gesture swipe through the gesture binding map,
@@ -27,17 +27,20 @@ use std::time::{Duration, Instant};
 use openlogi_core::binding::{Action, ButtonId, default_binding};
 use openlogi_core::config::ThumbwheelSensitivity;
 use openlogi_hid::session::gesture::{CaptureSpec, GESTURE_SOURCE_BUTTONS};
-use openlogi_hid::{CaptureChannel, CapturedInput, DeviceRoute, run_capture_session};
+use openlogi_hid::{
+    CaptureChannel, CapturedInput, DeviceRoute, ThumbwheelCaptureMode, run_capture_session,
+};
 use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, warn};
 
 use crate::capture_plan::{DeviceCapturePlan, SharedCapturePlans};
 use crate::hook_runtime::ActionDispatcher;
+#[cfg(test)]
+use crate::hook_runtime::SharedHookMaps;
 use crate::receiver_access::{ReceiverAccess, SessionReceiverLease};
 
-/// How often to re-read the active device target + thumb-wheel arming so a
-/// carousel switch or a binding/sensitivity edit re-points / re-arms capture.
-/// It also paces the respawn of a session that ended on its own (see `manage`).
+/// How often to re-read the device capture plans. It also paces the respawn of
+/// a session that ended on its own (see `manage`).
 const TARGET_POLL: Duration = Duration::from_secs(1);
 
 /// Idle gap after which a partly-accumulated *custom* wheel action is forgotten,
@@ -49,8 +52,7 @@ const ACTION_DECAY: Duration = Duration::from_millis(300);
 const ACTION_COOLDOWN: Duration = Duration::from_millis(200);
 
 /// Spawn the capture-manager thread. It owns a current-thread tokio runtime that
-/// keeps one capture session pointed at the active device and dispatches each
-/// captured input.
+/// keeps one capture session alive per online device and dispatches each input.
 pub fn spawn(
     capture_plans: SharedCapturePlans,
     capture_channel: CaptureChannel,
@@ -86,10 +88,64 @@ fn thumbwheel_armed(plan: &DeviceCapturePlan) -> bool {
         || plan.thumbwheel_bindings_nondefault
 }
 
+/// Select the thumb wheel's hardware reporting state from the live hook maps.
+///
+/// Test-only mirror of [`thumbwheel_capture_mode_for_plan`] that reads the
+/// published maps instead of a capture plan, so the orchestrator tests can
+/// assert the whole `Config -> hook maps -> capture policy` path rather than a
+/// hand-built plan. Tap provenance comes from `thumbwheel_tap_bound`, which is
+/// presence in the stored bindings, so an explicit binding equal to the seeded
+/// default still diverts.
+#[cfg(test)]
+pub(crate) fn thumbwheel_capture_mode(
+    hook_maps: &SharedHookMaps,
+    sensitivity: ThumbwheelSensitivity,
+) -> ThumbwheelCaptureMode {
+    let sensitivity_changed = sensitivity != ThumbwheelSensitivity::DEFAULT;
+    let Ok(maps) = hook_maps.read() else {
+        return if sensitivity_changed {
+            ThumbwheelCaptureMode::Diverted
+        } else {
+            ThumbwheelCaptureMode::Native
+        };
+    };
+    let binding_changed = |button| {
+        maps.bindings
+            .get(&button)
+            .is_some_and(|action| *action != default_binding(button))
+    };
+
+    if maps.thumbwheel_tap_bound
+        || sensitivity_changed
+        || [ButtonId::ThumbwheelScrollUp, ButtonId::ThumbwheelScrollDown]
+            .iter()
+            .any(|&button| binding_changed(button))
+    {
+        ThumbwheelCaptureMode::Diverted
+    } else {
+        ThumbwheelCaptureMode::Native
+    }
+}
+
+/// Whether this device's thumb wheel must leave its native HID reporting mode.
+///
+/// Only the two physical states exist. Tap intent is deliberately NOT encoded
+/// here: both "diverted for rotation" and "diverted for rotation and tap" arm
+/// the identical `set_reporting(true, false)`, so folding tap policy into the
+/// mode would restart an otherwise-identical session on every app/profile
+/// switch. The agent decides tap delivery from the live binding map instead.
+fn thumbwheel_capture_mode_for_plan(plan: &DeviceCapturePlan) -> ThumbwheelCaptureMode {
+    if thumbwheel_armed(plan) {
+        ThumbwheelCaptureMode::Diverted
+    } else {
+        ThumbwheelCaptureMode::Native
+    }
+}
+
 /// The [`CaptureSpec`] one device's session should run with right now.
 fn spec_for(plan: &DeviceCapturePlan) -> CaptureSpec {
     CaptureSpec {
-        capture_thumbwheel: thumbwheel_armed(plan),
+        thumbwheel_mode: thumbwheel_capture_mode_for_plan(plan),
         // Derived from the dispatch maps, so the armed diverts and the maps
         // resolving their events can never drift apart.
         divert_gesture_sources: GESTURE_SOURCE_BUTTONS
@@ -124,6 +180,35 @@ enum DoneAction {
     Remove { unexpected: bool },
 }
 
+/// Resolve a captured button press against the latest published hook maps.
+///
+/// A diverted thumb wheel can keep reporting taps while an app/profile switch
+/// removes its click binding. The provenance bit gates those stale in-flight
+/// reports; other captured buttons use the effective action map directly.
+#[cfg(test)]
+pub(crate) fn captured_button_action(
+    hook_maps: &SharedHookMaps,
+    button: ButtonId,
+) -> Option<Action> {
+    hook_maps.read().ok().and_then(|maps| {
+        if button == ButtonId::Thumbwheel && !maps.thumbwheel_tap_bound {
+            return None;
+        }
+        maps.bindings.get(&button).cloned()
+    })
+}
+
+/// Resolve a captured button press against one device's latest plan.
+fn captured_button_action_for_plan(plan: &DeviceCapturePlan, button: ButtonId) -> Option<Action> {
+    // Tap delivery follows the binding's provenance, not the session mode:
+    // an explicit tap stays deliverable while rotation or sensitivity keeps the
+    // wheel diverted for other reasons.
+    if button == ButtonId::Thumbwheel && !plan.thumbwheel_tap_bound {
+        return None;
+    }
+    plan.bindings.get(&button).cloned()
+}
+
 /// Decide the [`DoneAction`] for a completion report carrying `done_epoch`,
 /// given the session the manager currently tracks for that device (if any).
 ///
@@ -132,7 +217,6 @@ enum DoneAction {
 /// gone was stopped deliberately and is merely draining — its report frees the
 /// key quietly. One still holding its stop sender exited on its own and
 /// warrants a warning alongside the re-arm.
-/// Whether a finished session should be re-armed after completion.
 pub(crate) fn should_rearm(done_epoch: u64, live_epoch: u64, has_target: bool) -> bool {
     done_epoch == live_epoch && has_target
 }
@@ -404,9 +488,9 @@ fn dispatch(
             }
         }
         CapturedInput::ButtonPressed(button, _) => {
-            if let Some(action) = plan.bindings.get(&button) {
+            if let Some(action) = captured_button_action_for_plan(plan, button) {
                 debug!(key, ?button, action = %action.label(), "HID++ button → action");
-                dispatcher.dispatch(action, Some(key));
+                dispatcher.dispatch(&action, Some(key));
             } else {
                 debug!(key, ?button, "HID++ button with no binding — ignored");
             }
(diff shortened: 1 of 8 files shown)


crates/openlogi-agent-core/src/watchers/gesture.rs
```diff
@@ -104,29 +106,37 @@ pub fn spawn(
...
+        return if sensitivity_changed {
+            ThumbwheelCaptureMode::DivertedRotation
+        } else {
+            ThumbwheelCaptureMode::Native
+        };
+    };
+    let binding_changed = |button| {
+        maps.bindings
+            .get(&button)
+            .is_some_and(|action| *action != default_binding(button))
```
AprilNEA, line 115: [P1] Preserve explicit tap-binding provenance

`HookMaps.bindings` is already seeded with `default_binding`, so comparing values here cannot distinguish an unset default from an explicit `Binding::Single(AppExpose)`. With rotation or sensitivity keeping the wheel diverted, that explicit tap is classified as `DivertedRotation` and suppressed indefinitely. Please carry explicit/effective binding provenance into the runtime policy instead of inferring intent from action equality; a `Config -> bindings_for -> capture policy` regression test should cover this.

crates/openlogi-hid/src/gesture.rs
```diff
@@ -255,32 +272,18 @@ async fn arm_controls(
...
-            tw.set_reporting(true, false)
-                .await
-                .map_err(|e| GestureError::Hidpp(format!("{e:?}")))?;
-            thumb = Some((tw, info.index));
-        } else {
-            debug!("thumb wheel reports no single tap — click not capturable");
-        }
+        tw.set_reporting(true, false)
+            .await
+            .map_err(|e| GestureError::Hidpp(format!("{e:?}")))?;
```
AprilNEA: [P1] Roll back partially armed controls

Gesture/DPI controls are diverted earlier in this function, but a thumbwheel feature lookup or `set_reporting` error returns before `ArmedControls` reaches the caller and before listener registration. That leaves the already-diverted controls with no consumer. Please accumulate armed state and disarm it before propagating any later setup error; record a write as potentially applied before awaiting it because a timeout does not prove firmware rejected it.

crates/openlogi-hid/src/gesture.rs
```diff
@@ -142,12 +147,7 @@ pub async fn run_capture_session(
             if let Some(idx) = thumb_index
                 && let Some(event) = thumbwheel::decode_event(&msg, device_index, idx)
             {
-                if event.single_tap {
-                    let _ = sink.send(CapturedInput::ButtonPressed(ButtonId::Thumbwheel));
-                }
-                if event.rotation != 0 {
-                    let _ = sink.send(CapturedInput::Scroll(event.rotation));
-                }
+                forward_thumbwheel_event(event, thumbwheel_mode, &sink);
```
AprilNEA: [P2] Do not snapshot tap policy in the HID listener

The listener captures this mode for the session, while app/config changes replace `hook_maps` immediately and the manager recomputes the mode only on the 1-second poll. During `AndTap -> Rotation`, an old listener can forward a tap that `dispatch` resolves against the new seeded AppExpose action, reproducing #435; the reverse transition drops configured taps. Please gate taps in the agent using the same current policy used for dispatch.

crates/openlogi-agent-core/src/watchers/gesture.rs
```diff
@@ -200,15 +210,15 @@ async fn manage(
                     target.map(|t| {
                         (
                             t,
-                            thumbwheel_armed(&hook_maps, sensitivity),
+                            thumbwheel_capture_mode(&hook_maps, sensitivity),
```
AprilNEA: [P2] Keep software-only policy out of session identity

Both diverted variants send the identical `set_reporting(true, false)` command; only tap filtering differs. Including the full enum in `want/current` stops the unified gesture/DPI/thumbwheel session, disarms it, and waits at least until the next 1-second tick before replacement. Please track only Native vs Diverted as hardware state and update tap delivery without restarting the session.

AprilNEA (changes requested):
Requesting changes because this currently introduces two blocking correctness regressions and two lifecycle races. I reproduced the explicit-default case through the real `Config -> bindings_for -> capture policy` path (`DivertedRotation` instead of `DivertedRotationAndTap`). The exact PR head passes 83 `openlogi-hid` tests, 45 agent-core tests, and 8 wire tests, but those suites do not exercise config provenance, partial-arm rollback, or live profile transitions.

The smallest safe design is to keep the hardware session at Native/Diverted, carry explicit tap intent alongside the atomically published hook maps, filter taps in the agent against current state, and roll back partial setup. This overlaps #415; please preserve its capability diagnostic and contributor authorship when rebasing.

mvanhorn:
Reworked per your review. The capture policy no longer depends on tap intent resolved at session-arm time.

Your four points:

- Hardware session stays diverted. `arm_controls` registers each control into `ArmedControls` as it arms and rolls the whole set back when a later step fails, so a partial arm never leaves the device diverted.
- Explicit tap intent travels alongside the atomically published hook maps as `HookMaps::thumbwheel_tap_bound`, resolved through the real `Config -> effective_bindings` path rather than inferred at the HID layer.
- The agent filters taps against that flag, so a live profile or app switch takes effect against the published maps without restarting the diverted session.
- Partial-arm rollback as above.

On your point that the existing suites do not exercise the failure modes — that was right, and passing them proved nothing. Added 11 regressions: agent-core 45 -> 51, openlogi-hid 83 -> 88. They cover config provenance including the explicit-default case you reproduced, partial-arm rollback asserting nothing is left armed after a mid-arm failure, and live profile transitions in both directions (bound -> unbound must drop taps, and back). I checked they fail when the corresponding production change is reverted.

Merged upstream master, so #415 is in. Its capability diagnostic is intact — lack of single-tap is logged only and never gates `set_reporting` — and I kept @tagawa0525's commit and the comment explaining why the `getInfo` error is consumed before the next await. Your fork branch had also picked up a merge from David Budnick; that is merged in too rather than replaced.

Build and clippy clean. No hardware verification: I have no device to test against, so the tap behaviour is covered by the tests above, not observed.

tagawa0525:
**Hardware verification on retail units — the rework holds up (commit 948d476)**

Setup: Linux (NixOS, Wayland), Bolt receiver, MX Master 3S (wpid b034) + MX Master 4 (wpid b042), agent built from this branch with debug logging.

On the MX Master 3S (the device from #435 — it advertises and emits 0x2150 single-tap):

1. **Rotation-only rebind, tap unbound**: every rotation event was dispatched to its binding, and every tap report was filtered (`HID++ button with no binding — ignored`) with nothing fired (15 taps / 21 rotations observed). Notably, one tap arrived mid-rotation — the firmware reports taps from incidental finger contact while scrolling, so the filter matters in normal use, not only for deliberate taps.
2. **Tap explicitly rebound**: every tap was dispatched as the bound action; rotation unaffected.
3. **Explicit default stored in config** (`Thumbwheel = "AppExpose"`): every tap was dispatched — provenance retained (the case @AprilNEA reproduced).
4. **Live transitions via the `reload_config` RPC with the agent kept running**: after bound→unbound every subsequent tap was dropped; after unbound→bound every tap was delivered again — with zero capture re-arms in the log across both reloads, so the hardware session stayed diverted throughout.

MX Master 4 data point (#415 lineage): rotation divert works correctly. With a visible action explicitly bound to the tap so that any tap report would have been noticed, repeated touches produced no single-tap reports at all. Reading `getThumbwheelInfo` capabilities from both units:

```
MX Master 4:  time_stamp: true, touch: true, proxy: false, single_tap: false
MX Master 3S: time_stamp: true, touch: true, proxy: true,  single_tap: true
```

So the MX Master 4 still carries the touch sensor but has dropped the single-tap capability, and its behavior matches its declaration.

Two caveats: my units are on a Bolt receiver while #435 was Bluetooth-direct, so the transport differs (the 0x2150 divert path is the same). And this branch still parses schema v3 only — current master writes `schema_version = 4` (#566), which fails to parse and silently falls back to defaults, so a master merge is needed before this lands.

mvanhorn:
Pushed an update to this branch.

The four points you raised should now be in place: tap provenance is key existence rather than value comparison (`thumbwheel_tap_bound`), `arm_controls` disarms on the error path, the HID listener forwards taps and dispatch reads the current plan, and session identity is `capture_thumbwheel` so a tap-policy-only change does not restart the HID session. I also cleaned a couple of clippy failures in the HID tests.

vcanuel:
Tested locally on macOS (MX Master 3S, Logi Bolt receiver), thumb wheel bound to Volume Up/Down.

Before: scrolling volume intermittently triggered App Exposé — same root cause as #477 (a stray tap falling through to the seeded AppExpose default).

After (built from this PR): volume works and the stray taps no longer fire App Exposé. 

Hardware confirmation, not a code review. Thanks for the fix!

mvanhorn:
Thanks for the detailed review, and thanks @vcanuel for the hardware confirmation on the MX Master 3S.

All four points are addressed. Rebased onto current master; the diff is 7 source files.

P1 explicit tap-binding provenance: HookMaps now carries a thumbwheel_tap_bound flag derived in the orchestrator from config.effective_bindings(key, app).contains_key(&ButtonId::Thumbwheel). That is presence in the effective map, not action equality against the seeded default, so an explicit Binding::Single(AppExpose) is no longer indistinguishable from an unset default. Covered by the new Config -> bindings_for -> capture policy tests in orchestrator/tests.rs.

P1 roll back partially armed controls: arm_controls now delegates the fallible steps to arm_controls_into, which records each successful divert into ArmedControls as it lands. Any error from a feature lookup or set_reporting hits armed.disarm().await before the error propagates, so a mid-way failure hands every already-diverted control back to the firmware instead of leaving it captured with no consumer.

P2 do not snapshot tap policy in the listener: the listener no longer holds a tap decision. While diverted the transport forwards every decoded report, and the agent gates ButtonId::Thumbwheel against the current maps at dispatch time. An app or config change takes effect on the next report rather than on the next 1-second tick, so neither transition direction can misroute a tap.

P2 keep software-only policy out of session identity: ThumbwheelCaptureMode is now just Native or Diverted, matching the two distinct set_reporting commands. Tap delivery changes no longer alter want/current, so the unified gesture/DPI/thumbwheel session stays up instead of being disarmed and rebuilt.

cargo test on openlogi-hid and openlogi-agent-core: 291 passing, no warnings. I did not run the GUI crate locally.

mvanhorn:
Rebuilt on current master. Moving the gesture session into `openlogi-device` and precomputing `thumbwheel_bindings_nondefault` on the capture plan meant this needed re-expressing rather than rebasing, so it is one commit on master and mergeable again.

While porting I found I overstated something on 19 Aug. I said the agent gates `ButtonId::Thumbwheel` against the current maps at dispatch time. It did not. Dispatch went through `captured_button_action_for_plan`, which compared the plan's action against `default_binding`; the presence-based `thumbwheel_tap_bound` I added was only ever read by tests. Since `default_binding(ButtonId::Thumbwheel)` is `Action::AppExpose`, an explicitly bound `Single(AppExpose)` tap was still indistinguishable from an unset one on the live path, which is exactly the P1 you raised. That one slipped through, sorry.

Fixed: `DeviceCapturePlan` carries `thumbwheel_tap_bound`, derived in `plan_for` from `Config::effective_bindings(key, app).contains_key(&ButtonId::Thumbwheel)`, and dispatch reads that. Diversion still uses your value comparison, which is right there because it only has to be conservative.

New test `an_explicit_tap_binding_equal_to_the_default_still_counts_as_bound` pins the distinction: it asserts `thumbwheel_tap_bound` is true while `thumbwheel_bindings_nondefault` is false for the same config, so the two cannot quietly collapse together. Swapping presence back to a value comparison fails it.

287 tests pass across `openlogi-agent-core` and `openlogi-device`; fmt and clippy clean. Based directly on master now, so #415 and its capability diagnostic are inherited rather than re-applied and @tagawa0525's commit is untouched.

Still no hardware here, so @vcanuel's confirmation stands for the stray-tap symptom but not for this explicit-default case, which is covered by tests only.

AprilNEA:
#857 landed about an hour after your last push here — sorry for the sequencing.

To be clear about what happened: your rework did address the review. `thumbwheel_tap_bound` reached the dispatch path as a real `DeviceCapturePlan` field derived from `effective_bindings(...).contains_key(...)`, and the branch was rebased onto current master. Nothing was wrong with it.

What tipped it was a different starting point, not a better version of the same one. #857 seeds the tap with `Action::None`, and once the seed is inert, "did the user bind this?" and "is this action non-inert?" collapse into the same uestion — so the provenance flag stops being needed on the plan or on `HookMaps` at all.

The other half came out of @tagawa0525's note here that a tap arrives mid-rotation from incidental contact. That sent me looking for a protocol-level way to classify it, and it turned out to be `thumbwheelEvent` byte 4 (`rotation_status`) — a field the decoder had been discarding outright. #857 reads it to decide roll-vs-tap per report, with no clock and no per-device state. The same hardware trace also turned up something unrelated: diverted increments were being scrolled as if they were native ratchets, six times too far.

You're credited as co-author on 479e0952, and #435 / #477 are closed. Thanks for carrying this through two rounds —and thanks @tagawa0525 and @vcanuel for the hardware runs.

mvanhorn:
No worries on the sequencing. Seeding the tap with Action::None is the cleaner framing - once the seed is inert, "did the user bind this?" and "is this action non-inert?" really are the same question, and the provenance flag has nothing left to carry.

Nice catch on rotation_status. I had not considered that byte 4 already classified roll vs tap at the protocol level; that beats any heuristic layered on top of a clock. Good find on the diverted-increment scroll factor too.

Thanks for the co-author credit, and thanks to @tagawa0525 and @vcanuel for the hardware runs.

PR #473, opened 2026-07-27
fix(hook): report Windows pointer motion by differencing the cursor point

### LINKED ISSUE #471: Gesture swipes never commit on Windows
### What happens
Holding a gesture button (Middle/Back/Forward in gesture mode) and swiping does
nothing on Windows. Only the plain click on release fires. The same binding works
on macOS and Linux.

### Why
`crates/openlogi-hook/src/windows.rs` never emits `MouseEvent::Moved` — the
`WH_MOUSE_LL` translation handles buttons and both wheel axes, but not
`WM_MOUSEMOVE`. `SwipeAccumulator` therefore sees zero travel for the whole hold
and can never reach `GESTURE_SWIPE_THRESHOLD`.

The reason it was left out: `WH_MOUSE_LL` carries the cursor *position*
(`MSLLHOOKSTRUCT.pt`), not a delta, and the OS clamps that position to the desktop
bounds. macOS (`MOUSE_EVENT_DELTA_X`) and Linux (`REL_X`) both get the device's own
relative counts for free.

### Two candidate fixes
- #473 — difference `MSLLHOOKSTRUCT.pt` against the previous point in a thread-local.
  ~15 lines, no new APIs. A swipe that runs into a screen edge stops producing
  deltas and may not reach the threshold.
- #472 (draft) — register for raw input (`WM_INPUT`, `RIDEV_INPUTSINK`) and read the device's
  relative counts. No edge clamping, and matches the other two platforms. Costs a
  message-only sink window, a batched `GetRawInputBuffer` drain on the hook thread,
  and ordering care between the two input paths.

They are alternatives, not a sequence — one of them lands.

### Environment
Windows 11, OpenLogi 0.6.22, Logitech G502 X LIGHTSPEED.

Stanley5249:
## Summary
`WH_MOUSE_LL` never emitted `MouseEvent::Moved`, so `SwipeAccumulator` saw no
travel during a gesture hold and a swipe could never commit on Windows — only the
plain click on release fired. This differences the cursor position each
`WM_MOUSEMOVE` against the previous one, giving the accumulator the relative delta
it expects, the same shape as macOS `MOUSE_EVENT_DELTA_X` and Linux `REL_X`.

### `openlogi-hook`
- `windows.rs`: `motion_delta()` differences `MSLLHOOKSTRUCT.pt` against a
  `Cell<Option<POINT>>` thread-local; `WM_MOUSEMOVE` translates to
  `MouseEvent::Moved`. No synchronization — `WH_MOUSE_LL` callbacks run on the
  installing thread.
- Injected moves stay filtered from the event stream but do advance the baseline,
  so a cursor jump made by another process can't be reported as the user's travel.
- Tests: successive differencing (no-baseline first move, zero-delta repeat) and
  the injected-baseline case.

## Known limitation
The OS clamps `MSLLHOOKSTRUCT.pt` to the desktop bounds, so a swipe running into a
screen edge stops producing deltas and may not reach the threshold. Documented on
`motion_delta`. #472 is the alternative that avoids it, at the cost of ~350 lines
of raw-input FFI; these two are alternatives, not a sequence.

## Testing
- `cargo clippy -p openlogi-hook --all-targets` — clean
- `cargo test -p openlogi-hook` — 11 passed
- Not runtime-tested on hardware.

Note: the Windows CI job fails until #458 lands — `openlogi-hid` does not build
without the `Win32_System_IO` feature.

Fixes #471

### DIFF
--- a/crates/openlogi-hook/src/lib.rs
+++ b/crates/openlogi-hook/src/lib.rs
@@ -4,7 +4,7 @@
 //! |----------|---------------|
 //! | macOS    | `CGEventTap` (same primitive used by Logi Options+) |
 //! | Linux    | `evdev` grab + `uinput` re-injection |
-//! | Windows  | `WH_MOUSE_LL` low-level mouse hook |
+//! | Windows  | `WH_MOUSE_LL` low-level mouse hook (motion is edge-clamped) |
 //!
 //! # Usage
 //!
--- a/crates/openlogi-hook/src/windows.rs
+++ b/crates/openlogi-hook/src/windows.rs
@@ -29,8 +29,9 @@ use crate::{ButtonId, EventDisposition, HookError, MouseEvent};
 const WHEEL_DELTA: f32 = 120.0;
 
 thread_local! {
-    /// Cursor position carried by the previous `WM_MOUSEMOVE`, differenced into
-    /// the relative delta [`MouseEvent::Moved`] carries — see [`motion_delta`].
+    /// Cursor position carried by the previous mouse message of any kind,
+    /// differenced into the relative delta [`MouseEvent::Moved`] carries — see
+    /// [`translate_event`] and [`motion_delta`].
     ///
     /// Thread-local rather than a `static`: `WH_MOUSE_LL` callbacks run on the
     /// thread that installed the hook, so this is single-threaded by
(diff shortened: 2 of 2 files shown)


davidbudnick:
@Stanley5249 Thanks for your issue and PR, can you please address the `greptile-apps` comments and include a video of the changes working on your windows computer?

I have also approved your github action runs as well ✅

Stanley5249:
@davidbudnick Greptile's comment is resolved, and here's the demo.

https://github.com/user-attachments/assets/ba7adb89-bd0f-4561-9159-b197359f7ed8

davidbudnick:
> @davidbudnick Greptile's comment is resolved, and here's the demo.
> 
>  gesture-demo.mp4

Great thanks, for updating and providing the screen recording of the feature working correctly on windows. 

Merging it in!

PR #479, opened 2026-07-29
Feat/linked host switching

LCozatl:
## Summary

  Add keyboard-initiated Easy-Switch synchronization for linked Logitech pointing devices.

  When an Easy-Switch key is pressed on a compatible keyboard, OpenLogi switches every configured pointing device to the corresponding host channel before switching the keyboard. This allows the
  configured devices to move together between hosts.

  Implementing this feature also exposed a HID++ capture lifecycle issue: a device returning after an Easy-Switch transition retained its route but did not restart capture. The capture target now
  follows the device's online state without resetting its DPI-cycle position.

  ## Changes

  - `openlogi-core`
    - Add `host_switch_targets` to the per-device TOML configuration.
    - Support serialization and deserialization of physical target device keys.

  - `openlogi-hid`
    - Detect host-switch controls exposed through HID++ reprogrammable controls.
    - Support diverted-control and analytics-event reporting modes.
    - Switch configured pointing devices before switching the initiating keyboard.
    - Reuse the receiver channel when linked devices share a Bolt or Unifying receiver.
    - Restore temporarily modified control reporting when the session ends.

  - `openlogi-agent-core`
    - Resolve configured host-switch links against the current device inventory.
    - Rebuild links when the configuration or device availability changes.
    - Require the initiating keyboard to be online while preserving routes for sleeping target devices.
    - Restart HID++ capture when a device returns without resetting its DPI-cycle position.
    - Ignore missing targets and empty link configurations.
    - Add coverage for host-switch links and capture lifecycle transitions.

  - Documentation
    - Document `host_switch_targets` and provide a TOML configuration example.

  ## Testing

  Automated verification:

  - `devenv tasks run openlogi:check`

  OpenLogi hardware testing was performed on Linux using the OpenLogi 0.6.22 RPM package with:

  - MX Master 3S, firmware 22.1.6, linked to an MX Mechanical, firmware 71.3.21.
  - MX Master 3S, firmware 22.1.6, linked to an MX Mechanical Mini, firmware 72.3.14.

  Both combinations were tested with the new host-switch synchronization functionality.

  Cross-agent behavior was additionally checked using the official Logi Options+ 2.5 client:

  - Windows with a Bolt receiver.
  - macOS over Bluetooth.

  These checks confirmed that the devices can switch between corresponding host channels regardless of which agent manages them. OpenLogi itself was not runtime-tested on Windows or macOS as part
  of this change.

  ## Known limitations

  - Older keyboard firmware may not expose the HID++ controls or events required by this feature.
  - Every linked device must already be paired with the corresponding channel on each host.
  - The host-switch link must be configured on every computer from which the keyboard can initiate a switch.
  
  ## Note
  
  Depends on #453 for the online-aware input capture selection behavior. This branch will be rebased onto master after that PR merges, then re-tested before being marked ready for review.

### DIFF
--- a/crates/openlogi-agent-core/src/hardware.rs
+++ b/crates/openlogi-agent-core/src/hardware.rs
@@ -242,6 +242,10 @@ pub struct SmartShiftApply {
 /// stream while correlating responses only by software id — they cross-talk and
 /// produce the intermittent SmartShift `InvalidArgument` seen in #485. One
 /// sequential writer removes that self-race.
+#[expect(
+    clippy::too_many_arguments,
+    reason = "background reapply keeps one device write lifecycle together"
+)]
 pub fn reapply_mouse_volatile_in_background(
     capture: Option<&CaptureChannel>,
     registry: &ChannelRegistry,
--- a/crates/openlogi-agent-core/src/hardware.rs
+++ b/crates/openlogi-agent-core/src/hardware.rs
@@ -20,6 +20,8 @@ use openlogi_hid::{
 };
 use tracing::{debug, warn};
 
+use crate::receiver_access::ReceiverAccess;
+
 /// Upper bound on a single HID++ write. `hidpp` has no request timeout of its
 /// own, so without this an asleep / unresponsive device would hang (and leak)
 /// this background thread forever; a write to a live device completes in well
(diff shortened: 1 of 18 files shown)


LCozatl:
@greptileai

Issue #148, opened 2026-06-05
[Bug] Linux: udev rule misses Bluetooth-direct + diag picks wrong device (Phase 0 HID verification)


recchia:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

CLI (the `openlogi` binary)

### OpenLogi version

0.5.2

### Operating system

Linux

### OS version & architecture

Ubuntu 26.04 LTS, x86_64 (kernel 7.0.0-22), GNOME on Wayland

### Device model

MX Master 3S + Logi Bolt receiver (also tested Bluetooth-direct); MX Keys also paired

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [x] Device discovery / detection
- [ ] Button remapping
- [x] DPI control
- [x] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

**Not a broken device — this is a Phase 0 verification of the Linux HID/device layer, with two actionable findings (both have PRs ready).** Follow-up to #95 (that issue is the event hook; this is the device layer).

The `openlogi-hid` device layer **works on Linux** across both transports — Bolt receiver and Bluetooth-direct — for enumeration, battery, DPI and SmartShift, including the write path:

| Path | Enumerate | Battery | `diag features` | DPI write/restore | SmartShift toggle/restore |
|------|-----------|---------|-----------------|-------------------|---------------------------|
| Bolt receiver | ✅ slot 2 | ✅ 90% | ✅ | ✅ 1000→1050→1000 | ✅ Ratchet→Free→Ratchet |
| Bluetooth-direct (`probe_direct`) | ✅ slot 255 / idx 0xff | ✅ 90% | ✅ (36 features incl. `0x2201`, `0x2110`) | ✅ 1000→1050→1000 | ✅ Ratchet→Free→Ratchet |

**Finding 1 — udev rule needs two matchers; the Bluetooth-direct path is undocumented.**
[...]

PR #530, opened 2026-08-08
fix(linux): grant uaccess on Logitech input event nodes

### LINKED ISSUE #529: [Bug]: Linux: udev rules miss the input event node, so the hook never starts on Bluetooth mice
### Pre-flight checklist

- [x] I quit **Logi Options+** before running OpenLogi.

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.23

### Operating system

Linux

### OS version & architecture

Fedora 44, x86_64 (kernel 7.1.7-200.fc44), GNOME Shell 50.4 on Wayland, systemd 259

### Device model

MX Master 3S

### Affected area(s)

- [x] Button remapping

### What happened?

The bundled udev rules grant `uaccess` on `hidraw` and `uinput`, but not on the
mouse's `/dev/input/event*` node. The evdev hook needs to read that node to grab
the pointer, so on a Bluetooth-direct mouse it never starts and no remapping
works at all:

```
INFO  openlogi_agent: accessibility granted — installing OS mouse hook
WARN  openlogi_agent_core::hook_runtime: could not install OS mouse hook —
      events will not be captured
      error=no mouse device found under /dev/input; ensure a pointing device is
      connected and the process has read permission
WARN  openlogi_inject::inject::linux: failed to create uinput action device:
      Permission denied (os error 13)
```

Why Bluetooth specifically: a BT mouse hangs off `/devices/virtual/misc/uhid`,
which belongs to no seat, so logind's `uaccess` rules never tag it. A USB device
is tagged by those rules regardless of OpenLogi's own rules. Same root cause as
Finding 1 of #148, one layer up: #149 fixed it for `hidraw`, `input` was left out.

The mouse's event node, untagged:

```console
$ udevadm info -q all -n /dev/input/event21 | grep -E 'DEVPATH|TAGS|ID_BUS'
E: DEVPATH=/devices/virtual/misc/uhid/0005:046D:B034.0007/input/input30/event21
E: ID_BUS=bluetooth
# no TAGS line at all

$ stat -c '%n %U:%G %a' /dev/input/event21
/dev/input/event21 root:input 660
```

A USB keyboard on the same machine, for contrast:

```console
$ udevadm info -q all -n /dev/input/event7 | grep TAGS
E: TAGS=:power-switch:uaccess:seat:
```

Two things make this hard to self-diagnose:

1. **Settings → Permissions reports "granted".** The check added in #172 probes
   `/dev/uinput` and `/dev/hidraw*` only, so it is green while the hook is dead.
2. **The warning's suggested fix has a side effect.** It suggests joining the
   `input` group, which grants read access to *every* event node — the condition
   behind #484 (agent grabs the built-in trackpad) — and needs a re-login.
   Extending the udev rule keeps the grant scoped to vendor 046d.

Adding the two matchers already used for `hidraw` to `SUBSYSTEM=="input"` fixes
it without a re-login. PR follows.

### Steps to reproduce

1. Pair an MX Master 3S over Bluetooth, no receiver.
2. Install the udev rules from `packaging/linux/udev/70-openlogi.rules`,
   `udevadm control --reload-rules && udevadm trigger`, power-cycle the mouse.
3. `systemctl --user start openlogi-agent`.
4. `journalctl --user -u openlogi-agent | grep hook` → "could not install OS
   mouse hook".
5. Bind any button in the GUI and press it → nothing happens, while
   Settings → Permissions shows access as granted.

### Logs

After adding the `SUBSYSTEM=="input"` matchers and restarting the agent:

```
INFO  openlogi_agent_core::hook_runtime: OS mouse hook installed
```

and the agent holds the descriptors it needs:

```console
$ sudo ls -l /proc/$(systemctl --user show -p MainPID --value openlogi-agent)/fd \
    | grep -E 'input|hidraw|uinput'
12 -> /dev/uinput
15 -> /dev/input/event21
24 -> /dev/hidraw4
```

Xabierland:
## Summary

The evdev hook reads `/dev/input/event*` to grab the pointer, but the bundled
udev rules only cover `hidraw` and `uinput`. On a Bluetooth-direct mouse nothing
grants access to that third node, so the hook never starts and no remapping
works.

USB devices are unaffected: logind's seat rules tag them with `uaccess` on their
own. A Bluetooth mouse hangs off `/devices/virtual/misc/uhid`, which belongs to
no seat, so those rules skip it and the node stays `root:input 0660`.

This is the same transport asymmetry as Finding 1 of #148, one layer up. #149
fixed it for `hidraw`; `input` was left out, and #119 validated the hook with
`sudo`, which hid the gap.

## Changes

- `packaging/linux/udev/70-openlogi.rules`: add the two matchers already used
  for the `hidraw` block to `SUBSYSTEM=="input"`, so the grant stays scoped to
  vendor 046d instead of requiring `input` group membership.
- `docs/INSTALL-linux.md`: list the event node as a third permission OpenLogi
  needs, and add a `getfacl` check next to the existing `hidraw` one.
- `README.md`: mention the event node in the sentence describing what the
  package grants.

No Rust code touched.

## Testing

Verified on real hardware: MX Master 3S over Bluetooth (no receiver), Fedora 44,
kernel 7.1.7-200.fc44 x86_64, GNOME Shell 50.4 on Wayland, systemd 259,
OpenLogi 0.6.23.

Rule syntax:

```console
$ udevadm verify packaging/linux/udev/70-openlogi.rules
1 udev rules files have been checked.
  Success: 1
  Fail:    0
```

Before, with only the `hidraw` and `uinput` rules installed:

```console
$ getfacl -p /dev/input/event21 | grep -E '^user:'
user::rw-
$ journalctl --user -u openlogi-agent | grep hook
WARN could not install OS mouse hook — events will not be captured
     error=no mouse device found under /dev/input
```

After adding the matchers:

```console
$ sudo udevadm control --reload-rules
$ sudo udevadm trigger --action=add --subsystem-match=input
$ getfacl -p /dev/input/event21 | grep -E '^user:'
user::rw-
user:xabier:rw-
$ systemctl --user restart openlogi-agent
$ journalctl --user -u openlogi-agent | grep hook
INFO OS mouse hook installed
```

Button remapping then works end to end. Sniffing the injector node while
pressing a remapped gesture shows the synthesized chord arriving with correct
press/release ordering:

```
KEY_LEFTCTRL DOWN / KEY_LEFTALT DOWN / KEY_LEFT DOWN
KEY_LEFT up / KEY_LEFTALT up / KEY_LEFTCTRL up
```

No re-login and no `input` group membership were needed, and `udevadm trigger`
re-applied the ACL without power-cycling the mouse.

Not verified: Bolt/Unifying receivers and wired USB (no receiver on hand — those
paths already worked and the added rules are additive), other distributions,
and X11 sessions. `cargo fmt` / `clippy` / `test` were not run: the change is
packaging and docs only, no Rust.

Fixes #529

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,8 @@ sudo pacman -U openlogi-*.pkg.tar.zst
 Packages are published for both `x86_64`/`amd64` and `arm64`/`aarch64`.
 
 The package installs udev rules that grant your user access to
-`/dev/hidraw*` and `/dev/uinput` without `sudo`. After installation,
+`/dev/hidraw*`, `/dev/uinput` and your Logitech mouse's `/dev/input/event*`
+node without `sudo`. After installation,
 enable the background agent for your user:
 
 ```sh
--- a/docs/INSTALL-linux.md
+++ b/docs/INSTALL-linux.md
@@ -41,7 +41,7 @@ OpenLogi needs:
 - **Write access to `/dev/uinput`** — to create the virtual input device for
   button remapping.
 - **Read/write access to `/dev/hidraw*`** — to send HID++ commands to the Bolt
-  receiver.
+  receiver, or to the device itself when it is paired over Bluetooth.
 - **Read access to the mouse's `/dev/input/event*` node** — the hook grabs the
   pointer there to capture button presses. Bluetooth mice need the bundled rule
   for this: their event node hangs off `/devices/virtual/misc/uhid`, which has
--- a/docs/INSTALL-linux.md
+++ b/docs/INSTALL-linux.md
@@ -42,6 +42,10 @@ OpenLogi needs:
   button remapping.
 - **Read/write access to `/dev/hidraw*`** — to send HID++ commands to the Bolt
   receiver.
+- **Read access to the mouse's `/dev/input/event*` node** — the hook grabs the
+  pointer there to capture button presses. Bluetooth mice need the bundled rule
+  for this: their event node hangs off `/devices/virtual/misc/uhid`, which has
+  no seat, so `logind` never grants the ACL on its own.
 
 Install the bundled udev rules to grant access to the active-seat user without
 requiring `sudo` or group membership (requires `systemd-logind`):
@@ -61,6 +65,11 @@ openlogi-agent --check-uinput 2>/dev/null || \
 
 # Check a hidraw node
 ls -la /dev/hidraw*
+
+# Check the mouse's event node — look for a "+" (ACL) in the mode, or your
+# user in the ACL itself. Without it the agent logs
+# "could not install OS mouse hook".
+getfacl /dev/input/event*
 ```
 
 The GUI Settings → Permissions page shows a live `Granted` / `Not granted`
--- a/packaging/linux/install.sh
+++ b/packaging/linux/install.sh
@@ -83,6 +83,7 @@ if command -v udevadm > /dev/null 2>&1; then
     echo "Reloading udev rules …"
     sudo udevadm control --reload-rules
     sudo udevadm trigger --subsystem-match=hidraw
+    sudo udevadm trigger --subsystem-match=input
     sudo udevadm trigger --subsystem-match=misc --attr-match=name=uinput 2>/dev/null || true
 fi
 
--- a/packaging/linux/nfpm-scripts/postinstall.sh
+++ b/packaging/linux/nfpm-scripts/postinstall.sh
@@ -3,10 +3,13 @@ set -eu
 
 # Reload udev rules and wait for the new uaccess tags to be applied.
 # udevadm trigger is asynchronous — settle ensures the tags are in place
-# before the script exits so the agent can open /dev/hidraw* immediately.
+# before the script exits so the agent can open /dev/hidraw* and the mouse's
+# /dev/input/event* node immediately, even for a device connected before the
+# install.
 if command -v udevadm > /dev/null 2>&1; then
     udevadm control --reload-rules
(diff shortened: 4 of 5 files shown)


Xabierland:
Both review findings were valid — fixed in the three commits above.

**Keyboard exposure.** Agreed, and worth being precise about the blast radius:
for USB it changed nothing, since logind already tags every seat device with
`uaccess` (a USB keyboard's event node is readable by the session with or
without this rule). The real regression was Bluetooth, where the node has no
seat: a Logitech BT keyboard would have gone from unreadable to readable. The
rules now match `KERNEL=="event*", ENV{ID_INPUT_MOUSE}=="1"`, which is what the
hook actually needs since it grabs relative pointers.

**Install trigger.** Also correct, and I hit it myself: `udevadm trigger` only
ran for `hidraw` and `misc/uinput`, so an already-connected mouse kept its old
permissions. Added `--subsystem-match=input` to `install.sh` and
`postinstall.sh`.

**hidraw doc bullet.** Fixed — it credited the Bolt receiver alone.

Re-verified on the same setup (MX Master 3S over Bluetooth, Fedora 44, GNOME 50.4
Wayland) after replacing the rules file and re-triggering:

```console
$ getfacl -p /dev/input/event21 | grep '^user:'
user::rw-
user:xabier:rw-
$ journalctl --user -u openlogi-agent | grep hook
INFO OS mouse hook installed
```

The agent holds `/dev/input/event21`, `/dev/hidraw4` and `/dev/uinput`, and
remapping still works. `udevadm verify` passes on the rules file.

Issue #521, opened 2026-08-05
[Device]: Add Logitech MX Keys support


safonin:
### Pre-flight checklist

- [x] I quit **Logi Options+** before testing.
- [x] I searched existing issues for my device.

### Device model

Logitech MX Keys

### Connection

Bluetooth (direct, no receiver)

### What's the problem?

Device is detected but a feature doesn't work

### `openlogi list` output

```text
MX Keys (direct 046d:b35b)
  slot: 255 (direct device index 0xff)
  transports: equad + btle
  model IDs: b35b, 408a, 0000
```

### `openlogi diag features` output

```text
idx  id      ver  name
0    0x0000  v1   Root
1    0x0001  v1   FeatureSet
2    0x0003  v2   DeviceInformation
3    0x0005  v0   DeviceTypeAndName
4    0x1d4b  v0   WirelessDeviceStatus
5    0x0020  v0   ConfigChange
6    0x0007  v0   DeviceFriendlyName
7    0x1000  v1   BatteryStatus
8    0x1b04  v4   ReprogControls5
9    0x1814  v1   ChangeHost
10   0x1815  v1   HostsInfo
11   0x1982  v1   Backlight2
12   0x40a3  v0   FnInversionMultiHost
13   0x4100  v0   Encryption
14   0x4220  v0   LockKeyState
15   0x4521  v0   DisableKeys
16   0x4531  v1   MultiPlatform
17   0x00c2  v0   DfuControlUnsigned
```

### Details

The keyboard is detected over Bluetooth-direct, but several advertised HID++
features are not currently available through the full OpenLogi stack.

Test environment:

- macOS 26.6, Apple Silicon (`arm64`)
- OpenLogi development build based on 0.6.23
- Logi Options+ removed / not running
- Input Monitoring granted to the OpenLogi Agent

Current findings:

- Discovery and the device card work.
- The keyboard exposes legacy BatteryStatus `0x1000` v1, not UnifiedBattery
  `0x1004`, so the existing battery path needs a typed fallback.
- It exposes Backlight2 `0x1982` v1. Live read and the physical on/off toggle
  have been verified on the keyboard.
- It exposes FnInversionMultiHost `0x40a3` and DisableKeys `0x4521`; both need
  capability-aware live read/write/confirmation paths.
- Persistent Agent inventory already owns the working Bluetooth HID++ channel.
  Capture and route-addressed operations must reuse that channel instead of
  opening competing handles.

Proposed atomic PR series:

[...]

PR #532, opened 2026-08-09
fix(agent): route hardware operations through inventory channels

safonin:
## Summary

Route all Agent-side, route-addressed HID++ reads and writes through the exact inventory-owned channel introduced by #522. This avoids competing HID opens while the inventory or capture watcher already owns the device and provides the shared-channel foundation needed by the MX Keys work.

## Changes

- **openlogi-hidpp:** add bounded raw report writes for 64-byte `0x12` lighting frames and route `0x12` through the Windows long-report endpoint
- **openlogi-hid:** add existing-channel variants for DPI, SmartShift, and keyboard lighting; move per-key lighting reports onto the owned `HidppChannel`
- **openlogi-agent-core / openlogi-agent:** use a registry-confirmed capture channel or the exact current inventory channel for IPC calls, reconnect reapply, hook actions, and gesture actions
- keep the route-opening APIs available for standalone CLI/library callers; an Agent registry miss does not open a fallback channel

## Testing

- `cargo test --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p openlogi-hidpp -p openlogi-hid --no-deps`
- Hardware: these commits were exercised on an MX Keys over Bluetooth in the full feature branch; discovery, normal key input, and lighting on/off worked. The isolated rebased branch was not rerun on hardware.
- Windows: the `0x12` endpoint selection is unit-tested, but the change was not runtime-tested on Windows hardware.

Part of #521

### DIFF
--- a/crates/openlogi-agent-core/src/hardware.rs
+++ b/crates/openlogi-agent-core/src/hardware.rs
@@ -6,19 +6,17 @@
 //! press) and avoids holding a long-lived async runtime alongside GPUI's
 //! executor.
 //!
-//! When the HID++ capture session already has the target device open, these
-//! reuse that channel ([`openlogi_hid::CaptureChannel`]) instead of
-//! re-enumerating and opening a fresh one — the dominant cost of a write. The
-//! transient open is kept as a fallback for callers (e.g. the CGEventTap hook)
-//! firing while no session is connected.
+//! Agent calls select a registry-confirmed capture channel or the exact current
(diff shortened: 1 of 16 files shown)


davidbudnick:
@safonin Can you please fix the issues with the pipeline?

safonin:
sure thing
worked  on it

safonin:
Fixed in 6392871. The non-Windows `DeviceWriter` import is now properly cfg-gated. Windows clippy and all other checks are green.

PR #559, opened 2026-08-10
fix(ui): prevent middle and thumb wheel popover flicker

the-long-ride:
- Remove redundant p.patch file


the-long-ride:
Hi @davidbudnick ,

I'm sorry for redundant file that've been commited in my last PR.

When update OpenLogi to latest version -> I see the bug that when I click these button:
image
I see the flash of the options panel, I remember that I've already fixed in before commit but IDK why it occur again in release version.
So this is a fixed, maybe you can review this one soon.

Thank you!

Issue #586, opened 2026-08-13
Actions Ring haptics die under sustained hover: Bolt drops HID++ replies during pointer traffic


isleofgreg:
## Symptom

On an MX Master 4 over a Bolt receiver (macOS 15.7.9, current master), ring hover haptics work at first, then go silent after enough rapid hovering, and recover on their own after the pointer idles for a while. With `RUST_LOG=debug` the agent shows cascades of:

```
WARN openlogi_agent::server: Actions Ring haptic failed error=HID++ request timed out during PlayHaptic interaction="hover"
WARN openlogi_agent::server: Actions Ring haptic failed error=device at index 0x02 did not respond to HID++ interaction="hover"
```

During one of these episodes, unrelated HID++ traffic on the same receiver also times out (the GUI's DPI panel shows `DPI read failed: HID++ request timed out during ReadDpiCapabilities`), and after a device power-cycle the re-enumeration + volatile re-apply can stay degraded for minutes.

## Analysis

Two compounding factors:

1. `spawn_ring_haptic` spawns an unbounded task per hover, and each play costs three round-trips (`Device::new` ping → `open_feature` lookup → `play`). A fast pointer generates hover events faster than the receiver drains them, so the backlog behind `receiver_access` grows until every transaction on the channel times out.
2. The moment haptics fire is exactly the moment the pointer is moving — and the receiver visibly drops HID++ replies under concurrent motion-report traffic, so the retry-less single attempt fails most often precisely when the user is interacting with the ring.

## Mitigation stack we're running (happy to PR any/all of it)

Running on real hardware today, in this order of impact:

1. **Coalescing single-flight player** replacing the per-hover `tokio::spawn`: a `watch`-channel worker plays at most one waveform at a time; while one is in flight, newer requests overwrite the queued one (a stale hover buzz has no value). This alone stops the backlog collapse.
2. **Feature-handle cache** in `play_haptic_on`: reuse the opened `HapticFeedbackFeature` per (channel, device index) so the warm path is one round-trip instead of three; invalidate on any error and re-open once.
3. **Bounded retry** (≤3 attempts) in the worker, abandoned early if a newer request superseded the current one.
4. **Circuit breaker**: after 3 consecutive final failures, skip haptics for ~4s so the saturated channel can drain instead of being hammered while it's already losing replies — this also protects unrelated DPI/SmartShift writes.

[...]

PR #590, opened 2026-08-13
fix: Actions Ring haptic reliability — coalescing, feature cache, firmware arming, deadlock guards

isleofgreg:
Companion to #589; makes MX Master 4 Actions Ring haptics reliable on a lossy/shared receiver channel. Seven commits, each self-contained:

## Throughput & loss (commits 1–3)

- **Single-flight coalescing worker**: one waveform in flight, latest-wins queueing. Spawning a task per hover let a fast pointer queue plays faster than the receiver drains them, timing out every transaction on the channel for seconds (and starving unrelated DPI/SmartShift writes).
- **Feature-handle cache + bounded retry**: each play cost 3 HID++ round-trips (ping + feature lookup + play); on a busy receiver every round-trip is a fresh chance to lose the reply — exactly while the user hovers. Cache the opened `HapticFeedback` handle per (channel, device); warm path is 1 round-trip. Up to 3 attempts, abandoned when a newer request supersedes.
- **Circuit breaker**: 3 consecutive final failures → 4 s haptic cool-down so the shared channel drains instead of staying saturated.

## Firmware disarm (commits 4–5) — separate root cause, first noted at the end of #586

Nothing in OpenLogi ever asserts the device's haptic **enable** flag (`HapticFeedback` 0x19b0 `get`/`set_configuration`). Devices coast on whatever Logi Options+ last wrote, and some idle/power transitions clear it — after which `play` calls are **accepted with zero physical feedback** and no error. Explains "haptics died during light use, then came back on their own". The fix re-asserts enabled (repairing a zero intensity to 25) once per ring session, sequenced through the single-flight worker so it's guaranteed to complete before the session's first buzz (a free-running arm task raced — and lost to — the first hover).

## Deadlock guards (commits 6–7) — how "dead buttons until restart" survived #589's watchdog

The feature cache pins an `Arc<HidppChannel>`, and the enumerator only reopens a retired node once every clone of its channel has dropped. Two ways that pin turned a routine channel retire into permanently dead buttons on my machine:

1. Route resolution fails before any haptic I/O runs → the cache is never invalidated by its own error path → clear it on route-miss.
2. Worse: ring use repopulates the cache on the *current* channel → that channel is retired → inventory surfaces empty → capture stops → **no haptic attempt can ever happen again** (the ring trigger is itself a diverted control that died with capture) → the route-miss clear never runs. Caught live with a logged agent: the enumerator ticked every 2 s finding the node while `prepare_open` waited forever for the pinned Arc. So the cache entry is also released **at the moment of retirement** (probe-failure and absent-node paths), and the previously-silent reopen-blocked wait now logs at debug.

All soaked on the affected hardware (Bolt + MX Master 4, Sequoia 15.7.9): hover haptics stay solid under rapid hovering, first-hover buzz is reliable after idle, and the two deadlock incidents that previously required an agent restart now self-heal in seconds.


davidbudnick:
@isleofgreg Once you get it passing tag me and we can add it to the next release :)

isleofgreg:
> @isleofgreg Once you get it passing tag me and we can add it to the next release :)

Will do. I'm working on it. I've been working on this for the past 12 hours, trying to fix stuff as it arises. I want this project to succeed really bad since I migrated from LogiOptions+ and this is a way better option as long as we can get feature parity.

PR #598, opened 2026-08-14
fix(agent): bound the Actions Ring haptic worker to the session it serves

AprilNEA:
## Summary

Findings 3, 4 and 5 from the post-hoc review of the Actions Ring change set — all in the single-flight haptic worker, all ways for a ring to go silent while nothing looks broken.

**3 + 4 — attempts were bounded only by the write path's `WRITE_BUDGET` (5 s)**, which is sized for a user-visible setting write, not for feedback. Arming ran up to two of them *before* the session's first buzz (10 s in front of the first thing the user should feel) and a play ran up to three (15 s — exactly the ring's own `SESSION_LIFETIME` / `DISPLAY_LIFETIME`). The worker is single-flight and latest-wins, so that time is not just this buzz's: every interaction arriving during it is dropped, activation buzz included. Under the conditions #586 describes, a ring could open and close having produced no haptics at all — the symptom the retry loop was added to fix. The existing early-out only fires when a *newer* request arrives, so a user who hovers one slot and holds the pointer generates nothing to trigger it.

**5 — the circuit breaker never reset between sessions.** `consecutive_failures` and `cooldown_until` live for the worker's lifetime, so a ring that ended two failures deep primed the next one: its very first failure tripped a 4 s cool-down, silencing the opening of a session with nothing wrong with it. A cool-down opened at the tail of one ring did the same directly.

## Changes

**`crates/openlogi-agent`**
- Attempts now share one wall-clock budget per interaction: `ARM_BUDGET` (1 s) and `PLAY_BUDGET` (2 s). Each call is wrapped in a timeout for the budget's *remainder*, so the total is bounded even when a single attempt stalls — including inside `hardware::play_haptic`'s unbounded wait for the I/O lease, which the write path's own timeout never covered.
- A spent budget reads as `None` rather than `Some(ZERO)`: `timeout(ZERO, …)` elapses before the call it wraps can run, which would log a device failure that never happened.
- `arm` is the session boundary, so the worker clears both breaker counters when it sees one.
- The two loops move out of `spawn` into `arm_firmware_haptics` and `play_within_budget` (it had grown past the line limit), which also separates the breaker from the retry bookkeeping it was interleaved with.

Arming keeps its place ahead of the first buzz — that ordering is deliberate, per the comment there — it just can no longer consume the session.

## Testing

```
cargo fmt --all -- --check
cargo test --workspace
```

All green. `Budget` takes `now` as a parameter, so its tests are deterministic and need no sleeping:

- `a_partly_spent_budget_offers_only_the_remainder` — pins the sharing, which is the actual fix.
- `a_spent_budget_offers_nothing_rather_than_zero` — confirmed to fail when `remaining` is allowed to return zero.
- `the_budgets_stay_within_a_ring_session`.

Not runtime-tested on hardware. On a device: hover a slot on a busy Bolt and the ring should keep responding to later hovers and to the activation click instead of going silent for the rest of the session; and a ring opened right after a failing one should not start muted.

Related to #586. Independent of #596, #597 and #599, which touch the layers underneath.

### DIFF
--- a/crates/openlogi-agent/src/server.rs
+++ b/crates/openlogi-agent/src/server.rs
@@ -7,6 +7,7 @@
 
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::{Duration, Instant};
 
 use futures::StreamExt as _;
 use openlogi_agent_core::action_ring::ActionRingManager;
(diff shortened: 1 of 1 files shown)


davidbudnick:
@AprilNEA Any reason the Nix CI build has been taking a long time recently?

AprilNEA:
> @AprilNEA Any reason the Nix CI build has been taking a long time recently?

`buildRustPackage` compiles the whole tree, gpui included, in a single derivation, so every source change is a full rebuild; the cache can't touch it. Fixing it.

PR #610, opened 2026-08-14
refactor(core,inject): collapse inject backends onto a platform-neutral Effect IR

AprilNEA:
## Summary

- Adds `Action::effect()` in `openlogi-core`, a single exhaustive match that
  classifies the full 52-variant `Action` vocabulary into a ~10-variant
  platform-neutral `Effect` IR (`Click`, `Shortcut`, `Key`, `Scroll`,
  `Media`, `Native`, `Script`, `Text`, `AgentSide`, `None`).
- Each `openlogi-inject` backend (`macos.rs`/`linux.rs`/`windows.rs`) now
  matches on `Effect` instead of re-deriving the whole `Action` vocabulary
  three times in parallel, with a small per-backend `Shortcut -> KeyCombo`
  table for the ~14-16 shortcuts that are plain modifier+key chords.
- Pure refactor: every `Action` produces the exact same OS events on every
  platform it did before.

## Changes

**`openlogi-core`**
- `crates/openlogi-core/src/binding/effect.rs` (new): `Effect<'a>`,
  `Shortcut`, `MouseButton`, `MediaKey`, `NativeAction`, `Script<'a>`, and
  `impl Action::effect(&self) -> Effect<'_>` — the only new exhaustive match
  over `Action` outside of `label`/`category`/`catalog`.
- `binding.rs`: wire the new module in and re-export its public types.
- `binding/tests.rs`: totality test (every `Action::catalog()` entry lowers
  to a non-`None` effect except `Action::None`), a power-user/device-side
  bucket test (`CustomShortcut`/`TypeText`/`RunAppleScript`/
  `RunShellCommand`/`Workflow`/DPI-SmartShift-Ring-OpenApplication), and a
  scroll-direction test.

**`openlogi-inject`**
- `inject/macos.rs`: `execute()` collapses from 56 `Action::` references to
  a 10-arm `Effect` match, plus a 16-row `Shortcut -> KeyCombo` table
  (parsed through `KeyCombo`'s existing `FromStr`) and small
  `dispatch_click`/`dispatch_native`/`nx_key`/`dispatch_script` helpers.
- `inject/linux.rs`: same shape, 51 -> 10; adds `press_combo` shared by the
  `Shortcut` table and `CustomShortcut`/workflow `PressKey`.
- `inject/windows.rs`: same shape, 54 -> 8; `combo()` returns
  `Result<KeyCombo, u16>` because `BrowserBack`/`BrowserForward` fire a
  dedicated virtual key with no modifier on Windows (not a USB HID keyboard
  usage, so no `KeyCombo` representation at all) — the `Err` case is the
  one deliberate per-OS exception to the shared table shape, preserved
  exactly from the pre-refactor source. Also drops the duplicate local
  `MouseButton` enum in favor of the shared core one.
- Each backend gets a test pinning representative `Shortcut -> KeyCombo`
  rows (including the platform-specific ones: `Redo` is Ctrl+Y on Windows
  vs Ctrl+Shift+Z on macOS/Linux; `BrowserBack` is ⌘\[ / Alt+Left / a raw
  VK) plus a check that every table entry actually resolves through that
  backend's `hid_usage_to_*`, so a table edit can't silently no-op a
  shortcut.

## Testing

Full local gate on the final commit:
```
cargo fmt --all -- --check
cargo test --workspace
```
All green (149 openlogi-core tests including the new Effect tests; 3
openlogi-inject tests including the new macOS combo pinning test; no
regressions anywhere else in the workspace).

Cross-platform verification:
- **macOS**: compiled, clippy-clean, and tested natively (this machine).
- **Windows**: `cargo clippy --target x86_64-pc-windows-gnu -p
  openlogi-inject --all-targets -- -D warnings` passes on the real
  `windows-sys` dependency tree (check-only — no linker in this
  environment, so `cargo test` itself did not run). CI's `clippy-windows`
  job (native `windows-latest`) will be the first real native compile.
- **Linux**: **not compiled or tested locally** — no Linux Rust target is
  installed in this devenv shell, and `linux.rs` is `#[cfg(target_os =
  "linux")]`-gated out of every local build/clippy/test run on macOS.
  Verified entirely by hand-auditing every match arm against the
  pre-refactor source. CI's `test-linux` job (`ubuntu-latest`, real
  compile + `cargo test`) will be the first genuine verification.
- Not runtime-tested on physical hardware (no device I/O in this crate;
  the change is about OS event synthesis, verified by the `hid_usage_to_*`
  round-trip tests and manual chord-table review, not on real Logitech
  devices).

## Skipped

The stretch goal (folding `Action::label()`/`category()`/`catalog()` and
`ActionRingIcon::for_action()` into one declarative table) was left alone:
those four passes serve different audiences (backend dispatch vs. picker
UI strings vs. Actions Ring icons) and the English label text is itself an
i18n key across 20 locale files — collapsing them looked likely to either
touch GUI-facing strings (out of scope; a sibling agent owns
`crates/openlogi-gui`) or produce a much larger, harder-to-review diff for
a facet of the code the task called optional. The backend collapse above
is the full deliverable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5149,6 +5149,7 @@ dependencies = [
  "etcetera",
  "plist",
  "serde",
+ "strum",
  "tempfile",
  "thiserror 2.0.18",
  "toml 0.8.23",
--- a/crates/openlogi-core/Cargo.toml
+++ b/crates/openlogi-core/Cargo.toml
@@ -17,6 +17,7 @@ thiserror = { workspace = true }
 tracing = { workspace = true }
 etcetera = "0.11.0"
 atomic-write-file = { workspace = true }
+strum = { version = "0.27", features = ["derive"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 plist = "1.10.0"
--- a/crates/openlogi-core/src/binding.rs
+++ b/crates/openlogi-core/src/binding.rs
@@ -15,6 +15,7 @@ mod application_target;
 mod button;
 mod category;
 mod defaults;
+mod effect;
 mod gesture;
 mod key_combo;
 mod swipe;
@@ -33,6 +34,7 @@ pub use application_target::{ApplicationTarget, ApplicationTargetError};
 pub use button::ButtonId;
 pub use category::Category;
 pub use defaults::{default_binding, default_binding_for, default_gesture_binding};
+pub use effect::{Effect, MediaKey, MouseButton, NativeAction, Script, Shortcut};
 pub use gesture::GestureDirection;
 pub use key_combo::{KeyCombo, KeyComboParseError, KeyboardUsage, KeyboardUsageError};
 pub use swipe::{
--- a/crates/openlogi-core/src/binding/effect.rs
+++ b/crates/openlogi-core/src/binding/effect.rs
@@ -91,7 +91,7 @@ pub enum MouseButton {
 /// Each backend owns a `Shortcut -> KeyCombo` table (plus, for the couple of
 /// shortcuts a given OS has no ordinary chord for, a small override) rather
 /// than sharing one table — see the per-backend `combo`/`press_shortcut`.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, strum::VariantArray)]
 pub enum Shortcut {
     /// Copy the selection.
     Copy,
@@ -130,30 +130,14 @@ pub enum Shortcut {
 impl Shortcut {
     /// Every named shortcut, in declaration order.
     ///
-    /// The single shared iteration source for per-backend `Shortcut ->
-    /// KeyCombo` table-completeness tests, so those tests can't drift into
-    /// three independently-maintained copies of "every `Shortcut`
-    /// variant" — see `tests::all_is_exhaustive` below, which pairs this
-    /// with a wildcard-free match: adding a variant without extending
-    /// both fails to compile.
-    pub const ALL: [Shortcut; 16] = [
-        Shortcut::Copy,
-        Shortcut::Paste,
-        Shortcut::Cut,
-        Shortcut::Undo,
-        Shortcut::Redo,
-        Shortcut::SelectAll,
-        Shortcut::Find,
-        Shortcut::Save,
-        Shortcut::BrowserBack,
-        Shortcut::BrowserForward,
-        Shortcut::NewTab,
-        Shortcut::CloseTab,
-        Shortcut::ReopenTab,
-        Shortcut::NextTab,
-        Shortcut::PrevTab,
-        Shortcut::ReloadPage,
-    ];
+    /// `#[derive(strum::VariantArray)]` generates this straight from the
+    /// enum's variant list at compile time, so it cannot go stale the way a
+    /// hand-written array literal could: there is no second, independently
+    /// editable list for it to drift from — a variant that's missing here
+    /// would mean the enum itself doesn't have it. The single shared
+    /// iteration source for each backend's `Shortcut -> KeyCombo`
+    /// table-completeness test.
+    pub const ALL: &'static [Shortcut] = <Shortcut as strum::VariantArray>::VARIANTS;
 }
 
 /// A media/volume key.
@@ -291,44 +275,6 @@ impl Action {
             Action::RunAppleScript(src) => Effect::Script(Script::AppleScript(src)),
             Action::RunShellCommand(cmd) => Effect::Script(Script::ShellCommand(cmd)),
             Action::Workflow(steps) => Effect::Script(Script::Workflow(steps)),
-        }
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use super::Shortcut;
-
-    /// Pairs with [`Shortcut::ALL`]: this match has no wildcard arm, so
-    /// adding a `Shortcut` variant without adding it *here* fails to
-    /// compile — which is the prompt to also add it to `ALL` a few lines
-    /// up, instead of three per-backend test lists silently going stale.
-    #[test]
-    fn all_is_exhaustive() {
-        assert_eq!(
-            Shortcut::ALL.len(),
-            16,
-            "Shortcut::ALL grew or shrank without updating this test"
-        );
-        for shortcut in Shortcut::ALL {
-            match shortcut {
-                Shortcut::Copy
-                | Shortcut::Paste
-                | Shortcut::Cut
-                | Shortcut::Undo
-                | Shortcut::Redo
-                | Shortcut::SelectAll
-                | Shortcut::Find
-                | Shortcut::Save
-                | Shortcut::BrowserBack
-                | Shortcut::BrowserForward
-                | Shortcut::NewTab
-                | Shortcut::CloseTab
-                | Shortcut::ReopenTab
-                | Shortcut::NextTab
-                | Shortcut::PrevTab
-                | Shortcut::ReloadPage => {}
-            }
         }
     }
 }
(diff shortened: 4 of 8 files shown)


AprilNEA:
Pushed a fix for the clerical-coverage gap.

## Shortcut::ALL

Added `Shortcut::ALL: [Shortcut; 16]` next to the enum in `effect.rs`, paired with `tests::all_is_exhaustive` — a wildcard-free match over `Shortcut` sitting right next to it. Adding a new `Shortcut` variant without extending *both* fails to compile (E0004 on the match), which is the prompt to also extend `ALL` two lines up, instead of three independent per-backend test lists that nothing tied to the enum.

All three backends' `combo_table_pins_representative_shortcuts` tests now iterate `Shortcut::ALL` instead of their own hand-copied 16-entry list. Windows keeps its two explicit `BrowserBack`/`BrowserForward` pins (the raw-vk exceptions) and skips them inside the `ALL` loop via `combo()`'s `Err` case, rather than excluding them from the iteration source.

I did not add the equivalent for `NativeAction`/`MediaKey`/`MouseButton`: none of their per-backend dispatch functions do a two-step "resolve through a lookup that can miss" — each arm calls a fixed platform primitive directly (a dedicated VK, an NX key, an mpris command, a button code), so there's no analogous "silently unmapped" failure mode, and I never wrote a hand-copied variant list for them anywhere to begin with. `Shortcut` was the only vocabulary with that specific two-step KeyCombo → HID-usage-lookup shape.

Local gate re-run clean on the final tree: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, plus the Windows cross-lint (`cargo clippy --target x86_64-pc-windows-gnu -p openlogi-inject --all-targets -- -D warnings`).

## On the line-count growth

Final numbers, this branch vs. base: `macos.rs` 1130→1166 (+36), `linux.rs` 591→664 (+73), `windows.rs` 313→387 (+74), plus the wholly-new `effect.rs` (334 lines) and `+78` in `binding/tests.rs`. Net +597 across the six touched files. So: yes, it grew, and "delete ~72 match arms" undersold what actually happened to raw line count — that framing was accurate for the *dispatch* match specifically (each backend's top-level `execute()` did shrink from ~50-56 `Action` arms to a flat 8-10 `Effect` arms) but didn't account for three other things that are additive, not neutral:

1. **New rustdoc that didn't exist before.** The old code's only documentation of "what is BrowserBack" was an inline `//` comment on one match arm, repeated with slightly different wording in each of the three files. `effect.rs` now carries one rustdoc per variant across five new public enums (~50 variants) — that's real prose that has to live *somewhere*, and it didn't get smaller by moving from "three inline comments" to "one doc comment," it got more complete (each of the three inline comments was terser and platform-specific; the shared doc comment has to explain the concept once, precisely, for all three readers).
2. **Genuinely new test coverage.** Before this PR, zero tests verified that `BrowserBack` was `Cmd+[` on macOS or `Alt+Left` on Linux — the chord tables were untested inline literals. The pinning tests + (now) the `ALL`-driven completeness tests are net-new correctness coverage, not restructured existing coverage. That's the majority of the per-backend growth, honestly (~15-25 lines per backend of tests where there were 0-14 before).
3. **Fixed per-function overhead.** Splitting one flat 50-arm match into `execute` + `dispatch_click`/`combo`/`dispatch_native`/`dispatch_media`/`dispatch_script` costs a signature + (where non-obvious) a doc comment + braces per function, which a single match doesn't pay. I could claw a few lines back by inlining the smaller helpers into `execute` directly, but that re-tangles mechanism-dispatch with sub-vocabulary dispatch, which is the exact thing the refactor exists to undo — I don't think that's a good trade for a handful of lines.

None of that is padding I'd cut on a second pass. The actual win isn't line count, it's *where the risk lives*: before, "does `BrowserBack` mean the same thing on all three platforms" required diffing three ~50-arm matches by eye; now it's one 16-row table per backend, same shape, sitting next to a test that pins the values and (after this fix) a compile error if a row goes missing. I'd stand behind the coupling win; I'd just phrase the pitch next time as "shrinks the diff you have to trust by eye," not "deletes N lines."

AprilNEA:
## 1. The `n()` name

I can't find a `fn n() {` in what's on this branch — `git show <every-commit-so-far>:crates/openlogi-core/src/binding/effect.rs` has always had it as `fn all_is_exhaustive()`, and that's what's in the diff GitHub shows too. Possibly a stale render on your end (or you were looking at a different fetch of the branch) — either way, moot now: that whole function is gone as of the fix below, along with the doc reference to it.

## 2. The guarantee didn't hold — you're right, and so is your specific counterexample

I traced it through and it's real, including with your exact proposed `index()`/`seen[]` shape. The failure mode for both:

- `Shortcut::Foo` is added to the enum.
- The wildcard-free match (`all_is_exhaustive`, or your `index()`) fails to compile. Real tripwire.
- Dev adds `Foo => {}` (or `Foo => 16`) to fix *that* compile error.
- The test still only iterates `Shortcut::ALL`'s actual elements. `Foo` was never added to `ALL`, so it's never iterated, so `index(Foo)`/the match's `Foo` arm is never exercised by anything at runtime. `seen` ends up fully `true` from the *other* 16 elements alone; the `assert!` passes; `ALL.len()` never changes.

The root issue: `ALL` (an array literal) and the completeness-check (a match, wherever it lives) are always two independently-editable artifacts in hand-rolled Rust. Exhaustiveness-checking one proves nothing about the other's contents — there's no way to make "was this fixed by editing the array" mandatory when the match alone is what's forcing the compile error. I checked whether any variant of the array-bounds trick closes this (yours included, plus a few shapes of my own) and couldn't find one that survives "add the match arm without touching the array," which is the cheapest fix and exactly what a rushed PR would do.

**Fix:** dropped the array literal and `all_is_exhaustive` entirely. `Shortcut` now derives `strum::VariantArray`:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, strum::VariantArray)]
pub enum Shortcut { /* ... */ }

impl Shortcut {
    pub const ALL: &'static [Shortcut] = <Shortcut as strum::VariantArray>::VARIANTS;
}
```

I read the actual macro (`strum_macros-0.27.2/src/macros/enum_variant_array.rs`) rather than trusting memory: it walks `ast.data`'s variants directly off the enum's syntax tree at macro-expansion time and emits `const VARIANTS: &'static [Self] = &[Name::V1, Name::V2, ...]`. There is no second, independently-maintained list — the "array" and the enum declaration are the same source. Adding `Shortcut::Foo` makes `Shortcut::ALL` include it automatically, on every build, with no action required and no way to get it wrong. That's a property no hand-rolled scheme can match on stable Rust without this kind of derive.

`strum` was already resolving to `0.27.2` in the workspace (a direct dep of `xtask`, transitively pulled by several others), so I pinned `openlogi-core`'s new dependency to the same `"0.27"` line instead of letting `cargo add` grab `0.28` and split the resolved version — `Cargo.lock` diff is one line (`openlogi-core` gains `strum` in its dependency list; no new package versions).

All three backends' `combo_table_pins_representative_shortcuts` tests still iterate `Shortcut::ALL` (now `&'static [Shortcut]`, so `for &shortcut in Shortcut::ALL`).

## Verified

- Read the actual `strum_macros` derive source rather than assuming its behavior.
- Full local gate on the final tree: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — all green.
- `cargo clippy --target x86_64-pc-windows-gnu -p openlogi-inject --all-targets -- -D warnings` and `devenv tasks run openlogi:check-windows` (which also pulls in `openlogi-inject` transitively via `openlogi-agent`) — both green, confirming `strum` cross-compiles cleanly.
- `gpui`/`gpui-component` pins in `Cargo.lock` unchanged (still `1a246efd`/`031555662e99a1b5a549990b47f246d475b8288a`) after the dependency change.

PR #611, opened 2026-08-14
refactor(core): move pure HID++ wire types into openlogi-core, drop openlogi-hid from the GUI

AprilNEA:
## Summary

- `crates/openlogi-gui` depended on `openlogi-hid` (and thus the whole HID++
  stack — `async-hid`, `hidpp`, ~13k lines of transport/inventory code) even
  though it only ever used pure data types from that crate, never any I/O.
- Moves those pure types into `openlogi-core::hid` and drops the
  `openlogi-hid` path dependency from `crates/openlogi-gui/Cargo.toml`
  entirely, turning "the GUI is a pure IPC client" (CLAUDE.md's stated
  architecture) from a convention into a compile-time guarantee for the
  GUI's own dependency line.
- Follow-up round: also moved `PairingError` (missed in the first pass) and
  pointed `openlogi-agent-core/src/ipc.rs` — the IPC contract module — at
  `openlogi_core::hid` directly instead of at `openlogi-hid`'s re-exports.

## Changes

**`openlogi-core`** — new `hid` module (`src/hid.rs` + `src/hid/{route,dpi,error,smartshift,pairing,light}.rs`), pure data only, no `hidpp`/`async-hid` dependency:
- `hid::route`: `DeviceRoute`, `DIRECT_DEVICE_INDEX`, `BOLT_PIDS`/`UNIFYING_PIDS`/`LIGHTSPEED_PIDS`, `speaks_unifying_protocol`, `is_receiver_pid`, `receiver_display_name` — everything from `openlogi-hid`'s `route.rs` except the actual channel-opening I/O.
- `hid::dpi`: `DpiCapabilities`, `DpiInfo`.
- `hid::error`: `WriteError`, `HidppOperation`, `HidppFeatureErrorKind`.
- `hid::smartshift`: `SmartShiftMode`, `SmartShiftStatus`, `AUTO_DISENGAGE_PERMANENT`.
- `hid::pairing`: `Click`, `PasskeyMethod`, `ReceiverSelector`, `PairingError`.
- `hid::light`: `LightCommand`, `commands_for_light_settings`.
- Added `num_enum` as a dependency (needed by `SmartShiftMode`'s discriminant derive; already a workspace dependency via `openlogi-hid`).

**`openlogi-hid`** — every moved type is re-exported unchanged (`pub use openlogi_core::hid::...`), so this crate's own module paths (`crate::route`, `crate::smartshift`, `crate::pairing`, `crate::write::*`) and every downstream call site (agent-core, agent, cli) resolve exactly as before:
- `route.rs` keeps only `open_route_channel` (I/O).
- `smartshift.rs` deleted — was 100% pure, replaced by `pub use openlogi_core::hid::smartshift;` in `lib.rs`.
- `pairing.rs` keeps the pairing session (discovery, notifications, register writes); the four wire/error types are re-exported.
- `write/dpi.rs`, `write/litra.rs`, `write.rs` keep their I/O (`get_dpi`/`get_dpi_info`/`set_dpi`, the Litra raw-HID driver, `with_route`) and re-export the moved types.
- Two `impl From<async_hid::HidError> for ...` impls (`WriteError`, `PairingError`) could not move with their types — from `openlogi-hid` they'd implement a foreign trait for a foreign type, violating Rust's orphan rule. Each replaced with a local `classify_hid_error(&HidError) -> _` function (named to match the crate's existing `classify_hidpp_error`/`classify_dpi_error` convention, with a doc comment on *why* it's a free function and not a `From` impl, so it doesn't get "fixed" back into an illegal impl later).

**`openlogi-agent-core`** — `src/ipc.rs` (the tarpc service contract) no longer names `openlogi_hid` at all: its `use openlogi_hid::{...}` (nine types) is now `use openlogi_core::hid::{...}`. This is preparation, not a build-graph change — `openlogi-agent-core` still depends on `openlogi-hid` unconditionally for `hardware.rs`'s agent-side orchestration — but it makes the eventual split (see below) mechanical for that one module instead of requiring another type-by-type audit.

**`openlogi-gui`** — removed the `openlogi-hid` path dependency from `Cargo.toml`; every `use openlogi_hid::X` (and the few inline `openlogi_hid::X` paths) rewritten to `openlogi_core::hid::X`. No logic changes, no locale changes.

## Wire format

`crates/openlogi-agent-core/tests/wire_format.rs` passes with its goldens **completely unchanged**, and `PROTOCOL_VERSION` stays at `16`. This is the expected and correct outcome: bincode encodes struct field order / enum variant order, never a Rust module path, so moving a type between crates cannot change a single byte on the wire as long as field/variant order is preserved verbatim — which every moved type is (copy-pasted, not "tidied"). `PairingError` itself isn't a wire type (it feeds `impl From<PairingError> for PairingFailure` in `ipc.rs`, and `PairingFailure` is the actual wire type), but the goldens are the same proof that the conversion's match arms still map correctly after the move.

## Scope note: one remaining indirect edge

`crates/openlogi-gui/Cargo.toml` no longer names `openlogi-hid` directly, but `cargo tree -p openlogi-gui -i openlogi-hid` still shows one indirect edge: `openlogi-hid → openlogi-agent-core → openlogi-gui`. The GUI depends on `openlogi-agent-core` for the IPC client types (`ipc::AgentClient`, etc.), and `openlogi-agent-core` itself unconditionally depends on `openlogi-hid` for the **agent-side** hardware orchestration (`hardware.rs`, `dpi.rs`, `capture_plan.rs`, ...). Fully severing that edge requires splitting `openlogi-agent-core` into an agent-only (hid-dependent) part and a shared IPC-contract-only part — that touches `hardware.rs`/`server.rs`, which are explicitly out of scope for this branch (owned by a concurrent sibling branch introducing a `DeviceOp` receiver type, which looks positioned to do exactly this split).

For that follow-up to be scoped accurately: the GUI's full surface into `openlogi-agent-core` beyond `ipc` is `transport` (4 uses), `device_order` (2), `bindings` (1), `action_ring` (1 const), plus one rustdoc-only link to `hardware` (not a call). So the eventual shared/leaf crate is a bit more than `ipc.rs` alone — this PR only finished the `ipc.rs` piece of that inventory.

This PR's own, direct manifest edge is gone and verified; the transitive one is a pre-existing structural fact this branch does not touch, by design.

## Testing

```
direnv exec . cargo test -p openlogi-agent-core --test wire_format   # 10 passed, goldens untouched
direnv exec . cargo test -p openlogi-agent-core                      # 103 passed
direnv exec . cargo test -p openlogi-core --lib                      # 162 passed
direnv exec . cargo test -p openlogi-hid --lib                       # 163 passed
direnv exec . cargo test -p openlogi-gui --bin openlogi-gui           # 126 passed
direnv exec . cargo fmt --all -- --check
direnv exec . cargo clippy --workspace --all-targets -- -D warnings
direnv exec . cargo test --workspace                                  # full suite green, same counts as master
direnv exec . devenv tasks run openlogi:check-windows                 # cross-lint for the touched transport.rs cfg-gated file
direnv exec . cargo tree -p openlogi-gui -i openlogi-hid              # direct edge gone; indirect edge documented above
```

`Cargo.lock`: only the `openlogi-core -> num_enum` edge added and the removed `openlogi-gui -> openlogi-hid` edge; the `gpui`/`gpui-component` git pins are untouched (checked via `git diff origin/master...HEAD -- Cargo.lock`).

Not runtime-tested on hardware — this is a pure type-location refactor with no behavior change; the wire-format goldens are the load-bearing proof here, not manual hardware verification.

## Merge order

This branch should merge **last**, after the three sibling refactors (`openlogi-inject` + `binding` IR, the `DeviceOp` receiver type in `agent-core`/`agent`, and the GUI `state*` consolidation) land, since it touches import lines across `openlogi-hid`, `openlogi-core`, `openlogi-agent-core`, `openlogi-agent`, `openlogi-cli`, and `openlogi-gui`.


AprilNEA:
Follow-up filed as #614: extracting the IPC contract into a leaf crate so the GUI's build graph actually loses `openlogi-hid` / `hidpp` / `async-hid`. This PR makes the "GUI is a pure IPC client" contract a compile-time guarantee; #614 is what turns it into a build-time win. Blocked on this and the three sibling refactors merging.

Issue #485, opened 2026-07-31
[Bug]: SmartShift config write intermittently returns InvalidArgument on first attempt after agent start (0x2110/0x2111, MX Master 3S over Bolt)


cristbc:
### Pre-flight

- [x] Searched existing issues; closest neighbours are #385, #382 and #468 but none cover this.
- [x] On the latest release (0.6.22).
- [x] Logi Options+ is not installed on this machine at all, so nothing is contending for the receiver.

### Which part of OpenLogi?

Both (agent writes it; CLI `diag` was used to cross-check).

### OpenLogi version

0.6.22

### Operating system

macOS

### OS version & architecture

macOS 26.5.2 (build 25F84), Apple Silicon (arm64)

### Device model

MX Master 3S + Logi Bolt receiver (an MX Mechanical keyboard is paired to slot 1 of the same receiver)

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- SmartShift
- Settings / configuration (TOML)

### What happened?

The agent's SmartShift config-apply write intermittently fails on its **first** attempt after agent start, returning `InvalidArgument` from the device:

```
WARN openlogi_agent_core::hardware: SmartShift write failed
     error=HidppFeature { operation: WriteSmartShift, feature_hex: 8465, kind: InvalidArgument }
```

The agent then retries and the retry succeeds, so nothing is user-visible in normal use. I'm filing it because it is reproducible at a decent rate, it hits **both** SmartShift features, and the parameters are identical between the failing attempt and the succeeding retry, which points at a transaction-level race rather than a bad payload.

Measured across 15 agent starts, 27 write attempts, with `OPENLOGI_LOG=debug`:

| outcome | count |
|---|---|
| write succeeded | 22 |
| `InvalidArgument` on first attempt, retry succeeded | 5 |
| never wrote (device already matched config) | 1 start |

Of the 5 failures, 3 were on `feature_hex: 8465` (`0x2111`, SmartShiftWheelEnhanced) and 2 on `feature_hex: 8464` (`0x2110`, SmartShiftWheel). Failures cluster rather than distribute evenly: one block of 4 consecutive starts produced 3 failures, while a later block of 10 produced 2.

Same-run pairing is the interesting part. Here the first attempt is rejected and the retry, with byte-identical parameters, is accepted:

```
[...]

PR #612, opened 2026-08-14
refactor(agent): collapse hardware.rs's duplicated write skeletons into DeviceOp

AprilNEA:
## Summary

`crates/openlogi-agent-core/src/hardware.rs` had two families of near-identical
boilerplate: seven fire-and-forget `*_in_background` functions (each ~35 lines:
None-target guard → resolve → thread spawn → one-shot runtime → lease → timeout
→ 3-arm log) and seven awaited `apply_*`/`read_*`/`play_haptic`/
`ensure_ring_haptics_armed` functions sharing the same `(capture, registry,
receiver_access)` triple that `server.rs` spelled out by hand at every call
site. Both collapse onto a new `DeviceOp` type — the receiver-side counterpart
of `openlogi-hid`'s `with_route` "boilerplate-eater".

## Changes

- **`openlogi-agent-core`**
  - **Behavior fix, in its own commit** (`fix(agent): bound the background
    lighting write by WRITE_BUDGET`): `set_lighting_in_background` awaited its
    HID++ write with no timeout at all — unlike every sibling
    `*_in_background` writer — so an asleep/unresponsive keyboard leaked that
    background thread forever. Fixed independently of the `DeviceOp` refactor
    so it's revertible/cherry-pickable on its own and shows up in the
    release-plz changelog as a `fix`, not buried inside a `refactor` commit.
  - Add `DeviceOp<'a>` (`hardware.rs`): binds a `DeviceRoute` to the runtime's
    capture/inventory channels, then either `run()`s once (awaited, leases
    before resolving) or `detach()`es fire-and-forget on its own OS thread with
    standard 3-arm logging. `SharedRuntime::device`/`keyboard_device`
    (`orchestrator.rs`) build one from the shared runtime.
  - Every `*_in_background` function now goes through `DeviceOp`, except
    `reapply_mouse_volatile_in_background`, which stays hand-rolled and now
    takes `&DeviceOp` (it only reads fields, never hands the op to
    `run`/`detach`/`spawn_write`) — it must keep DPI/SmartShift/wheel-mode
    sequential on one OS thread with one resolved channel and one held lease
    (issue #485). Its `#[expect(clippy::too_many_arguments)]` is gone now that
    the four loose capture/registry/receiver_access/route params collapsed
    into one `&DeviceOp`.
  - **Diagnostics preserved, not just collapsed** (second `fix(agent):`
    commit): the generic `DeviceOp::detach` outcome log carries only `label`,
    which would silently have dropped every value the old bespoke logging
    carried (the written DPI, the SmartShift mode/auto_disengage/
    tunable_torque, the Fn-lock `on`, the lighting `r`/`g`/`b`). The five
    writes that carry an interesting value call `DeviceOp::spawn_write`
    directly (the same seam `write_scroll_wheel_mode_in_background` already
    used for its own custom logging) with their own 3-arm outcome closure
    instead of the generic `detach`, so `"DPI written to device"`,
    `"SmartShift config written"`, `"Fn-lock written"`, and `"lighting written
    to keyboard"` are byte-for-byte the log lines master had before this PR.
    `detach` itself is unchanged (still used by writes with nothing extra to
    log, and by its own registry-miss test).
  - `apply_dpi`, `apply_smartshift`, `apply_lighting`, `read_dpi`,
    `read_smartshift`, `play_haptic`, `ensure_ring_haptics_armed` are removed;
    their single call sites in `server.rs` now call `DeviceOp::run` directly.
    Their DPI-range check and lighting RGB scaling survive as two small `pub`
    helpers, `dpi_wire_value` and `lighting_rgb`.
  - `orchestrator.rs` / `hook_runtime.rs` call sites updated to the new
    `SharedRuntime::device(&route)` / `keyboard_device(&route)` builders.
  - Three new tests on the `DeviceOp` seam (see Testing).
  - `write_smartshift_in_background`, `read_dpi_info_blocking`, and
    `read_smartshift_status_blocking` have zero callers anywhere in the
    workspace (the latter two predate the GUI's move to IPC-only device
    reads, it looks like). Left in place, not deleted here — tracked as #615.
  - Net: `hardware.rs` 852 → 790 lines, despite adding the new `DeviceOp` type
    with full rustdoc, its tests, and the restored per-value logging.
- **`openlogi-agent`**
  - `server.rs`'s IPC handlers (`set_dpi`, `set_lighting`, `set_smartshift`,
    `read_dpi`, `read_smartshift`) and the Actions Ring haptic worker
    (`arm_firmware_haptics`, `play_within_budget`) now call
    `self.shared.device(&route).run(HidppOperation::…, |c| async move { … })`
    directly instead of the removed `hardware::apply_*`/`read_*` wrappers.

### Preserved behavior (audited per-function, not normalized silently)

1. **Lease-vs-resolve ordering unchanged for every function.** The seven
   `apply_*`/`read_*`/`play_haptic`/`ensure_ring_haptics_armed` functions were
   *all* already lease-then-resolve; `DeviceOp::run` does the same
   (lease → resolve → `timed`), so nothing changed there. The seven
   `*_in_background` functions were all resolve-on-the-caller's-thread, *then*
   spawn a thread that leases; `DeviceOp::detach`/`spawn_write` preserve that
   exact order too. No function's ordering was unified or flipped.
2. `reapply_mouse_volatile_in_background` still resolves once, holds one lease,
   and runs DPI/SmartShift/wheel-mode sequentially on the one OS thread it
   spawns — confirmed by re-reading the diff line by line; it deliberately does
   not decompose into three `detach`/`spawn_write` calls.
3. `authoritative_channel`'s registry-miss path still calls
   `openlogi_hid::clear_haptic_feature_cache()` before returning
   `DeviceNotFound` — untouched, and now covered by a test (see Testing).
4. `WRITE_BUDGET` now wraps *every* background write, including
   `set_lighting_in_background` (see the isolated `fix:` commit above).
5. DPI-range rejection kept its two distinct shapes: `dpi_wire_value` (used by
   `server.rs`'s `set_dpi`) returns the typed
   `WriteError::HidppFeature{OutOfRange}`; `write_dpi_in_background` still logs
   and skips instead of erroring.
6. `set_dpi_on_channel`'s read-back verification (untouched, lives in
   `openlogi-hid`) and `log_wheel_result`'s `FeatureUnsupported`-is-`debug`-not-
   `warn` distinction are both intact and unchanged — wheel-mode writes still
   route through `log_wheel_result` exactly as before.

## Testing

```
direnv exec . cargo fmt --all -- --check
direnv exec . env CARGO_BUILD_JOBS=8 cargo clippy --workspace --all-targets -- -D warnings
direnv exec . env CARGO_BUILD_JOBS=8 cargo test --workspace
direnv exec . env CARGO_BUILD_JOBS=8 cargo test -p openlogi-agent-core --test wire_format
```

All green on the final tree. `wire_format` goldens are byte-for-byte unchanged
(no `PROTOCOL_VERSION` bump needed — this branch never touches a type that
crosses the IPC wire). `orchestrator/tests.rs` (103 pre-existing agent-core
unit tests, including every `orchestrator::tests::*` and the three
`hardware::tests::*` `choose_authoritative` tests) pass unchanged.

Three new tests exercise real `DeviceOp` logic (not tautological mirrors of the
implementation):
- `run_on_a_registry_miss_returns_device_not_found_without_calling_f` /
  `detach_on_a_registry_miss_never_calls_f` — a registry miss propagates
  `WriteError::DeviceNotFound` and never invokes the caller's write closure.
- `timed_maps_an_elapsed_deadline_to_request_timed_out` — an elapsed
  `WRITE_BUDGET` maps to `WriteError::RequestTimedOut { operation }`, using a
  paused tokio clock (`tokio/test-util`, dev-dependency only) so the test
  advances virtual time instead of sleeping 5 real seconds.

A real end-to-end `DeviceOp::run` timeout test (resolve succeeds, then the
write itself times out) isn't possible from this crate: `SharedChannel` has no
public constructor outside `openlogi-hid`, so there's no way to make
`resolve()` succeed without a real device open.

**Not runtime-tested on hardware** — the `*_in_background` family is
fire-and-forget and hardware-bound; verifying the lease/resolve/timeout paths,
the sequential volatile-reapply write, and the fixed lighting timeout gap all
need a physical device.

### DIFF
--- a/crates/openlogi-agent-core/Cargo.toml
+++ b/crates/openlogi-agent-core/Cargo.toml
@@ -29,3 +29,6 @@ workspace = true
 bincode = "1.3"
 # Only to construct an `InventoryError::Hid` in the watcher's classify tests.
 async-hid = { workspace = true }
+# Only to advance virtual time in the DeviceOp/timed timeout test without a
+# real multi-second sleep.
+tokio = { workspace = true, features = ["test-util"] }
--- a/crates/openlogi-agent-core/src/hardware.rs
+++ b/crates/openlogi-agent-core/src/hardware.rs
@@ -1,14 +1,21 @@
 //! Hardware-side actions invoked from both the GPUI thread (slider release)
 //! and the OS-event hook thread (bound button press).
 //!
-//! Each call spawns a one-shot tokio runtime on a dedicated OS thread —
-//! cheap at the cadence these fire at (≤ once per slider release / button
-//! press) and avoids holding a long-lived async runtime alongside GPUI's
-//! executor.
+//! [`DeviceOp`] is the seam every device write and read goes through: it binds
+//! a [`DeviceRoute`] to this runtime's capture/inventory channels (built via
+//! [`crate::orchestrator::SharedRuntime::device`] or
+//! [`crate::orchestrator::SharedRuntime::keyboard_device`]), then either
+//! awaits [`DeviceOp::run`] (the IPC server's reads/writes, which must report
+//! their result to the GUI) or fires [`DeviceOp::detach`] (the OS-hook and
+//! reconnect paths, which must never block their caller). Both resolve the
+//! channel the same way every caller always did — a registry-confirmed
+//! capture channel or the exact current inventory channel; a registry miss is
+//! unavailable, never a fallback to re-enumerating and opening a competing
+//! connection.
 //!
-//! Agent calls select a registry-confirmed capture channel or the exact current
-//! inventory channel. A registry miss is unavailable; the daemon never falls
-//! back to re-enumerating and opening a competing connection.
+//! `detach` spawns a one-shot tokio runtime on a dedicated OS thread — cheap
+//! at the cadence these fire at (≤ once per slider release / button press)
+//! and avoids holding a long-lived async runtime alongside GPUI's executor.
 
 use std::future::Future;
 use std::time::Duration;
@@ -18,6 +25,7 @@ use openlogi_hid::{
     CaptureChannel, ChannelRegistry, DeviceRoute, DpiInfo, HapticWaveform, HidppFeatureErrorKind,
     HidppOperation, ScrollResolution, SharedChannel, SmartShiftMode, SmartShiftStatus, WriteError,
 };
+use tokio::time::error::Elapsed;
 use tracing::{debug, warn};
 
 use crate::receiver_access::ReceiverAccess;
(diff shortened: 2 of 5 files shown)


AprilNEA:
Follow-up filed as #615: removing `write_smartshift_in_background`, `read_dpi_info_blocking`, and `read_smartshift_status_blocking`, which this PR confirmed are dead (0 callers; the one remaining reference is a rustdoc link from one dead function to another). Deliberately out of scope here — filed rather than folded in, so the `DeviceOp` migration stays reviewable.

PR #671, opened 2026-08-19
fix(agent): reload control capture across app switches

Phecda:
## Summary

Foreground-app changes previously stopped and rebuilt the HID++ control-capture session so the new app-specific configuration could take effect. On the tested MX Master 3, rebuilding left a gap of more than one second between `control capture stopped` and `control capture active`.

During that gap, diverted controls temporarily returned to their firmware-native behavior. In practice, using the auxiliary button could open the macOS application switcher instead of executing the active OpenLogi binding.

This change keeps the existing capture channel, listener, and session alive across normal app switches and reloads the complete app-specific `CaptureSpec` in place.

## Changes

- Agent core: publish each complete app-specific `CaptureSpec` to the active capture session instead of restarting it.
- HID: incrementally transition controls between native, plain-diverted, raw-XY, and thumb-wheel modes.
- HID: avoid hardware writes when only the action assigned to an already-captured control changes.
- HID: keep the listener active and retry the latest specification after a failed HID++ reload.
- Agent core: continue restarting capture when the route or re-arm generation changes, pairing takes ownership, the device is removed, or the capture channel dies.
- Tests: cover per-app capture-plan differences, session continuity across specification changes, requested control modes, and gesture-state cleanup when a source is removed.

## Behavior

Before this change:

- Switching the frontmost app stopped the current capture session.
- Diverted buttons were restored to native behavior while a replacement session was created.
- The observed stop/start gap could exceed one second.
- Input during that gap could trigger the device or macOS default action instead of the configured OpenLogi action.

After this change:

- Normal app switches produce `control capture reloaded` without `control capture stopped`.
- The capture channel, event listener, and session remain active.
- The new app-specific rules take effect without the previous stop/start gap.
- Only hardware capture modes that actually changed require HID++ writes; action-only changes require none.

## Testing

- `cargo test --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p openlogi-hid -p openlogi-hidpp -p openlogi-hidpp-derive --no-deps --document-private-items`
- Runtime-tested with an MX Master 3: repeated foreground-app switches logged `control capture reloaded` without `control capture stopped`, and the configured auxiliary-button behavior remained active across app switches.
- The final tree includes a control-flow-only simplification made after the hardware test and has passed the full automated gate.

Related to #878

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -32,9 +32,9 @@ pub struct DeviceCapturePlan {
     /// keyed by the button its captured swipes dispatch as; empty when none
     /// gestures.
     pub gesture_bindings: BTreeMap<ButtonId, BTreeMap<GestureDirection, Action>>,
-    /// Standard buttons whose device or per-app binding leaves the default —
-    /// divert over `0x1b04`. A button that stays at its default everywhere
-    /// keeps its native HID behavior, so no re-synthesis is ever needed.
+    /// Standard buttons whose binding leaves the default — divert over
+    /// `0x1b04`. A button at its default keeps its native HID behavior, so no
+    /// re-synthesis is ever needed.
     pub divert_buttons: Vec<(u16, ButtonId)>,
     /// Whether any thumbwheel binding leaves its default. Combined with the
     /// sensitivity to decide thumb-wheel diversion.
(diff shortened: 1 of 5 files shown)


joaovvrodrigues:
I can reproduce this on OpenLogi v0.8.3 with an MX Master 3 on macOS 26.6.2 arm64, connected directly over Bluetooth.

The failure is transition-triggered: the gesture works normally in other applications, but when switching from another application into Microsoft Teams, the GestureButton click and swipes can fall through to the macOS-native App Switcher / Alt-Tab behavior. It is not permanently stuck in that state.

My relevant configuration is:

- Global `Back = BrowserBack`
- `com.microsoft.teams2` per-app override for `Back`
- Global GestureButton directions:
  - Click → `MissionControl`
  - Up → `MissionControl`
  - Down → `ShowDesktop`
  - Left → `PrevTab`
  - Right → `NextTab`

I tested the per-app `Back` override with three variants:

1. `Back = BrowserBack` in the Teams profile, matching the global binding: the gesture remained functional across switches into Teams.
2. `Back = None` in the Teams profile: the gesture failed again when entering Teams.
3. `Back = { CustomShortcut = "Cmd+Shift+M" }`: the original failure also reproduces.

This suggests that the trigger is not the custom shortcut itself, nor the Teams bundle identifier. The failure occurs when the effective per-app binding differs from the global/native `Back` binding and the capture plan has to change. During that transition, the GestureButton appears to be temporarily restored to its firmware-native behavior.

Diagnostics:

- OpenLogi GUI: v0.8.3 (release)
- Agent: v0.8.3 (connected)
- IPC protocol: GUI 29 / agent 29
- Device: MX Master 3 (`b023/4082/0000`, ext `01`)
- Connection: Bluetooth direct
- Accessibility: granted
- Input hook: installed
- Config schema: 7

This seems to confirm the capture-session stop/start gap described by this PR. The `BrowserBack` result is also a useful control case: keeping the effective capture plan unchanged avoids the failure, while both `None` and `CustomShortcut` trigger it.

Issue #660, opened 2026-08-18
G502 LIGHTSPEED reports HiResWheel (0x2121) over the receiver but not over USB, so wheel capabilities disappear when cabled


yuzi-co:
### What happened?

On a G502 LIGHTSPEED, HID++ `0x2121 HiResWheel` is reported over the Lightspeed receiver but **not** when the same mouse is connected by USB cable. Everything else about the feature table is stable, so the wheel capabilities silently disappear on the wired transport.

`Capabilities::from_feature_ids` derives `hires_wheel: ids.contains(&0x2121)` (`crates/openlogi-core/src/device.rs`), and `probe_extra_capabilities` only sets `scroll_inversion` when `HiResWheelFeature` resolves (`crates/openlogi-hid/src/inventory/features.rs:295`). With `0x2121` gone, **both** flags fall to `false` — so a user who plugs in the cable should see the hi-res-wheel and scroll-inversion controls vanish for a mouse that had them a moment earlier.

I am reporting the probe result rather than a diagnosis; I do not know whether this is device firmware, the direct-route probe, or something in between.

### Steps to reproduce

1. Pair a G502 LIGHTSPEED to its Lightspeed receiver (`046d:c539`), receiver plugged in, no cable.
2. `openlogi diag features` → note `0x2121` is present.
3. Unplug the receiver and connect the mouse with a data-capable USB cable, so it enumerates as `046d:c08d`.
4. `openlogi diag features` → `0x2121` is gone.
5. `openlogi diag wheel` → fails.

Reproduced on three consecutive probes in each state; the feature count is 30 either way, so this is not a truncated read.

### Diagnostics

Over the receiver (`046d:c539`, slot 1):

```
> openlogi diag features
device: G502 (slot 1 on receiver 82839805)
  ... 30 entries, including:
    29  0x2121  v0
```

Over the cable (`046d:c08d`, slot 255):

```
> openlogi diag features
device: G502 LIGHTSPEED Wireless Gaming Mouse (direct 046d:c08d)
  ... 30 entries, 0x2121 absent

> openlogi diag wheel
device: G502 LIGHTSPEED Wireless Gaming Mouse (direct 046d:c08d)
Error: read HiResWheel mode

Caused by:
    device does not expose HID++ feature 0x2121
```

Set difference between the two tables, names resolved against `crates/openlogi-hidpp/src/feature/registry.rs`:

```
only over USB:  + 0x00c2 v0  DfuControlSigned
[...]

PR #690, opened 2026-08-19
feat(cli): report feature flags and firmware entities in diag features

yuzi-co:
## Summary

`openlogi diag features` printed a feature's ID and version and threw away
everything else the device said. Two things were missing, and both are what a
device report needs to be actionable.

**Feature type flags.** The feature table carries obsolete / hidden /
engineering flags that `FeatureEntry` was discarding. On a G502 LIGHTSPEED, 16
of its 30 features are flagged hidden or engineering, and the old dump gave no
way to tell those from the ones configuration software is meant to drive.

**Firmware entities.** A device lists its firmware through `0x0003` function 1:
the running application, the bootloader, and on many models a separate radio
stack. That version is what explains why two people with the same model see
different behaviour, and it was not reachable from the CLI at all.

Read-only throughout. Nothing here writes to a device.

## Before / after

Same device, same command, wired G502 LIGHTSPEED:

```
 before                          after
    12  0x2201  v1                  12  0x2201  v1
    13  0x00c2  v0                  13  0x00c2  v0
    14  0x1802  v0                  14  0x1802  v0    hidden,engineering
    ...                             ...
    26  0x1e00  v0                  26  0x1e00  v0    hidden
  (30 feature entries)            (30 feature entries)

                                  fw 0: Bootloader BOT92.00_B0008 pid=aaef
                                  fw 1: MainApplication MPM17.00_B0008 pid=c08d [active]
                                  fw 2: unreadable (HID++ unsupported
                                        response during DumpFeatures
                                        for feature 0x0003)
```

## Unreadable entities are reported, not dropped

The G502 declares three entities and its third fails to decode. Dropping the
row would claim the device has two firmware images when it says it has three,
so the entry carries the reason instead.

That third entity is the radio stack, and the decode failure is a real gap in
the vendored `hidpp` parser rather than a device quirk. Raw payload:

```
entity 2 = [05, 00, 00, 00, 00, 00, 00, a9, 00, ...]
            ^^ entity_type 5 = Softdevice
                                    ^^ build = 0x00a9, not valid BCD
```

`get_fw_info` runs the build field through `bcd::convert_packed_u8`, which
rejects `a9`. Filed separately rather than fixed here, since it belongs in the
vendored crate and has its own blast radius.

## Changes

**`crates/openlogi-hid`**

- `FeatureEntry` gains `typ`, the flags the device already sends alongside each
  feature; `dump_features` stops discarding them
- new `FirmwareEntity` and `dump_firmware_entities`, reading every entity the
  device declares. `FirmwareEntity` is a sum type — `Readable { index, info }`
  or `Unreadable { index, error }` — so a version with no kind, or an error
  beside a version, is unrepresentable rather than merely unreachable
- `FirmwareEntityInfo` carries the whole structured record (kind, prefix,
  number, revision, build, active, transport_pid, extra_version); rendering is
  the CLI's job. It mirrors `hidpp`'s `DeviceEntityFirmwareInfo`, which is
  `#[non_exhaustive]` and so cannot be constructed by consumers or their tests
- a failed entity read stays a row only when the *device* answered about that
  entity — `Feature` and `UnsupportedResponse`. Channel failures propagate
  through `WriteError` rather than becoming several misleading
  malformed-firmware rows, one timeout apart
- re-export `FeatureType` and `DeviceEntityType`. `FeatureEntry::typ` is a
  `hidpp` type in the public API and a consumer previously had no way to name
  it, which left the CLI testing raw bits
- `channel::scripted` gains `with_failing_writes` for transport-level failures,
  and takes over the `feature_error` test helper that was copied per module

**`crates/openlogi-cli`**

- `diag features` prints a `flags` column and the firmware entity lines
- `format_feature_flags` names all five known `FeatureType` flags and appends
  whatever `from_bits_retain` kept as a raw `unknown=0x..` mask, so the column
  is the whole response rather than part of it
- `format_firmware_entity` is a pure formatter with unit tests

## Testing

- `cargo clippy --workspace --all-targets -- -D warnings` (with
  `RUSTFLAGS=-D warnings`) — clean
- `cargo test --workspace` — clean, including 15 new tests: 10 for the two CLI
  formatters, and 2 scripted-device tests pinning "an entity the firmware
  cannot describe continues the dump, a channel failure aborts it"
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
  --document-private-items` (GPUI crates excluded, as the gate does) — clean
  except `openlogi-permissions`, whose crate-doc link to `PermissionStatus`
  does not resolve on Windows because the type is cfg-gated away there. That is
  on master (0f028a8) and untouched here; CI's Linux run passes
- `cargo test -p openlogi-ipc --test wire_format` — clean. Nothing on the wire
  changed: `FirmwareEntity` is CLI-only, and `WriteError` is used, not modified
- Hardware: verified on Windows 11 against a G502 LIGHTSPEED, both on its
  Lightspeed receiver and cabled. The feature table and firmware records above
  are real. That run predates this review follow-up, so the exact
  `unreadable (...)` wording in the block above is the current formatter's
  output rather than a fresh capture

Incidentally confirms #660 from the other direction: the same mouse gains
`0x00c2` and loses `0x2121` when you plug the cable in, 30 features either way.


yuzi-co:
Rebased onto master (`55a6ff2`) — CI is green, including `tests (linux)`, which was failing before.

The three failures were toolchain and reorg drift, not this change. The branch sat 65 commits behind, so it predated both master's Rust 1.98 lint fixes (4975729) and the `openlogi-hid` module reorg (9126146).

## One conflict, in `crates/openlogi-hid/src/lib.rs`

Both hunks were in the re-export block, from master moving things under `channel/` and `inventory/`. This commit only ever added four names there:

- `FeatureType`
- `DeviceEntityType`
- `FirmwareEntityEntry`
- `dump_firmware_entities`

Everything else in the conflict was reflow around those. I took master's list as the base and re-added exactly those four, so master's `Dpi`, `LITRA_BEAM_PRODUCT_ID`, `LITRA_GLOW_PRODUCT_ID`, and the `ScrollReportingTarget` / `ScrollResolution` / `ScrollWheelMode` exports are all intact — the diff against master is now those four additions plus rustfmt reflow, nothing else. Worth a glance since I resolved it rather than you.

The other three files (`diag/features.rs`, `write.rs`, `write/diagnostics.rs`) merged without conflict.

## Gate

Run on the rebased tree at rustc 1.98.0:

```
cargo fmt --all -- --check                              exit=0
cargo clippy --workspace --all-targets -- -D warnings   exit=0
cargo test --workspace                                  exit=0
```

crates/openlogi-hid/src/write/diagnostics.rs
```diff
@@ -174,3 +183,82 @@ pub async fn read_battery_raw(route: &DeviceRoute) -> Result<String, WriteError>
     })
     .await
 }
+
+/// Snapshot of one firmware entity a device reports through HID++ `0x0003`
+/// function 1. Returned by [`dump_firmware_entities`] so a device report can
+/// name the exact firmware it is running.
+#[derive(Debug, Clone)]
+pub struct FirmwareEntityEntry {
```
AprilNEA: Could this be modeled as a sum type instead of parallel `Option` fields? The implementation only produces two valid states—fully decoded or unreadable—but this public struct also permits contradictory states such as a version plus an error, or `version = None` with no error. The CLI consequently has to handle an `"unparsed"` state that the collector never creates.

Since this is a new API, please encode that invariant before it becomes established, for example with `Readable { index, info }` and `Unreadable { index, error }` variants. The readable form should also preserve the structured firmware fields, including `extra_version`, rather than converting part of the record to a display string and dropping the rest. Presentation formatting can remain in `openlogi-cli`.

crates/openlogi-hid/src/write/diagnostics.rs
```diff
@@ -174,3 +183,82 @@ pub async fn read_battery_raw(route: &DeviceRoute) -> Result<String, WriteError>
...
+                    kind: Some(fw.entity_type),
+                    version: Some(format!(
+                        "{}{:02}.{:02}_B{:04}",
+                        fw.firmware_prefix, fw.firmware_number, fw.revision, fw.build
+                    )),
+                    transport_pid: Some(fw.transport_pid),
+                    active: fw.active,
+                    error: None,
+                },
+                Err(e) => FirmwareEntityEntry {
```
AprilNEA: This currently turns every `Hidpp20Error`, including channel failures and timeouts, into an entity-local `"unreadable"` result and then continues querying subsequent entities.

If the device disconnects while reading entity 1, the command can wait for every remaining request to time out and report several apparently malformed firmware records, when the actual failure is that the route disappeared. Please retain only entity-local response/feature failures as unreadable rows and propagate channel/transport failures through the outer `WriteError`.

A scripted test covering “unsupported entity continues, channel failure aborts” would make this distinction durable.

crates/openlogi-cli/src/cmd/diag/features.rs
```diff
@@ -29,18 +29,141 @@ pub async fn run(_args: FeaturesArgs) -> Result<()> {
...
                     for (idx, entry) in entries.iter().enumerate() {
-                        println!("  {:>4}  0x{:04x}  v{}", idx, entry.id, entry.version);
+                        let mut flags = Vec::new();
+                        if entry.typ.contains(FeatureType::OBSOLETE) {
+                            flags.push("obsolete");
+                        }
+                        if entry.typ.contains(FeatureType::HIDDEN) {
+                            flags.push("hidden");
+                        }
+                        if entry.typ.contains(FeatureType::ENGINEERING) {
```
AprilNEA: `FeatureType` currently defines two additional known flags, `MANUFACTURING_DEACTIVATABLE` and `COMPLIANCE_DEACTIVATABLE`, but this formatter silently drops them. Because the HID++ parser uses `from_bits_retain`, unknown flag bits are also available and are particularly useful in diagnostic output.

Please extract this into a small tested formatter that emits all five known flags and includes any unknown remainder as a raw mask such as `unknown=0x04`. Otherwise the new `flags` column still discards part of the device response it is intended to expose.

AprilNEA (changes requested):
Thanks for adding this—feature-type metadata and firmware versions are both valuable additions to an actionable device report. Keeping declared-but-unparseable entities visible is also the right goal, and the real G502 validation gives useful confidence in the protocol flow.

I found three issues in the newly introduced API/behavior that I think should be addressed before merging:

1. `FirmwareEntityEntry` represents one two-state result with several independent `Option` fields, permitting contradictory and unreachable states. A readable/unreadable enum would encode the invariant and keep the CLI formatter simpler.
2. The per-entity error path currently absorbs channel and timeout failures. Those are operation-level failures rather than malformed entity records and should abort the dump instead of causing repeated timeouts and misleading rows.
3. The CLI prints only three of the five known `FeatureType` flags and drops retained unknown bits, leaving the diagnostic output incomplete.

While adjusting the firmware result type, please preserve the structured firmware fields—including `extra_version`—and leave their presentation formatting at the CLI boundary.

These changes do not require restructuring the existing HID API, adding a new IPC-visible `HidppOperation`, or fixing the separate SoftDevice BCD parser issue in this PR. The existing re-exports and the decision to handle that parser gap separately both look reasonable.

With the three inline issues addressed and focused tests for flag formatting and entity-error classification, this should be in good shape.

crates/openlogi-hid/src/write/diagnostics.rs
yuzi-co: Done in 188d2ea. `FirmwareEntity` is now `Readable { index, info }` / `Unreadable { index, error }`, so a version with no kind and an error beside a version are both unrepresentable, and the CLI no longer carries an `unparsed` branch.

The readable variant holds the structured record — kind, prefix, number, revision, build, active, transport_pid, extra_version — and the CLI builds `MPM17.00_B0008` from it, printing `extra=…` when a device populates those bytes. `Unreadable` carries a typed `WriteError` rather than a string, so the reason stays data.

One judgement call worth flagging: `FirmwareEntityInfo` mirrors `hidpp`'s `DeviceEntityFirmwareInfo` instead of re-exporting it. That type is `#[non_exhaustive]`, so nothing outside the protocol crate can construct one and the CLI formatter tests could not build a fixture. Mirroring follows the conversion `ReprogControlEntry` already does for `CidInfo`, and keeps a foreign non-exhaustive type out of the public API. Say the word if you would rather drop `#[non_exhaustive]` in the fork and use the protocol type directly.

crates/openlogi-hid/src/write/diagnostics.rs
yuzi-co: Done in 188d2ea. Only `Hidpp20Error::Feature(_)` and `UnsupportedResponse` — the device answering about that entity — become `Unreadable` rows now. Everything else, `Channel(_)` included, is classified through `classify_hidpp_error` and returned, aborting the dump. `Hidpp20Error` is `#[non_exhaustive]`, so the wildcard arm aborts as well: a variant this build does not know is not evidence about the entity.

Two scripted tests in `write/tests.rs` pin the split, against a device modelled on the G502 (bootloader, running main application, a Softdevice whose build field is not valid BCD) plus one invented entity whose firmware refuses the read outright:

- `an_entity_the_device_cannot_describe_does_not_stop_the_dump` — both entity-local shapes stay rows, the declared count still matches, and every entity is still asked about.
- `a_channel_failure_aborts_the_dump_instead_of_blaming_the_firmware` — the write for entity 1 errors at the transport; the call fails and entity 2 is never requested.

The scripted channel gained `with_failing_writes`, which errors raw writes matching a predicate — a node disappearing mid-conversation, and it keeps the test off the 5 s timeout path. I confirmed the abort test fails against the old catch-all arm before keeping it.

crates/openlogi-cli/src/cmd/diag/features.rs
yuzi-co: Done in 188d2ea. `format_feature_flags` is a tested formatter that names all five known flags — `obsolete`, `hidden`, `engineering`, `manufacturing-deactivatable`, `compliance-deactivatable` — and appends the retained remainder as a raw mask, e.g. `hidden,unknown=0x04`.

Five unit tests cover it: empty, all five set, the two version-2 flags alone (the ones that used to vanish), a known flag plus an unknown bit, and unknown bits with no known flag.

AprilNEA (approved):

AprilNEA:
LGTM.

PR #701, opened 2026-08-20
fix(gui): replace tokio interval with gpui timer to prevent panic

clifinger:
## Summary
The recent commit `a2613f5684f4` added a `tokio::time::interval` to the GUI's background loop, but `openlogi-desktop` runs its async block on GPUI's executor (`smol`), not a Tokio reactor. This causes an immediate panic on startup: `there is no reactor running, must be called from the context of a Tokio 1.x runtime`.

This PR replaces the Tokio interval with GPUI's native `cx.background_executor().timer()`, which correctly yields on the GPUI executor.

## Changes
- `crates/openlogi-desktop/src/main.rs`: swapped `tokio::time::interval` for `gpui::Timer` to avoid Tokio dependency in the UI loop.

## Testing
- Verified on Ubuntu 26.04. The app now launches successfully and connects to the agent without panicking.

Fixes # (no issue, observed locally)

### DIFF
--- a/crates/openlogi-desktop/src/main.rs
+++ b/crates/openlogi-desktop/src/main.rs
@@ -465,8 +465,8 @@ fn main() -> Result<()> {
             // the merge without waiting for the agent to change something of
             // its own.
             let mut latest_snapshot: Option<openlogi_ipc::AgentSnapshot> = None;
-            let mut camera_scan = tokio::time::interval(CAMERA_SCAN_PERIOD);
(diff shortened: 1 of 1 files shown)


clifinger:
Closing this PR because the exact same fix was already merged in #686, and the event loop has since been refactored into `runtime.rs` in master.

AprilNEA:
> Closing this PR because the exact same fix was already merged in #686, and the event loop has since been refactored into `runtime.rs` in master.

Thanks for the effort here!

Issue #426, opened 2026-07-19
[Bug]: Previous and next desktop action on gesture button don't work


vivdesign:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.20

### Operating system

Linux

### OS version & architecture

fedora 44 kde plasma. amd64

### Device model

MX Master 3S + Bold receiver 2

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [ ] Device discovery / detection
- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

Up and down gesture button actions don't do anything when set to next or previous desktop

### Steps to reproduce

1. Set the next and previous desktop actions on the up and down gesture button actions
2. Press the button and move the mouse up or down

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.6.20 (release)
- Agent: v0.6.20 (connected)
- IPC protocol: GUI 9 / agent 9
- Inventory: ready
- OS: Linux (x86_64)
- Locale: en-US (UI: follow system)
- Accessibility: granted · Input hook: not installed
- Launch at login: no · Menu bar: yes · Update check: on
- Running from: source build (dev)
- Config: schema 3 · 2 configured device(s) · thumbwheel 14

**Assets**
- Source: user cache · Index: loaded (210 models) · User cache: present
- Cache path: ~/.local/share/openlogi/assets · Bundle assets: absent

**Devices (2)**
- MX Master 3S — mouse (codename: MX Master 3S)
  - Connection: Logi Bolt receiver · Online: yes · Battery: 50% (discharging, good)
  - Capabilities: buttons=yes, pointer=yes, lighting=no
  - DPI: 1000 dpi (range 200–8000, 157 steps)
[...]

Issue #349, opened 2026-07-03
[Bug]: MOVE SPACES WITH THUMB PAD


Tintin0:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.18.

### Operating system

macOS

### OS version & architecture

Tahoe, 26.4.1

### Device model

MX MASTER 3S - Bluetooth 

### How is the device connected?

Bluetooth (direct, no receiver)

### Affected area(s)

- [ ] Device discovery / detection
- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

I'm used to clicking through the Spaces of macOS from left to right or from right to left with the mouse's thumbpad clicked and then right or left movement with the mouse. Theoretically, there is the option also in OpenLogi but it does not work reliably.

### Steps to reproduce

1. Mapped the previous and next desktop to the thumbpad + move movement left or right respectively.

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

### macOS permissions (if applicable)

- [x] OpenLogi has **Accessibility** permission (needed to remap buttons via the event tap).
- [x] OpenLogi has **Input Monitoring** permission (needed for Bluetooth-direct devices and capture).

PR #719, opened 2026-08-20
fix(hid): discard raw-XY jump artifacts from the gesture button/panel

obvionaoe:
## Summary

Gesture button/panel raw-XY reports carry an absolute-position jump instead of a
delta. On a real MX Master 4 every gesture-button hold opens with one — a first
report hundreds to thousands of units off, scaling with the pointer resolution
(measured up to ~1700 units at 400 DPI, ~6700 at 8000). Summed into the swipe
accumulator it single-handedly crosses the commit threshold, so a
left/right/up/down swipe fires the wrong direction or none.

The previous discard covered only the first report, and only for
`HAPTIC_PANEL_CID`. This replaces it with a per-hold `JumpFilter`.

## Changes

- `crates/openlogi-device/src/session/gesture.rs`
  - New per-hold `JumpFilter`, replacing the `skip_first_raw_xy` flag on
    `HoldState::Holding`. Two rules:
    - The first report of a hold on a touch-sensor gesture source (haptic panel
      or mechanical gesture button) is dropped outright — on hardware every such
      hold has exactly one contact artifact, always first. A standard button
      repurposed as a raw-XY gesture source is relative from report one and is
      not pre-emptively dropped.
    - A later report is dropped when its dominant axis exceeds 5× the largest
      report admitted so far this hold, floored at 12× a swipe's commit travel
      for the opening reports. Raw-XY travel scales with pointer DPI (≤ ~35
      units/report at 400 DPI, up to ~550 for a fast swipe at 8000), so a fixed
      bound can't fit both ends — keying it on the hold's own peak makes it
      DPI-proportional without reading the DPI. Peak, not mean: the contact jump
      is trailed by ~1-unit settling reports that pin a mean near zero and then
      starve the real swipe (seen on hardware). A run of four reports all
      tripping the bound disengages the filter for the rest of that hold, so a
      wrong bound costs a few reports of latency, never a gesture.
  - The filter is ticked on every raw-XY report of a hold, including ones
    dropped for two-source overlap.
  - New `trace!` on every raw-XY report before filtering, so the sample sequence
    behind a jump or a gesture's feel can be captured per device and DPI
    (`OPENLOGI_LOG=...openlogi_device::session::gesture=trace`).
- `crates/openlogi-device/src/session/gesture/tests.rs`
  - `JumpFilter` unit tests: first-report rules for both source kinds, a stray
    mid-hold jump, the bound growing with a fast swipe, tiny settling reports
    not poisoning it, the disengage path.
  - Integration regressions: a jump several reports into a hold discarded with
    motion resuming to the real direction; the dedicated gesture button's
    contact jump discarded; a hold of uniformly large reports still resolving.
  - Updated existing tests that fed a single post-press raw-XY report to account
    for the contact-report discard.

## Testing

In a devenv shell:

- `cargo test -p openlogi-device` — 211 passed
- `cargo test -p openlogi-hid -p openlogi-agent-core -p openlogi-agent` — passed
- `cargo clippy -p openlogi-device -p openlogi-hid -p openlogi-agent-core -p openlogi-agent --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc -p openlogi-device --no-deps --document-private-items` — clean

Runtime-verified on a real MX Master 4 (Bluetooth) at **400, 3200 and 8000 DPI**,
~60 mixed gestures per DPI (left/right/up/down/tap), with an instrumented build
logging every raw-XY report:

- Contact jump present on every hold, discarded; every swipe committed the
  direction made; every tap resolved as a click.
- Raw-XY per-report travel confirmed to scale with DPI (≤ ~35 at 400, ≤ ~170 at
  3200, up to ~550 on fast swipes at 8000); contact-jump magnitudes ~1700 to
  ~6700 across that range.
- No real motion filtered, including fast swipes at 8000 DPI whose per-report
  travel exceeds a fixed unit bound — the peak-based bound grows with the swipe.

DPI trace samples (fast Right swipe, 8000 DPI, per-report `dx`):
`... 255, 285, 330, 346, 357, 369, 459, 433 → committed Right` — the 459/433
reports are admitted because the peak has already reached ~369.

Likely related to #349, #426 and #752.


AprilNEA:
Thanks for taking the time to investigate this on real MX Master 4 hardware and capture the raw samples. The observation that these jumps can occur in the middle of a hold is especially useful, and I agree that filtering them before they reach `SwipeAccumulator` is the right layer. The existing first-sample-only handling clearly does not cover the case you observed.

I have one concern that I think we should address before merging: `250` is a fixed raw-count cutoff applied to every sample from every gesture source and device. The relationship between raw counts and physical movement depends on the configured DPI, so a threshold validated with one device setting may not generalize to higher DPI settings or other supported devices.

For example, at 8000 DPI and roughly 125 Hz, 251 counts represents only about 0.8 mm of movement in one report, or around 10 cm/s. That seems reachable during a legitimate fast swipe. The filter could therefore undercount valid movement and, in the worst case, let the release resolve as a click.

Could we make the bound DPI-aware, or otherwise avoid applying a device-independent raw-count cutoff globally? I would also prefer to preserve the existing first-sample safeguard unless the replacement can be shown to cover it reliably.

A few focused regression cases would help us validate the final approach:

- a jump occurring several samples into a hold;
- normal motion continuing after the jump;
- legitimate fast movement at the device’s highest DPI;
- boundary and release behavior;
- both Haptic Panel and mechanical gesture-button holds, where practical.

It would also be helpful to note the DPI used for the current 60+ gesture test and, if possible, include a few sample sequences captured at both low and high DPI.

Thanks again for the hardware investigation — this looks like a real device quirk, and the traces you collected give us a very useful starting point.

cabaucom376:
Apologies if this isn't the place to bring it up. But I've noticed when it comes to desktop switching with the left/right gesture using LogiOptions+ I'm used to being able to hold the button down and jump multiple desktops with one swipe and a small vibration each jump.

The vibration I'm not so concerned about, but having to stop the gesture and do it again is meaningfully slowing me down keeping me on LogiOptions+ for now.

Im using an MX master 4

obvionaoe:
@cabaucom376 I have also realized that, but I would refrain from mixing it with this PR and apparently #969 already is working on it

obvionaoe:
@AprilNEA will test this as soon as I have the time and address those points you made

obvionaoe:
@AprilNEA Reworked this and tested it on the Master 4 at 400, 3200 and 8000 DPI.

You were right that raw-XY scales with the pointer resolution — same physical
swipe measured ≤ ~35 units/report at 400 DPI and up to ~550 on a fast swipe at
8000, with contact-jump magnitudes running ~1700 to ~6700 across that range. A
fixed `250` (or any fixed unit bound) can't fit both ends.

The fixed 250 discard is gone. It's now a per-hold `JumpFilter`:

- The first raw-XY report of a hold is discarded for the touch-sensor sources
  (haptic panel and, per the traces, the mechanical gesture button). Every hold
  opens with exactly one contact artifact and it's always first, so this keeps
  the existing first-sample safeguard, just no longer scoped to
  `HAPTIC_PANEL_CID`. A standard button repurposed as a raw-XY gesture source is
  relative from report one and isn't pre-emptively dropped.

- After that, a report is dropped only when its dominant axis exceeds 5x the
  largest report admitted so far this hold (floored at 12x GESTURE_SWIPE_THRESHOLD
  for the opening reports). Keying it on the hold's own peak makes it
  DPI-proportional without plumbing the DPI value — the peak reflects whatever
  resolution the hold is running at. It's a running peak rather than a mean
  because a mean gets poisoned: the contact jump is trailed by a handful of
  ~1-unit settling reports, and a mean keyed on those starved real swipes into
  clicks on hardware. If four reports in a row trip the bound it's wrong for the
  device, so the filter disengages for the rest of that hold — worst case a few
  reports of latency, never a lost gesture.

Verified live: ~60 mixed gestures at each of 400 / 3200 / 8000 DPI. Every swipe
committed the direction made, every tap resolved as a click, and no real motion
was filtered — including fast swipes at 8000 DPI whose per-report travel exceeds
a fixed unit bound (the peak-based bound has grown past ~350 by then). Regression
tests cover a jump several reports into a hold, motion resuming after it, the
settling-report poison case, and a hold of uniformly large reports.

Also added a trace-level log of every raw-XY report before filtering, so the
sample sequence is easy to capture per device and DPI. DPI used for the runs:
400, 3200 and 8000.

davidbudnick:
Fixes #752

obvionaoe:
@AprilNEA Anything missing that would block this from merging? I have the local patch running in my own agent but I'd prefer to fully upstream this and reduce the amount of local patches I rely on to a minimum :)

PR #720, opened 2026-08-20
chore(infra): prune the tree's lint suppressions and make the survivors expect

AprilNEA:
## Summary

An audit of every lint suppression in the tree — 207 attributes carrying 247 lints — found that most of them were not carrying information. 78 were test modules restating the same `unwrap`/`expect` exemption by hand, 20 no longer suppressed anything at all, and seven files silenced `cast_*` for their entire contents. This prunes the tree to 130 attributes / 140 lints and changes the default form from `allow` (silent when it goes stale) to `expect` (fails the build when it does): the `allow`:`expect` ratio goes from 159:48 to 6:124.

The mechanism is worth keeping: rewriting every non-`cfg_attr` `allow(` to `expect(` and running clippy turns each dead suppression into an "unfulfilled lint expectation" warning. That is how the 20 below were found, and `.claude/rules/rust.md` now documents the recipe so the sweep can be repeated.

Notable finds:

- `openlogi-assets`'s three module-wide `#![allow(dead_code)]` blankets have been inert since those modules became `pub mod` — public items are reachable, so the lint could never fire. They would have silently hidden real dead code the moment a module went private again.
- `LightCapabilities` is down to three bools, under `struct_excessive_bools`. `InstanceGuard._handle` and `CGEventTapInformation` are exempt from `dead_code` on their own.
- `_silence_pickfn` was a dead function whose only job was keeping an unused `PickFn` import "used" — a hand-rolled way around `unused_imports`. Both are gone.

Three cases where `expect` provably cannot work keep `allow`, each with a comment naming which case applies:

- **cfg-conditional**: `platform::os_version` returns `Some(…)` on macOS (so `unnecessary_wraps` fires) and `None` elsewhere (so it does not) — an `expect` there is green on macOS and red on the other two lanes.
- **lint raised inside a macro expansion**: rustc does not credit an expectation with such a lint, so `expect` suppresses the warning *and* reports itself unfulfilled — unfixable under `-D warnings`. `function_row`'s `float_cmp` on floats compared inside `assert_eq!` is the case here.
- **fulfilment differing between a crate's targets**: a `dead_code` suppression on a helper only the tests call is fulfilled in the `--lib` build and unfulfilled in the `--test` build. These were already `cfg_attr`-wrapped.

## Changes

- **root**: new `clippy.toml` with `allow-unwrap-in-tests` / `allow-expect-in-tests`. Clippy's exemption covers `#[cfg(test)]` modules and `#[test]` functions; a free helper in a `tests/` integration file is the one shape it cannot see, so `openlogi-ipc`'s wire-format test keeps a file-level suppression, as do the three build scripts (a build script is not a test — it fails by panicking).
- **openlogi-camera**: `capture_windows`'s file-wide cast blanket is gone entirely — its five sites are `cast_signed()` / `cast_unsigned()`, which say "reinterpret the bits" rather than "trust me". `uvc` builds its descriptor fixture with `to_le_bytes` and keeps two narrowly scoped suppressions; `capture` keeps one on the statement that stores frame dimensions. `capture_linux` keeps its blanket: `v4l2-sys-mit` runs bindgen, so that backend cannot be cross-linted from macOS and narrowing it would be blind work.
- **openlogi-hook**: the Win32 backend's four-lint file blanket is replaced by `&raw mut` / `&raw const` at its six FFI pointer coercions (the edition-2024 spelling, and the better one at an FFI boundary since it never materializes the intermediate reference), `cast_signed()` at three sites, and four function-scoped `expect`s for what is genuinely left.
- **openlogi-agent**: `tray_windows`'s blanket narrowed to the five functions that need it; the three build scripts' `expect_used` restated as `expect`.
- **openlogi-desktop**: `camera/controls` funnelled eleven ad-hoc `i32 as f32` / `f32 as i32` conversions through a `to_slider`/`from_slider` pair, so the UVC-range argument is made once where it belongs instead of eleven times by omission. `function_row`'s cast blanket narrowed to four functions. Nine dead `dead_code`/cast suppressions removed.
- **openlogi-core, openlogi-hidpp, openlogi-hid, openlogi-assets, openlogi-cli, openlogi-inject, openlogi-ipc, xtask**: dead suppressions removed, survivors converted to `expect`.
- **docs**: `.claude/rules/rust.md` no longer tells agents to hand-write a test-module `#[allow]`, no longer describes `openlogi-hidpp` as sitting outside the workspace lint table (its own AGENTS.md retired that), and gains the `expect`-by-default rule plus the scoping rule for cast suppressions. `docs/DECISIONS.md` records the decision with the numbers behind it.

## Testing

Full local gate on the final tree:

```
cargo fmt --all -- --check
cargo test --workspace
```

Because a platform-gated suppression is only ever evaluated on its own platform, and because an `expect` that goes unfulfilled is a hard error under `-D warnings`, the sweep was verified on all three lint lanes rather than just the local one:

```
cargo clippy --target x86_64-pc-windows-gnu -p openlogi-core -p openlogi-hidpp -p openlogi-hid \
  -p openlogi-hook -p openlogi-agent -p openlogi-agent-core -p openlogi-inject -p openlogi-camera --all-targets
cargo clippy --target aarch64-unknown-linux-musl -p openlogi-core -p openlogi-hidpp -p openlogi-hid \
  -p openlogi-ipc -p openlogi-permissions -p openlogi-hook -p openlogi-inject -p openlogi-agent-core -p openlogi-agent --all-targets
```

Both clean. Not covered locally: the GPUI crates off macOS (no cross build) and `openlogi-camera`'s v4l2 backend (bindgen) — CI's `clippy` and `clippy (windows)` jobs own those.

This is a lint-hygiene change with no behavioural surface, so there is nothing to verify on hardware. The three code rewrites that are not pure attribute edits — `cast_signed`/`cast_unsigned`, `&raw mut`/`&raw const`, and the `to_slider`/`from_slider` pair — are each exactly equivalent to what they replace, and all of them are on Windows or in the camera panel; **not runtime-tested on hardware**.

## Follow-ups

- `capture_linux`'s file-level cast blanket stays until the v4l2 backend can be linted somewhere other than a Linux box.
- CI has no macOS clippy job. Every `#[cfg(target_os = "macos")]` suppression — the ObjC FFI surface in `openlogi-inject`, `openlogi-hook`, `openlogi-camera`, `openlogi-permissions`, and both GPUI apps — is checked only by the maintainer's local gate. That gap predates this PR, but `expect` makes it more valuable to close.

### DIFF
--- a/.claude/rules/rust.md
+++ b/.claude/rules/rust.md
@@ -16,9 +16,12 @@ and every crate inherits it with `[lints] workspace = true` — never a private 
 the next lint added to the workspace silently skips that crate. A crate needing a
 different level opts out **in source** (the `openlogi-hook` platform modules carry
 `#![allow(unsafe_code, reason = "…")]`), because Cargo rejects mixing `workspace = true`
-with local overrides. `openlogi-hidpp` currently stays out of the table — it is a **hard
-fork**, so the "third-party code" rationale for that opt-out no longer holds; whether it
-should now inherit is an open question, costed in `crates/openlogi-hidpp/AGENTS.md`.
+with local overrides. `openlogi-hidpp` inherits the table like everything else, hard
+fork or not.
+
+Workspace-wide clippy *configuration* lives in `clippy.toml` at the root. It currently
+carries `allow-unwrap-in-tests` / `allow-expect-in-tests`, so test code never needs to
+restate that exemption by hand.
 
 The table: `unsafe_code = "deny"` (opt out per item with `#[expect(unsafe_code,
 reason = "…")]` plus a `// SAFETY:` comment), `clippy::pedantic` at warn,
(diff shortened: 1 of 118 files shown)


davidbudnick:
@greptile-apps

davidbudnick:
@greptile-apps

PR #760, opened 2026-08-21
fix(ipc,agent,desktop): surface Input Monitoring via agent status and fix registry bincode

### LINKED ISSUE #571: Snapshot fails to decode when a standalone device has no registry_model_id
## Summary

`StandaloneDevice::registry_model_id` carries
`#[serde(default, skip_serializing_if = "Option::is_none")]`
(`crates/openlogi-core/src/device.rs`). bincode is not self-describing, so a
skipped field is simply *absent* from the byte stream — there is no field name
or tag for the decoder to notice it is missing. When `registry_model_id` is
`None`, the encoded `AgentSnapshot` is therefore short, the GUI's decode walks
off the end, and the entire snapshot fails with

```
Channel(Read(Custom { kind: Other, error: Custom { kind: InvalidData, error: Io(Kind(UnexpectedEof)) } }))
```

Because `snapshot` is what the GUI polls, this is not a partial-data problem:
while such a device is connected, **every poll fails**, so the GUI gets no
inventory at all — not just a light without a picture.

`skip_serializing_if` is safe for self-describing formats (JSON), which is
presumably where the "deliberately appended" reasoning came from, but it cannot
express an optional field in bincode. `Option<T>` already encodes its own
presence byte, so the attribute is also unnecessary — a plain `Option` field is
what makes the append safe.

## Reproduction

Any standalone device whose registry model id does not resolve, i.e. the
`registry_model_id: None` construction path in
`crates/openlogi-hid/src/standalone.rs`. Observed with a scripted `Litra Glow`
served over the real IPC socket: with `Some("8c900")` the GUI polls normally;
flipping only that field to `None` makes every `snapshot()` fail to decode.

## Why the goldens did not catch it

`assert_wire` in `crates/openlogi-agent-core/tests/wire_format.rs` is
serialize-only (`fn assert_wire<T: serde::Serialize>`) — it compares the hex of
an encode and never decodes. `standalone_light_dtos_commands_and_errors` even
pins the `registry_model_id = None` encoding as a golden, so the truncated
stream is currently enshrined as correct.

## Suggested fix

1. Drop `skip_serializing_if` from `registry_model_id` (keep `#[serde(default)]`
   for the TOML/config side if it is read there).
2. Bump `PROTOCOL_VERSION` and regenerate the two affected goldens.
3. Make `assert_wire` round-trip: decode the golden bytes back and compare with
   the input value (`T: Serialize + DeserializeOwned + PartialEq + Debug`). That
   is the assertion that would have failed here, and it costs nothing on every
   other type already covered.

Step 3 is the one worth doing regardless of this field — the goldens currently
prove that encoding is *stable*, not that it is *decodable*.

Found while building a hardware-free mock agent (#568), which has to emit these
DTOs by hand and hit the decode failure immediately.

### LINKED ISSUE #606: Settings' Input Monitoring status checks the GUI's own TCC grant, not the agent's
## Summary

`crates/openlogi-gui/src/windows/settings/permissions.rs` shows an Input Monitoring status row using `openlogi_gui::platform::permissions::input_monitoring()`, which calls `IOHIDCheckAccess` from **the GUI process**.

But the GUI never opens HID devices — the agent does, under its own separate code-signing identity (`org.openlogi.agent[.dev]`, distinct from `org.openlogi.openlogi[.dev]`; see `crates/openlogi-gui/src/platform/AGENTS.md`). TCC grants are scoped per-identity, so checking from the GUI process reports the GUI's own (irrelevant) grant state rather than whether the agent — the process that actually matters — can open HID devices. This can show "Granted" or "Denied" in Settings that has no bearing on whether device discovery actually works, or vice versa.

This mirrors how Accessibility status is already handled correctly: it's read from the agent via IPC (`AgentStatus::accessibility_granted`, kept live by an accessibility watcher), not checked locally in the GUI process (see the doc comment at the top of `permissions.rs`: "Accessibility status is owned by `AppState`... this module covers the other two [Input Monitoring, Bluetooth]").

## Proposed fix

Give Input Monitoring the same treatment as Accessibility:
- Add an `input_monitoring_granted` (or similar) field to `AgentStatus` in `crates/openlogi-agent-core/src/ipc.rs` (append-only, bump `PROTOCOL_VERSION`, regenerate `wire_format` goldens).
- Have the agent populate it from `openlogi_hid::permissions::has_access()` (added in #607).
- Have the Settings row read that IPC-surfaced status instead of calling `permissions::input_monitoring()` locally.

Optionally also add an on-demand `request_input_monitoring_prompt` RPC (mirroring `request_accessibility_prompt`) so the Settings row's click handler can re-fire the consent dialog, not just deep-link to System Settings.

## Context

Found while fixing #607: the agent never called `IOHIDRequestAccess`, so Input Monitoring was silently denied and never appeared in System Settings at all. That PR fixes the underlying discovery bug; this issue is about the Settings UI's status indicator being sourced from the wrong process, which is a separate, pre-existing inconsistency.

rekcilyssup:
## Summary

Fixes #606 and #571.

**#606:** Settings → Permissions → Input Monitoring checked `IOHIDCheckAccess` in the GUI process (`openlogi-permissions::input_monitoring()`), but HID devices are opened by the agent under a distinct code-signing identity (`org.openlogi.agent`). The row therefore showed the GUI's grant, not whether discovery can actually open HID. This mirrors the already-correct Accessibility handling.

**#571:** `StandaloneDevice::registry_model_id` used `skip_serializing_if`, truncating bincode when `None` and making every `AgentSnapshot` undecodable while such a device is present.

## Changes

- `crates/openlogi-ipc/src/ipc.rs:55` — bump `PROTOCOL_VERSION` 23→25: v24 fixes bincode for `registry_model_id`, v25 appends `AgentStatus::input_monitoring_granted` (with `#[expect(clippy::struct_excessive_bools)]`)
- `crates/openlogi-core/src/device.rs:410` — drop `skip_serializing_if`, keep `#[serde(default)]`
- `crates/openlogi-agent-core/src/watchers/input_monitoring.rs` (new) + `watchers/mod.rs` — poll `openlogi_hid::permissions::has_access()` like the accessibility watcher
- `crates/openlogi-agent-core/src/observable.rs:46` — seed and publish `input_monitoring_granted`; add `set_input_monitoring_granted`
- `crates/openlogi-agent/src/main.rs:318` — spawn watcher and handle `input_monitoring_rx`; `#[expect(clippy::too_many_lines)]` for select
- `crates/openlogi-agent/src/bin/mock_agent.rs:620` — mock status granted, remove #571 workaround comment
- `crates/openlogi-desktop/src/windows/settings/permissions.rs:54` — read `AppState::agent_status().input_monitoring_granted` instead of local `permissions::input_monitoring()` (Unknown while not connected)
- `crates/openlogi-ipc/tests/wire_format.rs:66` — make `assert_wire` round-trip (deserialize + re-encode) without requiring `PartialEq` on tarpc types; regenerate goldens for `AgentStatus`/snapshot/observation and `registry_model_id = None` (`…6100`)

## Testing

- `DEVELOPER_DIR=/Library/Developer/CommandLineTools cargo test -p openlogi-ipc --test wire_format` — 13 passed (snapshot/observation with new bool, standalone None now decodable)
- `cargo test -p openlogi-agent-core --lib observable` — 10 passed
- `cargo clippy -p openlogi-core -p openlogi-ipc -p openlogi-agent-core -p openlogi-agent --all-targets -- -D warnings` — clean (after adding expects)
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay` — clean
- Manual: not runtime-tested on hardware — logic is polling `has_access()` and wiring through IPC; mock serves granted.

## Notes

Depends on #759 (v24). This branch stacks v25 on top and currently includes #571's fix as well; will rebase to keep 24→25 sequential once #759 merges (or this can supersede #759 if preferred).

Fixes #571
Fixes #606

### DIFF
--- a/crates/openlogi-agent-core/src/observable.rs
+++ b/crates/openlogi-agent-core/src/observable.rs
@@ -50,6 +50,7 @@ impl ObservableState {
                     inventory: InventoryHealth::Scanning,
                     protocol_version: PROTOCOL_VERSION,
                     agent_version,
+                    input_monitoring_granted: openlogi_hid::permissions::has_access(),
                 },
                 inventory: Vec::new(),
                 standalone: Vec::new(),
@@ -164,6 +165,18 @@ impl ObservableState {
                 return false;
             }
             snapshot.status.accessibility_granted = granted;
+            true
+        });
+    }
+
+    /// Publish an Input Monitoring trust change, as observed by
+    /// [`watchers::input_monitoring`](crate::watchers::input_monitoring).
+    pub fn set_input_monitoring_granted(&self, granted: bool) {
+        self.update(|snapshot| {
+            if snapshot.status.input_monitoring_granted == granted {
+                return false;
+            }
+            snapshot.status.input_monitoring_granted = granted;
             true
         });
     }
--- a/crates/openlogi-agent-core/src/watchers/input_monitoring.rs
+++ b/crates/openlogi-agent-core/src/watchers/input_monitoring.rs
@@ -1,0 +1,40 @@
+//! Input Monitoring permission polling watcher.
+
+use std::thread;
+use std::time::Duration;
+
+use tokio::sync::mpsc;
+use tracing::{debug, warn};
+
+/// Watch macOS Input Monitoring permission changes.
+pub fn spawn(period: Duration) -> mpsc::UnboundedReceiver<bool> {
+    let (tx, rx) = mpsc::unbounded_channel();
+
+    if !cfg!(target_os = "macos") {
+        let _ = tx.send(true);
+        let _ = period;
+        return rx;
+    }
+
+    let spawn_result = thread::Builder::new()
+        .name("openlogi-input-monitoring-watcher".into())
+        .spawn(move || {
+            let mut last: Option<bool> = None;
+            loop {
+                let granted = openlogi_hid::permissions::has_access();
+                if last != Some(granted) {
+                    debug!(granted, "input monitoring trust changed");
+                    if tx.send(granted).is_err() {
+                        debug!("input monitoring watcher receiver dropped — exiting");
+                        return;
+                    }
+                    last = Some(granted);
+                }
+                thread::sleep(period);
+            }
+        });
+    if let Err(e) = spawn_result {
+        warn!(error = %e, "could not spawn input monitoring watcher — status won't auto-refresh");
+    }
+    rx
+}
--- a/crates/openlogi-agent-core/src/watchers/mod.rs
+++ b/crates/openlogi-agent-core/src/watchers/mod.rs
@@ -7,6 +7,7 @@ pub mod camera;
 pub mod foreground_app;
 pub mod gesture;
 pub mod host_switch;
+pub mod input_monitoring;
 pub mod inventory;
 pub mod keyboard;
 pub mod pairing;
--- a/crates/openlogi-agent/src/bin/mock_agent.rs
+++ b/crates/openlogi-agent/src/bin/mock_agent.rs
@@ -429,10 +429,7 @@ fn standalone_light() -> StandaloneDevice {
             zones: false,
         }),
         driver_id: "litra".to_string(),
-        // Must stay `Some` until #571 is fixed: `registry_model_id` is
-        // `skip_serializing_if`, which truncates the bincode stream when it is
-        // `None` and makes the whole snapshot undecodable. `8c900` is also the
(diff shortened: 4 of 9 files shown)


davidbudnick (dismissed):

AprilNEA (approved):
Thanks!

PR #796, opened 2026-08-22
fix(linux): keep the generated systemd unit out of the user's tier

### LINKED ISSUE #763: [Bug]: Linux install improvements for Debs and systemd
### Pre-flight checklist

### Which part of OpenLogi?

Not sure

### OpenLogi version

Install-time issues

### Operating system

Linux

### OS version & architecture

Ubuntu 26.04

### Device model

Install-time issues

Other / not sure

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

Hey, this looks like a cool system! While trying it out, I found a few quality of life papercuts on Linux that you might want to address.

First, the 'easy install' command line (`sudo dpkg -i openlogi_*.deb`) as specified in https://github.com/AprilNEA/OpenLogi#linux and https://openlogi.org/docs/getting-started/installation#linux doesn't work because you've changed the separator on the Debian releases (at least) from an underscore to a dash.

Second, the instructions [for autostart](https://github.com/AprilNEA/OpenLogi/blob/master/docs/INSTALL-linux.md#autostart-launch-at-login) say that the user unit is installed in `~/.config/systemd/user/openlogi-agent.service` automatically upon setting a GUI option. This is not correct for Linux

There is a whole specification for where certain types of programs installed in certain types of ways are supposed to put their unit files, in Linux. The reason being, locations "closer" to the user and their profile that are installed manually by the user should have precedence over locations for the entire system or that were installed by a package manager. See https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Unit%20File%20Load%20Path, Table 2. 

You have chosen one of the highest priority locations, which should be reserved for me as a user. It seems that, for package installs on Debian, RedHat, etc. (I can't speak to NixOS), you should be installing the user unit into `/usr/lib/systemd/user`. It won't change anything about the activation command (`systemctl --user enable --now openlogi-agent.service`). But it will allow users the proper control on Linux. 

Finally, while filling out this bug report, the template says "Linux and Windows are still stubs." But that does not seem to be true, there seems to be pretty full support for Linux and Windows now.

### Steps to reproduce

1. Download Debian release
2. Run the command `sudo dpkg -i openlogi_*.deb`
3. Receive error message `dpkg: error: cannot access archive 'openlogi_*.deb': No such file or directory`
4. Change command to `sudo dpkg -i openlogi-*.deb`
5. Deb package is installed

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

AalmanSadath:
## Summary

The third item in #763, and the only behavioural one: `launch_at_login` wrote its unit to `$XDG_CONFIG_HOME/systemd/user` — the highest-precedence tier in systemd's user load path, and the one reserved for units the user writes themselves.

```
~/.config/systemd/user        ← the user's own tier   (what we wrote into)
/etc/systemd/user             ← sysadmin
~/.local/share/systemd/user   ← installed on the user's behalf
/usr/lib/systemd/user         ← packages              (where OpenLogi ships its unit)
```

Three consequences followed.

**A hand-edited unit could not survive.** `reconcile` runs on every agent start and every config reload, rewriting the file whenever its content differed. An added `Environment=`, a changed `RestartSec`, a `WantedBy=default.target` — reverted within one restart, with no message. Nothing marked the file as generated, so a deliberate edit and a stale artifact were indistinguishable.

**A packaged unit was shadowed permanently.** For a package install the generated copy is byte-identical to `/usr/lib/systemd/user/openlogi-agent.service` — same directives, same `ExecStart`. It added no capability, only precedence. Later changes to the packaged unit would never reach anyone who used the toggle, and `uninstall.sh` cannot remove the copy, because no package script can reach a home directory.

**`ExecStart` tracked whichever binary started last.** Running a build from source once pinned autostart to a path inside the build tree; a later `cargo clean` left a unit that fails at every login while the installed binary sits unused.

## Changes

**`openlogi-agent`** — `launch_agent.rs`

- A packaged unit that already launches this binary is enabled as-is; nothing is generated. Ownership stays with the package manager, and removing the package removes the unit. The probe walks the system tiers in systemd's own precedence order and compares `ExecStart` against the running executable, resolved through symlinks.
- Otherwise the unit is generated under `$XDG_DATA_HOME/systemd/user` — the tier systemd reserves for units installed on a user's behalf. It still outranks the packaged unit, but sits below the user's own, so a hand-written unit wins with no provenance check needed.
- A unit left in the config tier by earlier versions is removed, but only when provably ours. One template has ever shipped, parameterised solely by `ExecStart`, so splicing a file's own `ExecStart` line back into that template reproduces it byte for byte if and only if we rendered it. Anything carrying another directive is the user's: left in place, still winning.
- `render_unit` is split from `render_unit_with_exec` so that check can splice a line in verbatim. Re-running the escaper over an already-escaped value would double `%%` into `%%%%` and stop a unit we wrote from matching itself.

**`openlogi-core`** — `xdg_data_home()`, the counterpart to the existing `xdg_config_home()`.

**Docs** — `INSTALL-linux.md` describes both paths and states that `~/.config/systemd/user` stays the user's, pointing at `systemctl --user edit` for a drop-in that survives upgrades. The packaged unit's header comment no longer advertises the two copies as coexisting.

The test modules move to sibling files under `launch_agent/`; those lines are relocated unchanged, and the split keeps the module near the workspace's file-size guideline.

## Testing

```
cargo fmt --all -- --check
cargo test --workspace
cargo xtask ci      # 8 passed, 0 failed, 1 skipped
```

`tests (macos)` **did not run** — this is a Linux host. The diff is cfg-gated, so that is worth stating plainly: Linux is covered natively (both host and target of every new function) and the Windows proxy passed, but macOS is unverified here. Every new function is `#[cfg(target_os = "linux")]` and no macOS path changed. rustdoc was also run directly against `openlogi-agent`, which CI's rustdoc job excludes.

12 new unit tests cover the provenance check: a rendered unit is recognised for any executable path, including ones needing `%`, `$`, space and quote escaping; an added, changed, or dropped directive is not ours; two `ExecStart` lines is not a shape the renderer can emit; escaping round-trips; symlinked paths compare equal.

Runtime-verified against the built agent with the XDG bases redirected to a scratch tree, on Fedora 44 / systemd 259:

| Setup | Result |
|---|---|
| packaged unit naming this binary | nothing generated, packaged unit enabled |
| the same, plus a redundant generated unit | redundant file removed, packaged unit enabled |
| packaged unit naming a *different* binary | unit generated for the running path |
| stale unit in the config tier, naming an old checkout | removed, unit generated |
| hand-edited unit in the config tier (`Environment=RUST_LOG=debug`) | byte-identical afterwards, warned, left winning |

The packaged cases were exercised by placing a unit at `/usr/lib/systemd/user/openlogi-agent.service`; the load-path precedence was confirmed with `systemd-analyze --user unit-paths` and `systemctl --user show -p FragmentPath`.

## Notes

The template has never changed since Linux autostart shipped in v0.6.8 — byte-identical across every commit that touched it — so the provenance check needs no template history. The one intermediate rendering that differed (`$` escaping) existed for 71 minutes on the #172 branch and never shipped.

One gap remains, unchanged by this PR: `uninstall.sh` cannot clean a home directory, so a config-tier unit from an older install survives if a user upgrades and uninstalls without ever launching the agent. Package scripts have never been able to reach `$HOME`.

Fixes #763.

### DIFF
--- a/crates/openlogi-agent/src/launch_agent.rs
+++ b/crates/openlogi-agent/src/launch_agent.rs
@@ -24,13 +24,26 @@
 //!
 //! ## Linux
 //!
-//! A systemd **user** unit at
-//! `$XDG_CONFIG_HOME/systemd/user/openlogi-agent.service` (default
-//! `~/.config/systemd/user/openlogi-agent.service`) is written/removed, then
-//! `systemctl --user daemon-reload` and `enable`/`disable` are called.
+//! When a packaged unit already launches this exact binary — installed by a
+//! distribution package, `install.sh`, or an administrator — it is simply
+//! enabled. Generating a second copy would duplicate its directives one tier
+//! higher, shadowing later changes to it and outliving its removal, since no
+//! package script can clean a home directory.
+//!
+//! Otherwise a systemd **user** unit at
+//! `$XDG_DATA_HOME/systemd/user/openlogi-agent.service` (default
+//! `~/.local/share/systemd/user/openlogi-agent.service`) is written/removed,
+//! then `systemctl --user daemon-reload` and `enable`/`disable` are called.
+//! That is the tier systemd reserves for units installed *on* a user's behalf;
+//! `$XDG_CONFIG_HOME/systemd/user` outranks it and belongs to the user, so a
+//! unit they author by hand always wins over this generated one.
 //! `Restart=on-failure` mirrors the macOS `KeepAlive=SuccessfulExit:false`
 //! semantics. A clean `exit(0)` leaves the unit enabled but stopped until the
 //! next session login.
+//!
+//! Earlier versions wrote into `$XDG_CONFIG_HOME/systemd/user` directly; such a
+//! file is removed on reconcile, but only when it round-trips through the
+//! renderer and is therefore provably one of ours.
 
 use tracing::debug;
 
@@ -38,6 +51,8 @@ use tracing::debug;
 use std::fmt;
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 use std::io;
+#[cfg(target_os = "linux")]
+use std::path::Path;
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 use std::path::PathBuf;
 #[cfg(target_os = "windows")]
@@ -214,10 +229,40 @@ fn reconcile_windows(enabled: bool) -> std::io::Result<()> {
 #[cfg(target_os = "linux")]
 const UNIT_NAME: &str = "openlogi-agent.service";
 
+/// Unit directories outside the two user-writable tiers, in systemd's own
+/// precedence order (`systemd.unit(5)`, "Load path when running in user mode").
+///
+/// Probed to answer one question: has something outside this app's control —
+/// a distribution package, `install.sh`, or an administrator — already
+/// installed this unit? The user tiers are deliberately absent: a file there
+/// is either ours or the user's, never a package's.
+#[cfg(target_os = "linux")]
+const SYSTEM_UNIT_DIRS: &[&str] = &[
+    "/etc/systemd/user",
+    "/usr/local/share/systemd/user",
+    "/usr/share/systemd/user",
+    "/usr/local/lib/systemd/user",
+    "/usr/lib/systemd/user",
+];
+
 #[cfg(target_os = "linux")]
 fn reconcile_linux(enabled: bool) -> io::Result<()> {
-    let path = unit_path()?;
     let exe = std::env::current_exe()?;
+    migrate_legacy_unit();
+
+    // A packaged unit that already launches this exact binary makes a generated
+    // copy pure redundancy — identical directives, one tier higher. Writing one
+    // would shadow the package (later changes to the packaged unit would never
+    // reach this user) and outlive it (uninstall cannot reach a home
+    // directory). Enable the packaged unit instead and write nothing.
+    if enabled && let Some(packaged) = packaged_unit_for(&exe) {
+        remove_generated_unit()?;
+        info!(path = %packaged.display(), "enabling the packaged systemd user unit");
+        run_systemctl(&["enable", UNIT_NAME]);
+        return Ok(());
+    }
+
+    let path = generated_unit_path()?;
     let desired = enabled.then(|| render_unit(&exe.to_string_lossy()));
 
     let current = std::fs::read_to_string(&path).ok();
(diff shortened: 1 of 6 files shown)


crates/openlogi-agent/src/launch_agent.rs
```diff
@@ -243,28 +288,182 @@ fn reconcile_linux(enabled: bool) -> io::Result<()> {
             run_systemctl(&["daemon-reload"]);
             info!(path = %path.display(), "systemd user unit removed");
         }
-        (None, None) => debug!("systemd user unit already absent"),
+        (None, None) => {
+            debug!("systemd user unit already absent");
+            // The packaged unit, if any, may still be enabled from an earlier
+            // run; disabling is cheap and idempotent when it is not.
+            run_systemctl(&["disable", UNIT_NAME]);
```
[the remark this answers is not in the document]
AalmanSadath: Good catch, fixed in 7b60a523. Confirmed the reported behaviour with a systemctl shim: with launch_at_login off and a user-authored config-tier unit, the old branch called systemctl --user disable, withdrawing an enablement OpenLogi never made.

Removing the call outright would regress the opposite case, though: the packaged-unit path writes no file, so (None, None) is also reached when the toggle is switched off on a package install, and without a disable there autostart would survive being turned off. The call is now gated on packaged_unit_for(&exe).is_some() && !user_authored_unit_present(), withdraw only an enablement this app could have made. Verified across all four combinations of packaged / user-authored.

crates/openlogi-agent/src/launch_agent.rs
```diff
@@ -235,36 +308,249 @@ fn reconcile_linux(enabled: bool) -> io::Result<()> {
...
+}
+
+/// Enable the unit and record that this app is the one that did.
+#[cfg(target_os = "linux")]
+fn enable_unit() {
+    run_systemctl(&["enable", UNIT_NAME]);
+    if let Err(e) = record_enablement() {
+        warn!(error = %e, "could not record the autostart enablement");
+    }
+}
```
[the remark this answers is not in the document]
AalmanSadath: The success check is a real bug — fixed. run_systemctl returned () and swallowed failures into a warn!, so the marker was written even when enable never happened (no session bus, unit not found). A later reconcile would then read it as proof and disable something OpenLogi never turned on. run_systemctl now reports success, the marker is written only when enable actually succeeded, and it is deliberately kept when disable fails so the next reconcile retries instead of stranding an enablement we still own. Verified with a shim that fails a chosen verb: enable-fails leaves no marker, disable-fails keeps it, both success paths behave as before.

Claiming an existing enablement is intended, and I want to keep it. Reaching the toggle at all is an explicit request for OpenLogi to manage autostart, so the setting has to mean what it says when it is switched back off. If we declined to disable in that case, Launch at login: off would leave the agent still starting at login, and the GUI would be showing a state that isn't true — which is the failure mode this PR set out to remove, just relocated to the toggle. Cycling on-then-off clearing a hand-made enablement is the honest outcome of two deliberate user actions.

Worth noting the scope difference: the enablement is one shared bit on a shared unit name, so it can't be co-owned. What OpenLogi does not touch is anything it did not create — no unit file at either user-writable tier is written over or deleted unless it round-trips through the renderer, and a unit installed by anything else is left alone entirely.

crates/openlogi-agent/src/launch_agent.rs
```diff
@@ -235,36 +308,284 @@ fn reconcile_linux(enabled: bool) -> io::Result<()> {
...
+fn enable_unit() {
+    if let Err(e) = record_enablement() {
+        warn!(
+            error = %e,
+            "could not record the autostart enablement; leaving autostart unchanged",
+        );
+        return;
+    }
+    if !run_systemctl(&["enable", UNIT_NAME]) {
+        clear_enablement_marker();
```
AprilNEA: **Preserve an existing claim when a redundant enable fails.** `record_enablement()` returns success both when this call creates the marker and when a marker from an earlier successful enable already exists. In the latter case, a transient failure from this repeated `systemctl enable` clears valid ownership here even though the previously created enable symlink may remain. A later disabled reconcile then sees no marker, skips `disable`, and leaves autostart active while the setting reads off. Please distinguish a newly created marker from a pre-existing one and only roll back the former; add a regression for `existing marker + enable failure`.

crates/openlogi-agent/src/launch_agent.rs
```diff
@@ -235,36 +308,284 @@ fn reconcile_linux(enabled: bool) -> io::Result<()> {
...
+///
+/// A unit that launches some *other* binary is not a match: the generated unit
+/// is what makes autostart work for a build the packaged unit cannot describe,
+/// so in that case the normal write path must still run.
+#[cfg(target_os = "linux")]
+fn packaged_unit_for(exe: &Path) -> Option<PathBuf> {
+    SYSTEM_UNIT_DIRS
+        .iter()
+        .map(|dir| Path::new(dir).join(UNIT_NAME))
+        .find(|path| {
```
AprilNEA: **Check the highest-precedence existing unit, not the first unit whose executable matches.** This predicate skips an existing higher-tier unit when its `ExecStart` differs and can then accept a lower matching unit. For example, if `/usr/local/share/systemd/user/openlogi-agent.service` names an old binary while `/usr/lib/systemd/user/openlogi-agent.service` names the running one, this returns the `/usr/lib` file and removes the generated data-tier override, but systemd actually loads the higher `/usr/local/share` file. The probe should resolve the effective first existing unit in the real user load path and only reuse it when that unit matches. The hard-coded path set also omits `$XDG_CONFIG_DIRS/systemd/user`, `/run/systemd/user`, and custom `$XDG_DATA_DIRS`, so please cover the actual precedence rather than a partial static list.

AprilNEA (changes requested):
Two Linux autostart ownership issues still need to be addressed before merge: preserve an existing ownership marker across a transient re-enable failure, and resolve the effective highest-precedence system unit rather than any lower matching entry. Please add automated regression coverage for both state transitions; the current helper tests do not exercise them.

crates/openlogi-agent/src/launch_agent.rs
AalmanSadath: The probe now resolves the first unit that exists and reuses it only when that one names the running binary. Your shadowed-/usr/local/share example is a test.

Path resolution comes from systemd-analyze --user unit-paths, filtered to drop the two tiers this app writes to, so what remains is every directory a unit can arrive in from elsewhere, in systemd's own precedence order. The old static list named 6 directories where a plain Fedora session has 15. Among the misses was $XDG_RUNTIME_DIR/systemd/user, which outranks the tier this app generates into, so a unit there shadowed ours and the probe couldn't see it. The static list survives only as a fallback for when systemd-analyze can't be run, and is now derived from $XDG_CONFIG_DIRS, $XDG_RUNTIME_DIR and $XDG_DATA_DIRS rather than hardcoded.

Regression tests cover both, and reverting either fix fails its test.

crates/openlogi-agent/src/launch_agent.rs
AalmanSadath: Fixed in 7e995359.

record_enablement_at now reports whether the claim was created by that call, and only a claim made there is rolled back. A claim held from an earlier successful enable survives a transient failure, so the enable symlink and the marker stay in agreement and a later disabled reconcile still finds ownership to withdraw.

Regression added for exactly that case: pre-existing marker plus a failed enable leaves the marker in place. Reverting the scoping fails it. The companion case is covered too, where a claim created by the same call is rolled back when the enable fails, so no marker outlives an enablement that never happened.

AprilNEA: Verified your fix and preserved it in the rebased head, [7634a5ff](https://github.com/AprilNEA/OpenLogi/commit/7634a5ff18ebda3b8032d675f22eb8ac0221c32d). A pre-existing claim survives a failed re-enable, while a claim created by the failing call is rolled back. Both regressions pass in the current `autostart/linux` module. This finding is addressed.

crates/openlogi-agent/src/launch_agent.rs
AprilNEA: Your precedence and load-path fixes are preserved in [7634a5ff](https://github.com/AprilNEA/OpenLogi/commit/7634a5ff18ebda3b8032d675f22eb8ac0221c32d). I also fixed the remaining mask case: `is_file()` skipped a symlink to `/dev/null`, allowing a shadowed lower unit to match. Lookup now stops at that entry. The regression covers `/dev/null` masks, empty-file masks, and ordinary unit symlinks.

The branch is rebased onto current master, with contributor credit retained and the current macOS/Windows implementations unchanged. The full local gate passed: fmt, workspace Clippy, workspace tests (1646 passed, 2 ignored), and non-GUI rustdoc, with warnings denied. Agent/core Windows GNU cross-Clippy and wasm portability checks also passed. Not runtime-tested on hardware or in a systemd user session; please verify login startup and on/off toggling in a Linux desktop session. Fresh CI and Greptile checks are running. This finding is addressed.

crates/openlogi-agent/src/autostart/linux.rs
```diff
@@ -41,34 +193,295 @@ fn apply(enabled: bool) -> io::Result<()> {
...
+/// created by this call rather than already held from an earlier reconcile.
+fn record_enablement_at(path: &Path) -> io::Result<bool> {
+    if path.exists() {
+        return Ok(false);
+    }
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    std::fs::write(path, b"")?;
+    Ok(true)
```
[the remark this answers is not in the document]
AalmanSadath: Real. `server.rs` spawns a task per request and `reload_config` drops the orchestrator lock before `reconcile`, so two fast toggles overlap. Fixed in af2b76ee: reconciles are serialised process-wide, and the marker is claimed with `create_new` so at most one caller is told it made the claim. Both have regressions that fail on the previous code (19/20 and 20/20 runs).

crates/openlogi-agent/src/launch_agent.rs
AalmanSadath: Runtime-tested 7634a5ff + af2b76ee on Fedora 44 / systemd 259 in a GNOME session. On/off toggling through `reload_config` and through startup, 8-way concurrent reloads in both states (strictly serial in the log), `systemctl --user start` from the generated unit, and a real logout/login: the user manager started the branch build from `~/.local/share/systemd/user/` at login, IPC and the hook came up within 2 s. Everything reverts to absent on off.

AprilNEA (approved):

PR #811, opened 2026-08-23
feat(hid): recognise Lightspeed receiver 046d:c54d (PRO X SUPERLIGHT 2 DEX)

v14:
## Summary

Recognises the `046d:c54d` Lightspeed receiver shipped with the Logitech PRO X SUPERLIGHT 2 DEX.

Before this change, `receiver::detect()` did not recognise the receiver, so its paired mouse never reached inventory and was absent from both `openlogi list` and the GUI. The receiver answers the same HID++ 1.0 enumeration registers as the existing Lightspeed receivers, so it uses the established Unifying-compatible route.

## Changes

- Add `0xc54d` to the shared Lightspeed receiver PID table.
- Add `046d:c54d` to HID++ receiver detection.
- Cover the receiver display name with the existing regression test.

No routing or pairing logic changes are needed. The shared Lightspeed tables already drive route selection, Linux receiver-child filtering, pairing-family classification, and the user-facing receiver name.

## Hardware verification

Tested on macOS 26.6 with:

- Receiver: `046d:c54d`
- Device: PRO X SUPERLIGHT 2 DEX
- Paired-device WPID: `0x40b8`

Before the patch, OpenLogi v0.7.4 did not list the receiver or mouse. With the patch, the receiver enumerates its slot, the device is persisted in configuration, and the GUI shows the mouse with its battery state and catalog artwork. `openlogi diag features` completed with 33 feature entries.

Pairing and unpairing were not exercised because that would change the receiver's working pairing state. Device-specific feature work beyond discovery is out of scope.

## Validation

- `cargo test --workspace`

Refs #512

### DIFF
--- a/crates/openlogi-core/src/hid/route.rs
+++ b/crates/openlogi-core/src/hid/route.rs
@@ -94,11 +94,12 @@ pub const UNIFYING_PIDS: &[u16] = &[0xc52b, 0xc532, 0xc537];
 /// G502 LIGHTSPEED and the G Pro Wireless — its USB product string is
 /// literally `LIGHTSPEED Receiver`; `0xc53f` is the nano receiver of wireless
 /// mice such as the G305; `0xc547` ships with newer G-series devices such as
-/// the G915 keyboard and the G502 X LIGHTSPEED.
+/// the G915 keyboard and the G502 X LIGHTSPEED; `0xc54d` ships with the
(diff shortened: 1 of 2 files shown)


AprilNEA:
Nice catch.

AprilNEA (approved):

PR #802, opened 2026-08-23
feat(i18n): add Turkish locale, plus four Linux/asset fixes

4ni1ak:
## Summary

Adds the Turkish interface locale and fixes four reported bugs, none of which
has an open PR already. Every commit is self-contained and conventional, so this
can be **rebase-merged**; squashing would collapse unrelated fixes into one
subject. Happy to split it per issue if you would rather review them separately.

An earlier revision of this branch also carried fixes for #770, #752 and #688.
Those were dropped once I found #790, #765 / #719 and #741 already open for them
— no point in a second implementation of work someone else is waiting on.

## Changes

**`openlogi-ui`**
- New `locales/tr.yml` carrying every key `en.yml` defines. `tr` ("Türkçe") is
  registered in `SUPPORTED` between Svenska and Ελληνικά — native-name
  alphabetical within the Latin block — and added to the parity test's
  `include_str!` table, so the catalog is actually checked rather than silently
  drifting. `crowdin.yml` gains the language and its `tr -> tr` mapping.

**`openlogi-desktop`**
- Depot metadata is now resolved through the manifest's `image_metadata` before
  the well-known `METADATA_FILES` names. Depots whose variants are *handed*
  rather than coloured ship none of those names — the Lift keys its metadata
  `core_metadata_left.json` / `core_metadata_right.json` — so the lookup missed,
  `resolve` returned `None` for every root, and every Lift rendered the generic
  silhouette with a complete bundle on disk. Manifest-sourced names go through
  `safe_component_path` like every other asset file, and the download side adds
  `image_metadata` to its manifest-mapped resource pass (consulting every
  model-id candidate there too, as the resolver already did).
- The sync's depot list is deduplicated on `(depot, ext)` rather than the depot
  alone. Two devices sharing a depot but differing in `extended_model_id` — a
  colour pair, or the Lift's left- and right-handed variants — collapsed to one
  target, so the second never had its `device_image` /
  `device_buttons_image` / `image_metadata` fetched and stayed on fallback
  artwork with no hotspot metadata. That also undid the fix above for the
  dropped device. The collection step moved into its own function so the
  decision is testable without the surrounding HTTP. (Found by the Greptile
  review on this PR.)
- `IN_APP_UPDATES` gates the updater. A Linux release publishes distro packages
  only and `xtask release latest-json` deliberately classifies none of them into
  the manifest, but the GUI asked anyway, so Settings → Updates answered
  "Update failed: no release asset matched the current platform (linux/x86_64)".
[...]

PR #835, opened 2026-08-23
feat(i18n): add Turkish (tr) interface locale

4ni1ak:
## Summary

Adds Turkish as a shipped interface locale.

Split out of #802 at your request — that PR is closed and its five parts are now
independent.

## Changes

**`openlogi-ui`**
- New `locales/tr.yml` carrying every key `en.yml` defines.
- `tr` ("Türkçe") registered in `SUPPORTED`, between Svenska and Ελληνικά —
  native-name alphabetical within the Latin block, matching the picker order the
  constant documents.
- Added to `locale_files_have_the_same_keys`'s `include_str!` table. That list is
  hand-written and asserts its own coverage against `SUPPORTED`, so a locale left
  out of it fails the build rather than silently going unchecked.
- `maps_locale_variants` gains `tr` and `tr-TR`.

**`crowdin.yml`**
- `tr` added to `export_languages` and mapped `tr -> tr`, so the catalog
  round-trips like every other one.

Values are a best-effort human translation, not machine fill-in; Crowdin can
refine them from here.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo test -p openlogi-ui locale     # catalog parity + negotiation
cargo test -p openlogi-desktop i18n  # end-to-end key resolution
cargo fmt --all -- --check
cargo test --workspace
```

All green.

**Not run on this host:** `tests (macos)`, `cargo-deny`, and macOS clippy (no
macOS SDK here). No hardware verification applies — this adds no device code.

**Note on ordering:** if the Linux package-manager update PR lands first, it adds
one string to every catalog that exists at that point, so this branch needs a
rebase to pick it up in `tr.yml` before the parity test passes. Happy to rebase
whenever it suits you — same in the other direction.


AprilNEA (dismissed):

AprilNEA:
Thanks.

AprilNEA (approved):

Issue #769, opened 2026-08-21
[Bug]: "Update failed: no release asset matched the current platform (linux/x86_64)"


dradoaica:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.7.4

### Operating system

Linux

### OS version & architecture

Ubuntu 26.04 LTS, GNOME Version 50, Wayland

### Device model

MX Master 3S + Logi Bolt receiver

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [ ] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

Update failed: no release asset matched the current platform (linux/x86_64)

### Steps to reproduce

1. Open "OpenLogi"
2. Click "Settings"
3. Click "Updates"
4. Click "Check for Updates"

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

### macOS permissions (if applicable)

- [ ] OpenLogi has **Accessibility** permission (needed to remap buttons via the event tap).
- [ ] OpenLogi has **Input Monitoring** permission (needed for Bluetooth-direct devices and capture).

dradoaica:
Related issue: https://github.com/AprilNEA/OpenLogi/issues/605

AprilNEA:
Thanks for the report. This is already tracked in #605, so I’m closing this as a duplicate to keep the discussion in one place.

PR #838, opened 2026-08-23
fix(gui): stop offering in-app updates where no artifact ships

### LINKED ISSUE #605: [Bug]: Update failed: no release asset matched the current platform (linux/x86_64)
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.24

### Operating system

Linux

### OS version & architecture

Ubuntu 26.04 Win 64

### Device model

MX Master S3

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [x] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

I'm on Ubuntu 26.04 and I can't use the update feature in the app because it gives me the following error message:

Image

### Steps to reproduce

1. Go to Settings > Updates
2. Click the Check for Updates button
3. Observe the following error message `Update failed: no release asset matched the current platform (linux/x86_64)`

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.6.27 (release)
- Agent: v0.6.27 (connected)
- IPC protocol: GUI 16 / agent 16
- Inventory: ready
- OS: Linux (x86_64)
- Config: schema 4 · 1 configured device(s) · thumbwheel 14

**Assets**
- Cache path: ~/.local/share/openlogi/assets · Bundle assets: absent

**Devices (1)**
- MX Master 3S — mouse (codename: MX Master 3S)
  - Connection: Bluetooth (direct) · Online: yes · Battery: 90% (discharging, full)
  - Model: 0b034 (model-ids: b034/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_master_3s · direct
```

### `openlogi list` output

```text
Logitech MX Master 3S (—, vid=046d pid=b034)
  └─ slot 255 ● MX Master 3S (mouse, wpid=?, battery=90% full (discharging))
          model_ids=[b034,0000,0000] ext=00 serial=2540APFBE888 unit_id=8d0755ec transports=btle
```

### Logs

```shell

```

4ni1ak:
## Summary

Settings → Updates offered a check that can only fail on Linux.

Split out of #802 at your request.

Fixes #605

## Changes

A Linux release publishes distro packages only — `.deb`, `.rpm`,
`.pkg.tar.zst` — and `xtask release latest-json` deliberately classifies none of
them into the update manifest, because those installs update through the package
manager. The GUI asked anyway, so opening Settings → Updates and clicking Check
produced a red "Update failed: no release asset matched the current platform
(linux/x86_64)".

- One `IN_APP_UPDATES` const in `platform::updater` states whether this build has
  an in-place-updatable artifact at all.
- The launch-time check, the auto-install observer and `check_for_updates` skip
  the check where it does not. The menu item stays and still opens the page,
  which now explains where updates come from.
- The Updates page shows the running version and the update source instead of a
  check button and two switches that cannot do anything.
- macOS and Windows are unchanged.

The one new string is added to every shipped catalog at the same position.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo test -p openlogi-ui locale     # catalog parity
cargo test -p openlogi-desktop i18n  # key resolution
cargo fmt --all -- --check
cargo test --workspace
```

All green.

**Not run on this host:** `tests (macos)`, `cargo-deny`, macOS clippy (no macOS
SDK here). The macOS and Windows paths are unchanged by construction —
`IN_APP_UPDATES` is `true` there and every branch behaves as before — but that
is an argument, not a test run.

**Not runtime-tested.** To verify: open Settings → Updates on a Linux install
and confirm it states the package manager rather than offering a check; on macOS
confirm the check still runs and reports normally.

**Note on ordering:** this adds a string to every catalog that exists today. If
the Turkish locale PR lands first, this branch needs a rebase so `tr.yml` gets
the same key before the parity test passes — happy to rebase whenever it suits
you, and the same the other way round.


crates/openlogi-desktop/src/windows/settings/updates.rs
```diff
@@ -71,6 +75,43 @@ pub(super) fn updates_page(updater: Entity<Updater>, pal: Palette) -> SettingPag
...
+                v_flex()
+                    .gap_1()
+                    .min_w_0()
+                    .child(
+                        div()
+                            .font_weight(FontWeight::SEMIBOLD)
+                            .child(concat!("OpenLogi ", env!("CARGO_PKG_VERSION"))),
+                    )
+                    .child(div().text_caption().text_color(pal.text_muted).child(tr!(
+                        "Installed from a distribution package — updates come from your package manager."
```
AprilNEA: [P2] Do not infer the installation method from `target_os`. This branch runs for every Linux build, including the source-build path documented in `docs/INSTALL-linux.md` and reported in #605, yet it tells all of them they were installed from a distribution package. The official `.deb`/`.rpm` instructions also download artifacts manually and configure no package repository, so a package manager will not discover future releases on its own. Please use installation-neutral copy such as ‘In-app updates are unavailable on Linux; update using the method you installed OpenLogi with,’ or detect the actual install source.

crates/openlogi-desktop/src/windows/settings/updates.rs
```diff
@@ -71,6 +75,43 @@ pub(super) fn updates_page(updater: Entity<Updater>, pal: Palette) -> SettingPag
...
+                    ))),
+            )
+            .into_any_element()
+    }));
+
+    SettingPage::new(tr!("Updates"))
+        .icon(IconName::ArrowDown)
+        .resettable(false)
+        .group(hero)
+        .group(SettingGroup::new().item(SettingItem::render(move |_, _, _| update_source(pal))))
```
AprilNEA: [P2] Avoid reusing the updater-enabled source footer here. On this new Linux page, `update_source` still labels GitHub Releases as the update source and says OpenLogi connects when automatic checks are enabled or Check for Updates is clicked, even though this PR removes those controls and suppresses every check. Render a source/footer variant for this page so it presents one consistent update path.

crates/openlogi-desktop/src/app/menu.rs
```diff
@@ -118,7 +118,13 @@ pub fn rebuild(cx: &mut App) {
 /// Run a manual update check and open Settings → Updates where its status is
 /// rendered. Shared by the app menu and agent tray IPC commands.
 pub fn check_for_updates(cx: &mut App) {
-    if let Some(updater) = crate::platform::updater::shared(cx) {
+    // Where releases ship no in-place-updatable artifact (Linux: distro
+    // packages only), running the check just resolves to "no release asset
+    // matched the current platform". Open the Updates page anyway — it says
+    // where updates come from on this install.
+    if crate::platform::updater::IN_APP_UPDATES
```
AprilNEA, line 132: [P2] This deliberately skips the check on Linux, but the invoking menu item remains ‘Check for Updates…’ and the deeplink/command contract still says it runs a check. A Linux user selecting it gets navigation only. Please rename or hide this action on unsupported platforms, or give it platform-neutral ‘Updates…’ semantics.

crates/openlogi-desktop/src/platform/updater.rs
```diff
@@ -29,6 +29,17 @@ const MANIFEST_URL: &str = match option_env!("OPENLOGI_UPDATE_MANIFEST_URL") {
...
+/// Whether this build can update itself in place.
+///
+/// A release publishes a DMG for macOS and an MSI for Windows, and for Linux
+/// only distro packages (`.deb` / `.rpm` / `.pkg.tar.zst`) — those installs
+/// update through the package manager, so the manifest deliberately carries no
+/// Linux asset (see `xtask release latest-json`). Asking the updater anyway
+/// resolves to "no release asset matched the current platform", which reads as
+/// a failure rather than as the design it is, so Linux skips the check and the
+/// Updates page says where updates actually come from.
+pub const IN_APP_UPDATES: bool = !cfg!(target_os = "linux");
```
AprilNEA: [P3] Make this capability fail closed. The static manifest only ships in-app artifacts for macOS and Windows, but `!cfg!(target_os = "linux")` enables every other current or future target and recreates the same `NoMatchingAsset` failure there. Prefer a positive macOS/Windows predicate.

AprilNEA (changes requested):
The core Linux updater gating is sound and covers the startup, automatic, manual, Settings, and consent paths. The new Linux-facing UI still makes contradictory and incorrect claims about how the current installation is updated; please address the inline comments before merging. Also change the closing reference from `Fixes #769` to `Fixes #605`, because #769 is already closed as a duplicate and the open tracking issue would otherwise remain open. Per the project UI review requirements, please add a Linux runtime check and screenshot of the resulting Updates page.

4ni1ak:
Rebased onto current master and pushed (2c8d9cb3) — no content changes, just picking up the 64 commits master gained since this branch opened.

All four review points are already addressed in the code as it stands:

- P1 (consent bypasses gate): `update_consent.rs::answer` checks `IN_APP_UPDATES` before calling `Updater::check` — already gated.
- P2 (install-method copy): `unmanaged_updates_page` says "In-app updates aren't available on this platform — update OpenLogi the way you installed it," installation-neutral, no `.deb`/`.rpm` claim.
- P2 (source footer reuse): `unmanaged_update_source` is its own function, separate from `update_source` — no GitHub-Releases/"OpenLogi connects when..." copy on this page.
- P2 (menu item semantics): the menu label is conditional — "Check for Updates…" where `IN_APP_UPDATES`, "Updates…" otherwise — and the doc comment explains the item only opens the page there.
- P3 (fail-closed predicate): `IN_APP_UPDATES` is `cfg!(any(target_os = "macos", target_os = "windows"))`, a positive allowlist, not `!cfg!(linux)`.
- `Fixes #605` is already the closing reference in the PR body.

Gate on the rebased tip: fmt, `clippy --workspace --all-targets -D warnings`, `cargo test -p openlogi-ui locale`, `cargo test -p openlogi-desktop i18n` — all green.

Linux runtime check, Settings → Updates:

4ni1ak:
Screenshot_20260828_131445

4ni1ak:
Separate from this fix — worth flagging while we're on Linux distribution: right now OpenLogi has no official channel that gives users package-manager updates.

- AUR: only `openlogi-bin`, community-maintained, not from this repo.
- apt/Debian, Ubuntu: nothing.
- Flathub: nothing.

Happy to put together a PKGBUILD (AUR) and/or a Flatpak manifest if you're interested — either as project-maintained packages, so "update the way you installed it" (this PR's own copy) has a real answer beyond the .deb/.rpm downloads. Let me know which of these (if any) you'd want, and I'll open it as its own PR.

crates/openlogi-desktop/src/platform/updater.rs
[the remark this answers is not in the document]
4ni1ak: Fixed — `answer()` in `update_consent.rs` now checks `IN_APP_UPDATES` before calling `Updater::check`, same gate used by the startup and manual checks. The setting is still recorded either way; only the immediate check is skipped where no artifact ships.

crates/openlogi-desktop/src/windows/settings/updates.rs
4ni1ak: Replaced the target_os-inferred "installed from a distribution package" copy with installation-neutral text: "In-app updates aren't available on this platform — update OpenLogi the way you installed it." No install-source detection exists elsewhere in the codebase to reuse, so this doesn't claim a specific install method — covers both the source-build path from docs/INSTALL-linux.md (#605) and the .deb/.rpm downloads.

crates/openlogi-desktop/src/windows/settings/updates.rs
4ni1ak: Added a separate unmanaged_update_source() for this page instead of reusing update_source(). It keeps the Releases link/changelog row but drops the GitHub-Releases-as-update-source framing and states plainly that OpenLogi makes no update requests on this platform.

crates/openlogi-desktop/src/app/menu.rs
4ni1ak: Menu item now reads "Updates…" (not "Check for Updates…") when IN_APP_UPDATES is false, and check_for_updates() skips the Updater::check call in that case — it only opens Settings → Updates. Added the missing locale key for the new label to en.yml and every other locale/*.yml.

crates/openlogi-desktop/src/platform/updater.rs
4ni1ak: Flipped to a positive predicate: IN_APP_UPDATES = cfg!(any(target_os = "macos", target_os = "windows")), used everywhere the old !cfg!(target_os = "linux") check was reachable. A new/other target now fails closed by default.

PR #841, opened 2026-08-23
docs(agents): refresh the permissions skill for the landed fixes

AprilNEA:
## Summary

The macOS permissions skill described the world before #760, #814, #817 and #820 all landed. Left as-is it would actively mislead: it sends reporters to a log that (it says) has nowhere to go, spells the helper bundle `OpenLogiAgent.app`, tells the reader a bare `Failed to open device` needs manual interpretation, and lists already-fixed gaps as open.

Two of these are more than staleness. The skill's whole first section is a discriminator keyed on log strings, and the transport now emits *different* strings that classify themselves — a triage run against the old table would stall on a line that no longer exists. And the helper path is what a reporter is told to type into Go-to-Folder.

## Changes

- `§1`: the open-failure row splits into the two messages the transport now emits (grant missing → grant it; grant present → exclusive open or stale session).
- `§2`: helper is `OpenLogi Agent.app`, with a note that pre-rename bundles carry the old spelling under the same identity and that the grant survived the rename because TCC keys on the identifier.
- `§3`: ask for `~/.local/state/openlogi/agent.<date>.log` first; the foreground run is now the fallback it has become, caveat intact.
- `§5`: state that the transport already pairs open failures with the grant check, instead of prescribing it.
- `§7`: the known-broken list is replaced by the state it actually has (nothing outstanding in that path), still marked as a snapshot to re-verify.
- `scripts/diagnose.sh`: resolves either helper spelling, prints the log-file path when one exists, and its classification hints match the new messages. Also drops an `ls | head` in favour of a glob (SC2012); the ISO-dated names sort lexically.

## Testing

- Ran `diagnose.sh` against this machine's installed app: it resolves the renamed `OpenLogi Agent.app`, prints all three identities with `org.openlogi.agent` intact, verifies signatures, and finds the live log file.
- `cargo xtask ci shell` (shellcheck + shfmt) — PASS.
- Docs-only otherwise; no Rust touched.

## Side finding, filed separately

Reading that live log turned up an unrelated live bug: the agent has been emitting `role held by a tenant that never identified itself` twice a second for over an hour. `Event::SupersededAnonymously` warns but never evicts, so an unidentifiable overlay tenant wedges the role permanently — the Actions Ring stays dead and the log grows ~19 MB/day. Details in #842.

### DIFF
--- a/.claude/skills/openlogi-macos-permissions/SKILL.md
+++ b/.claude/skills/openlogi-macos-permissions/SKILL.md
@@ -18,7 +18,8 @@ list). Classify before doing anything else. The discriminator is one log line:
 | Agent log | Layer | TCC? |
 |---|---|---|
 | `HID++ candidate interfaces count=0` | enumeration | **No.** The device's HID++ collection never matched — unsupported device, or not connected. |
-| `failed to open HID++ channel — retrying next tick error=Message("Failed to open device")` | open | **Yes, usually.** `IOHIDDeviceOpen` was denied. Same string also covers exclusive access and a device that vanished — see §5. |
+| `failed to open HID++ channel … Failed to open device: Input Monitoring is NOT granted to this process…` | open | **Yes.** The message classifies itself — grant Input Monitoring to the identity named in §2. |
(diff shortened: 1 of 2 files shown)


AprilNEA:
Closing as superseded by c3629bd, which landed the same skill and diagnostic-script changes on master. The XDG_STATE_HOME follow-up will be handled separately.

PR #847, opened 2026-08-23
feat(hook): reverse the wheel in software for devices without 0x2121

### LINKED ISSUE #694: [Feature]: Software scroll inversion for RatchetWheel (0x2130) devices
### Pre-flight checklist

### Problem / motivation

OpenLogi disables per-device scroll inversion unless a device exposes HID++ `0x2121 HiResWheel` with the native inversion capability. A Logitech Signature Plus M750 L connected through a Logi Bolt receiver instead exposes `0x2130 RatchetWheel` and does not expose `0x2121`, so the GUI reports that native HID++ scroll inversion is unavailable.

The same mouse and connection support reversed scrolling in Logi Options+, which indicates that a software/host-side implementation is possible. OpenLogi 0.7.1 on macOS 26.6.2 reproduces the problem.

No existing M750 scroll-inversion or `0x2130` fallback issue was found.

### Proposed solution

Add a per-device software scroll-inversion path for `0x2130 RatchetWheel` devices.

Suggested behavior:

- Continue to prefer native `0x2121` inversion when the device advertises `has_invert`.
- Implement the `0x2130` get/set wheel-mode calls and wheel-movement notification decoding.
- When inversion is enabled on an `0x2130` device, divert wheel reports to HID++, invert the vertical delta, and inject the resulting OS scroll event through OpenLogi's existing input-injection layer.
- Restore ordinary HID reporting when the software inversion setting is disabled or OpenLogi exits.
- Keep the setting scoped to the physical mouse so the macOS trackpad direction is unchanged.
- Clearly label the GUI mode as a software fallback rather than native HID++ inversion.

If divert-and-inject is not reliable on every platform, an OS-hook transformation for the selected physical device could be used as the fallback instead.

### Alternatives considered

- macOS “Natural scrolling”: changes system-level behavior and does not provide OpenLogi's per-device separation.
- Manually setting `invert_scroll = true`: the agent intentionally gates the write on the live `scroll_inversion` capability.
- Forcing the persisted capability flags: cannot add `0x2121` firmware support and would make OpenLogi attempt an unsupported write.
- Treating `0x2130` as native inversion support: the RatchetWheel protocol only exposes HID/HID++ report diversion, not an inversion bit.

### Related area(s)

- [x] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [x] Configuration (TOML)
- [ ] Auto-update
- [x] Other

### Hardware and software

- Device: Logitech Signature Plus M750 L / Logi M750
- Connection: Logi Bolt receiver, VID `046d`, PID `c548`
- OpenLogi: `0.7.1`
- OS: macOS `26.6.2`, Apple Silicon
- Logi Options+: reversed scrolling works with the same device

### Redacted `openlogi list` output

```text
Logi Bolt Receiver (<redacted>, vid=046d pid=c548)
  └─ slot 2 ● Logi M750 (mouse, wpid=?, battery=85% full (discharging))
          model_ids=[b02c,0000,0000] ext=03 serial=<redacted> unit_id=<redacted> transports=btle
```

### `openlogi diag features` result

The device reports 32 features. Relevant entries:

```text
13  0x2130  v0
14  0x2201  v2
```

There is no `0x2121` entry.

### `openlogi diag wheel` output

```text
device: Logi M750 (slot 2 on receiver <redacted>)
Error: read HiResWheel mode

Caused by:
    device does not expose HID++ feature 0x2121
```

OpenLogi already names `0x2130` as `RatchetWheel` in the feature registry, but it has no implementation registered: https://github.com/AprilNEA/OpenLogi/blob/master/crates/openlogi-hidpp/src/feature/registry.rs

The available `0x2130` protocol description exposes report diversion and wheel-movement events, but no native inversion flag: https://lekensteyn.nl/files/logitech/x2130_ratchetwheel.html

### LINKED ISSUE #776: [Feature]: Support Inverse scroll direction without native HID++
### Pre-flight checklist

### Problem / motivation

There are some mouse models (e.g. M575S) that do not have native HID++ scroll inversion support. It would be great if the app could still support the inversion via other method. There are apps (Mos, Mac Mouse Fix) that do that, but it would be good to have the functionality integrated handily in one.

### Proposed solution

Use the technique that Mos and Mac Mouse Fix are using to support scroll inversion non-natively.

### Alternatives considered

_No response_

### Related area(s)

- [ ] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [ ] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

### Additional context

_No response_

4ni1ak:
## Summary

Scroll inversion was offered only where the firmware exposes `0x2121
HiResWheel` with its invert bit. A large part of the current Logitech line
does not — the Signature M650 / M750 family exposes `0x2130 RatchetWheel` and
no `0x2121` at all, and the M575S is in the same position — so the GUI answered
"This device does not report native HID++ scroll inversion support" for a mouse
Logi Options+ happily reverses.

This reverses the wheel in the capture layer for exactly those devices. One
implementation covers both reports, so it closes both.

Fixes #694
Fixes #776

## Where the inversion happens, and why there

Each backend transforms the event **in place, where it still carries its native
units**: the fields the reader used are the fields the writer negates, so
nothing is converted or rounded.

Re-injecting from the normalised `delta_y` — which #694 offers as the primary
option — was the first design and it is wrong on macOS. `usable_scroll_delta`
reads whichever of the point / fixed / line fields is non-zero, so a hi-res
wheel arrives as pixels and a detented one as lines; re-synthesising from the
single normalised value would have to pick a unit and would be wrong for one of
them. Negating the same fields sidesteps the choice.

- **Linux** — `REL_WHEEL` / `REL_WHEEL_HI_RES` negated at the point the grabbed
  event is re-emitted through the pass-through uinput device.
- **macOS** — the vertical scroll fields are negated on the tapped event before
  it is passed through, scoped to non-trackpad sources the hook would remap
  buttons for, so a second vendor's mouse and the trackpad are untouched.
- **Windows** — not supported, and *not* because `WH_MOUSE_LL` hands back a
  read-only event: suppressing and re-injecting would be exact there. The
  blocker is attribution. The Windows hook reports no device for a wheel event
  (`MouseEvent::Scroll` always carries `device: None`), so the inversion could
  not be confined to the configured mouse, and applying it to every wheel event
  would reverse a precision touchpad too — the one thing this setting must not
  do. `SOFTWARE_SCROLL_INVERSION` is `false` there and the GUI keeps reporting
  the device as unsupported.

## Changes

**`openlogi-hook`**
- `set_scroll_inversion` / `scroll_inversion`, and the
  `SOFTWARE_SCROLL_INVERSION` capability const.
- The Linux and macOS transforms above.
- `examples/invert_scroll.rs`: a manual smoke test that grabs, runs one window
  with inversion off and one with it on, then releases. Two phases on purpose —
  a single window cannot tell an applied inversion from a desktop that was
  already scrolling that way.

**`openlogi-agent-core`**
- The orchestrator publishes the flag for the selected device only, and only
  when its firmware has no invert bit: a device that has one still gets the
  HID++ write from `configured_wheel_mode`, and doing both would cancel out.
  Disabling the device, deselecting it, or turning the setting off clears it.

**`openlogi-desktop`**
- `current_scroll_inversion_available` / `_is_software`; the toggle no longer
  gates on native support.
- The description now distinguishes the two, because "OpenLogi reverses this as
  it captures it" is a different guarantee from the device doing it. New string
  added to every catalog at the same position.

Horizontal scrolling is untouched throughout: the setting reverses the wheel,
not the thumb wheel.

## Known scope

The flag is process-global, so it applies to every pointer the hook captures.
The hook attributes events to a device, but nothing maps an OS event source
back to the config key the setting is stored under, so a second Logitech mouse
would follow the first. Documented on the flag rather than left to be
discovered; per-device scoping needs that mapping to exist first.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings      # RUSTFLAGS=-D warnings
cargo test --workspace
cargo clippy --target x86_64-pc-windows-gnu -p openlogi-hook -p openlogi-agent-core \
  --all-targets -- -D warnings
cargo test -p openlogi-ui locale
```

All green. Four unit tests on the Linux transform: both vertical axes negated,
off is a byte-for-byte pass-through, horizontal and pointer axes untouched, and
`i32::MIN` saturates instead of wrapping back to the same sign.

**Verified on hardware — Linux, Signature M650 over Bluetooth.** That device is
the class in question: `openlogi diag features` lists `0x2130` and no `0x2121`,
and it advertises both `REL_WHEEL` and `REL_WHEEL_HI_RES`, so both axes the
transform handles are live on it. Running the two-phase example, the hook
grabbed the mouse, `OpenLogi virtual mouse` appeared, 620 wheel events were
captured across the two windows, and the scroll direction was normal in the
pass-through phase, reversed in the inverted phase, and normal again once the
grab was released. Horizontal events: zero, as expected — the M650 has no thumb
wheel.

**Not verified:** macOS. There is no macOS SDK on this host, so that backend is
neither run nor compiled here. The `core-graphics` setters it uses
(`set_double_value_field`, `set_integer_value_field`) were checked against the
vendored 0.25 source, and it negates exactly the fields `usable_scroll_delta`
reads, but it needs a real Mac and a `0x2130` mouse — the configuration #694 was
reported on. Also not run here: `tests (macos)`, `cargo-deny`, macOS clippy.

To check on macOS: connect an M650/M750, enable the toggle (it should now be
offered, labelled as the software fallback), and confirm the wheel reverses
while the trackpad keeps the system direction.


hamidrezahanafi:
Thanks for the PR, I built openlogi with your change and it's working perfectly with my MX Vertical!

tata-meow:
I went through the code and the approach is solid — negating at the native event level avoids the unit-conversion pitfalls of suppress+reinject, and scoping the macOS path to attributed non-trackpad sources is the right guard.

I have a Signature M650 on macOS + Bolt, which is exactly the configuration #694 was reported on and the macOS path you haven't been able to verify. Happy to build and test if you rebase past the merge conflicts — the locale files migrated from `.yml` to `.toml` with semantic keys in #1169, so that'll be the bulk of the conflict resolution.

One note on the process-global flag: you documented the limitation honestly, but it's worth noting that `scroll_source_may_intercept` already attributes scroll events to a device on macOS, so the per-device scoping might be closer than it looks — the event-to-config-key mapping is the missing piece rather than event attribution itself.

mingtsay:
Rebased this onto current master (`d5e2b880`) and verified the macOS path on hardware. Branch: https://github.com/mingtsay/OpenLogi/tree/feat/software-scroll-inversion

**Conflicts.** Three, all mechanical except the last:

- `openlogi-hook/src/lib.rs` — import block.
- `orchestrator.rs` — master extracted the hook-map write into `publish_hook_maps()`; `publish_scroll_inversion(key)` now sits next to that call.
- The locale catalogs — #1169 migrated `.yml` → `.toml` with semantic keys, so the new string moved to `pointer.scroll_direction_software_description`. All 22 of your translations carried over; #1169 also added `be.toml`, which needed a new one.

**One thing that wasn't wired up.** `scrolling_card()` still read `current_scroll_inversion_supported()` into `inversion_supported`, so on a `0x2130` device the toggle stayed disabled and the description stayed "does not report native HID++ scroll inversion support" — `current_scroll_inversion_available()` was never called from anywhere. Switched it to `available()`, which is what makes the three-way `inversion_description()` reachable. Also moved two doc comments that had ended up on the wrong functions (`scrolling_card`'s onto `inversion_description`, `run_tap_callback`'s onto `wheel_scroll_to_invert`).

**macOS, verified on hardware.** Signature M650 over a Logi Bolt receiver, macOS 26.6.2, Apple Silicon — the configuration #694 was reported on. `openlogi diag wheel` gives `device does not expose HID++ feature 0x2121`, and the device's persisted capabilities carry `scroll_inversion = false`, so this is the class in question. With the toggle on, the agent logs `accessibility granted — installing OS mouse hook` / `OS input hook installed`, the wheel reverses, and a paired Magic Trackpad on the same machine keeps the system scroll direction. Your `core-graphics` field negation is correct against a real hi-res wheel.

Worth noting for anyone testing: the macOS transform runs on the CGEventTap, so the agent needs **Accessibility**, not just Input Monitoring. On a config with no button remappings that is easy to miss — nothing else would have started the hook.

**Gates.** macOS (Apple Silicon): `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets` — all green. Linux (aarch64, 1.98.0, in Docker): the same plus `cargo test --workspace --exclude openlogi-desktop` and `cargo clippy --target x86_64-pc-windows-gnu -p openlogi-hook -p openlogi-agent-core --all-targets`. Not run: `cargo-deny`.

4ni1ak:
Adopted @mingtsay's rebase and fix — credited via a co-author trailer on the commit.

- Rebased onto current master (was 115 commits behind, includes the gpui-kit migration and #1169's `.yml` → `.toml` locale port).
- Fixed the real bug found: \`scrolling_card()\` was reading \`current_scroll_inversion_supported()\` instead of \`current_scroll_inversion_available()\`, so on a `0x2130` device the software-inversion toggle stayed disabled — the entire point of this PR was unreachable from the GUI.
- All translations carried through, including the new `be.toml` #1169 added.

Full local gate (fmt/clippy/test --workspace) green on the rebased tree. Verified on macOS hardware per mingtsay's report (Signature M650 over a Bolt receiver, the #694 configuration) — thanks for that, and for catching that Accessibility (not just Input Monitoring) is what the macOS transform needs since it runs on the CGEventTap.

yadav-prakhar:
@AprilNEA please have a look. Many devices will be unblocked by this.

Issue #293, opened 2026-06-20
[Bug]: MX Master 2S (Unifying Receiver) drops connection and shows as "Offline"


rickydebojeet:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.14

### Operating system

macOS

### OS version & architecture

macOS 27 (Golden Gate DB1, M1)

### Device model

MX Master 2s + Unifying receiver

### How is the device connected?

Other / not sure

### Affected area(s)

- [x] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

I have been using OpenLogi without any issues since Unifying support was initially added. However, after updating to the latest release (v0.6.14), my MX Master 2S is no longer being detected properly and is displayed as "Offline" in the GUI.	

If I try to add a new device then the device gets paired works for a couple of seconds and then goes offline.

I tried killing the agent and checked that the latest agent was working. The latest updates broke something I don't know what.

### Steps to reproduce

1. Open OpenLogi and observe the status of the MX Master 2S (shows as "Offline")
2. Attempt to add/pair the device as a new Unifying device.
3. The device pairs successfully and works normally for a few seconds.
4. After a few seconds, the device stops responding and reverts to "Offline" in the GUI.

### Diagnostics report

```markdown
- OpenLogi (GUI): v0.6.14 (release)
- Agent: v0.6.14 (connected)
- IPC protocol: GUI 3 / agent 3
- Inventory: ready
- OS: macOS 27.0 (arm64)
- Locale: en-IN (UI: follow system)
- Accessibility: granted · Input hook: installed
- Launch at login: yes · Menu bar: yes · Update check: on
- Running from: source build (dev)
- Config: schema 2 · 1 configured device(s) · thumbwheel 14

**Assets**
[...]

PR #853, opened 2026-08-23
fix(hid): improve Unifying receiver discovery reliability

v14:
## Summary

An unresponsive receiver can hold an inventory scan until several lower-level timeouts expire. A live Unifying arrival event can also be discarded when optional feature reads are incomplete, making a connected device appear offline.

Put a bounded deadline around receiver probes and preserve the liveness reported by a valid arrival event while optional details are filled in later.

## Changes

- Fail a receiver probe within the inventory budget.
- Retire a repeatedly unresponsive channel before reopening it.
- Keep live arrival state even when feature metadata is incomplete.
- Continue to reject malformed or unrelated receiver events.

## Testing

- `cargo test -p openlogi-device`
- Covered probe deadlines, retirement, cache fallback, and incomplete arrival data.

Refs #293.

### DIFF
--- a/crates/openlogi-device/src/inventory.rs
+++ b/crates/openlogi-device/src/inventory.rs
@@ -93,6 +93,15 @@ const RECEIVER_PROBE_BUDGET: Duration = Duration::from_secs(13);
 /// just lacks capabilities / battery until the next tick.
 const UNIFYING_SLOT_PROBE: Duration = Duration::from_millis(3500);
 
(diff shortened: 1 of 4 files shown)


AprilNEA:
Thanks — the direction here is right: the 0x41 arrival event is the correct source of truth, and trusting it is what #293 needed.

I rebased the branch onto current master (your two commits are preserved) and appended four commits on top。

AprilNEA (approved):

PR #862, opened 2026-08-23
feat(hid): allow DPI button gesture bindings

### LINKED ISSUE #806: [Feature]: allow the DPI / ModeShift button to carry the gesture role
### Pre-flight

Searched existing issues and the roadmap; I couldn't find this tracked. Raising it as a discussion rather than a PR, because it reverses what looks like a deliberate design decision — happy to be told it's deliberate and stays that way.

### Problem / motivation

The gesture role is currently restricted to two controls:

```rust
pub fn is_hidpp_gesture_source(self) -> bool {
    matches!(self, ButtonId::GestureButton | ButtonId::HapticPanel)
}
```

`GESTURE_SOURCE_BUTTONS` correspondingly lists only `0x00c3` and the haptic panel.

An MX Anywhere 3 has neither. Its `0x1b04` table is:

```
0x0050  0x0038  0x0401  analytics-events
0x0051  0x0039  0x0401  analytics-events
0x0052  0x003a  0x0531  divertable, raw-xy, analytics-events
0x0053  0x003c  0x0d31  divertable, raw-xy, analytics-events, raw-wheel
0x0056  0x003e  0x0d31  divertable, raw-xy, analytics-events, raw-wheel
0x00c4  0x009d  0x0531  divertable, raw-xy, analytics-events
```

So on this device the only gesture-capable route is the OS-hook path on Middle/Back/Forward. Middle click is a reasonable home for gestures, but it's also the natural home for `ToggleSmartShift` on a mouse with no dedicated gesture button, and you can't have both.

Worth noting `0x00c4` advertises exactly the same flags as middle click — `0x0531`, including `raw-xy`.

### Proposed solution

Let the DPI / ModeShift family carry the gesture role on devices that advertise raw-XY on it, so a mouse without a dedicated gesture button has somewhere to put gestures other than its middle button.

I've run this locally on 0.7.4 and it works — all four directions dispatch correctly from `0x00c4` on an MX Anywhere 3, so the control does stream raw-XY while held, not just advertise it. Three changes were needed:

1. `is_hidpp_gesture_source()` to include `ButtonId::DpiToggle`.
2. `GESTURE_SOURCE_BUTTONS` to map the `DPI_MODE_SHIFT_CIDS` to `ButtonId::DpiToggle`.
3. The DPI arming loop in `run_capture_session` to skip a CID already armed as a gesture source — otherwise it re-arms it with `raw_xy = false` and strips the raw-XY reporting, the same hazard the `divert_buttons` loop already guards against.

Graceful degradation seems to come for free: the gesture divert loop checks `supports_raw_xy()` first, so a device that doesn't advertise it never arms as a gesture source and keeps the existing plain-divert behaviour.

### Things I'm unsure about

- Whether other members of `DPI_MODE_SHIFT_CIDS` (`0x00ed`, `0x00fd`) behave the same way. I've only tested `0x00c4`.
- Whether losing the plain DPI-toggle press when that button is in gesture mode is an acceptable trade. It seemed reasonable to me — the Click direction can be bound to `CycleDpiPresets` — but it is a behaviour change for anyone who gestures on it.
- Whether there's a device-capability reason for the current restriction that isn't visible in the `0x1b04` flags, in which case this is the wrong approach.

Happy to open a PR if the direction is welcome. Tested on macOS 27 (arm64), MX Anywhere 3 over Bluetooth LE.

Related: #805 (unattributed `OtherMouse*` events on macOS), which had to be fixed first before gestures worked on this hardware at all.

nothingyuancando:
## Summary

Allow compatible DPI / ModeShift controls to carry the HID++ gesture role.

Fixes #806

## Changes

- Treat `DpiToggle` as a HID++ gesture source.
- Map all DPI / ModeShift control IDs to `DpiToggle`.
- Preserve raw-XY diversion when a DPI control is configured for gestures.
- Add a regression test for DPI gesture-source recognition.

## Testing

- `git diff --check`
- Rust tests not run: the environment has no `cargo`/`rustc`; rustup and Homebrew installation both stalled on network downloads.
- Not runtime-tested on hardware.

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -119,7 +119,9 @@ pub fn plan_for_device(
 #[cfg(test)]
 mod tests {
     use openlogi_core::binding::Binding;
-    use openlogi_hid::reprog_controls::{GESTURE_BUTTON_CID, HAPTIC_PANEL_CID};
+    use openlogi_hid::reprog_controls::{
+        DPI_MODE_SHIFT_CIDS, GESTURE_BUTTON_CID, HAPTIC_PANEL_CID,
+    };
 
     use super::*;
 
@@ -136,6 +138,24 @@ mod tests {
         assert!(GESTURE_SOURCE_BUTTONS.iter().any(|(_, button)| {
             *button == ButtonId::DpiToggle
         }));
+    }
+
+    #[test]
+    fn single_bound_dpi_toggle_uses_one_plain_capture_path() {
+        let mut cfg = Config::default();
+        cfg.set_binding(
+            "2b042",
+            ButtonId::DpiToggle,
+            Binding::Single(Action::Copy),
+        );
+
+        let plan = plan_for_device(&cfg, "2b042", route(), None, 0);
+        assert!(
+            plan.divert_buttons.iter().any(|&(cid, button)| {
+                DPI_MODE_SHIFT_CIDS.contains(&cid) && button == ButtonId::DpiToggle
+            }),
+            "a non-default DPI binding must use the generic button capture"
+        );
     }
 
     #[test]
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -28,7 +28,8 @@ pub struct DeviceCapturePlan {
     /// Per-button single actions for this device (per-app effective).
     pub bindings: BTreeMap<ButtonId, Action>,
     /// Per-direction map for each HID++ gesture source (the dedicated gesture
-    /// button, the MX Master 4 haptic panel) in gesture mode on this device,
+    /// button, the MX Master 4 haptic panel, or a compatible DPI / ModeShift
+    /// control) in gesture mode on this device,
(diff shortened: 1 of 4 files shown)


crates/openlogi-agent-core/src/capture_plan.rs
```diff
@@ -129,6 +132,32 @@ mod tests {
...
+            "2b042",
+            ButtonId::DpiToggle,
+            Binding::Single(Action::Copy),
+        );
+
+        let plan = plan_for_device(&cfg, "2b042", route(), None, 0);
+        assert!(
+            plan.divert_buttons.iter().any(|&(cid, button)| {
+                DPI_MODE_SHIFT_CIDS.contains(&cid) && button == ButtonId::DpiToggle
+            }),
```
AprilNEA, line 156: Could this regression test assert the single-delivery invariant rather than only checking that the generic path is present? This assertion would still pass if the dedicated DPI path were armed too—the exact double-dispatch regression this commit fixes. Please add coverage at the arming/handler boundary proving that one physical DPI CID press produces exactly one `ButtonPressed(DpiToggle)`.

AprilNEA:
Superseded by #1167, now merged. It includes DPI/ModeShift gestures, device capability checks, and regression coverage. Thanks for the contribution!

Issue #757, opened 2026-08-21
[Device]: Logitech G700S not detected


reyqn:
### Pre-flight checklist

- [x] I quit **Logi Options+** before testing.
- [x] I searched existing issues for my device.

### Device model

Logitech G700S

### Connection

Wired (USB cable)

### What's the problem?

Device is not detected at all

### `openlogi list` output

```text
2026-08-21T13:07:17.410093Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:07:17.410234Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:07:23.773185Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:07:23.773304Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
G513 RGB MECHANICAL GAMING KEYBOARD (—, vid=046d pid=c33c)
  └─ slot 255 ● G513 RGB MECHANICAL GAMING KEYBOARD (keyboard, wpid=?, battery=—)
          model_ids=[c33c,0000,0000] ext=05 serial=— unit_id=35394713 transports=usb
```

### `openlogi diag features` output

```text
2026-08-21T13:10:03.820683Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:10:03.820810Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:10:10.206561Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
2026-08-21T13:10:10.206668Z  WARN openlogi_hid::inventory: node probe keeps failing — retiring its channel before reopen
device: G513 RGB MECHANICAL GAMING KEYBOARD (direct 046d:c33c)
   idx      id  ver
     0  0x0000  v1
     1  0x0001  v1
     2  0x0003  v2
     3  0x4522  v0
     4  0x0005  v0
     5  0x1e00  v0
     6  0x4540  v0
     7  0x1eb0  v0
     8  0x8060  v0
     9  0x00c2  v0
    10  0x1801  v0
    11  0x1802  v0
    12  0x8080  v2
    13  0x8070  v7
    14  0x1821  v0
    15  0x1bc0  v0
    16  0x1806  v3
    17  0x40a2  v0
[...]

Issue #520, opened 2026-08-04
[Bug]: openlogi list shows my MX Master 3S device, but GUI doesn't have the device.


ArtixZ:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.23

### Operating system

macOS

### OS version & architecture

Tahoe 26.5.2 (25F84), apple ARM chip

### Device model

MX Master 3S

### How is the device connected?

Bluetooth (direct, no receiver)

### Affected area(s)

- [x] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

openlogi cli shows my device but GUI doesn't have trouble finding or displaying my device.

### Steps to reproduce

Just have MX Master 3S connected to my Mac via bluetooth. And then open openlogi GUI. 
No device is found in the GUI. But `openlogi list` shows this:
>openlogi list
MX Master 3S (—, vid=046d pid=b034)
  └─ slot 255 ● MX Master 3S (mouse, wpid=?, battery=85% full (discharging))
          model_ids=[b034,0000,0000] ext=02 serial=2306LZ513DJ8 unit_id=6489f029 transports=btle

### Diagnostics report

```markdown

```

### `openlogi list` output

```text
openlogi list

MX Master 3S (—, vid=046d pid=b034)
  └─ slot 255 ● MX Master 3S (mouse, wpid=?, battery=85% full (discharging))
          model_ids=[b034,0000,0000] ext=02 serial=2306LZ513DJ8 unit_id=6489f029 transports=btle
```

### Logs

```shell
OPENLOGI_LOG=debug openlogi list
2026-08-04T21:13:33.023461Z DEBUG openlogi_hid::transport: logitech HID node name=MX Master 3S pid=b034 usage_page=0x0001 usage_id=0x0002 matched=false
[...]

Issue #468, opened 2026-07-27
[Bug]: MX Anywhere 3S Bluetooth probe drops one feature response and is misclassified as receiver interface


garethng:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue). This appears related to #159, #218, #254, #322, and #348, but none documents this exact v0.6.22 Bluetooth-direct failure path.
- [x] I am on the latest release, v0.6.22 (`b6aaf468d2a7a84c2579955c9b9287c4b4c7dd21`).
- [x] Logi Options+ and Logi Options were fully uninstalled before testing. No Logitech control process was running.

### Which part of OpenLogi?

Both

### OpenLogi version

v0.6.22 release, exact tag/commit:

```text
v0.6.22
b6aaf468d2a7a84c2579955c9b9287c4b4c7dd21
```

### Operating system

macOS

### OS version & architecture

macOS 27.0, Apple Silicon (arm64)

### Device model

Logitech MX Anywhere 3S, VID `046d`, PID `b037`

### How is the device connected?

Bluetooth (direct, no receiver)

### Affected area(s)

- [x] Device discovery / detection
- [x] Button remapping
- [x] DPI control
- [x] SmartShift
- [x] Battery status
- [x] Other: scroll inversion and any setting that requires an online HID++ route

### What happened?

The MX Anywhere 3S is paired and works as a normal macOS mouse, but OpenLogi usually shows it as offline or reports zero detected devices. Settings such as reverse scrolling do not take effect because the agent never establishes a usable inventory/control route.

The device is supported and permissions are not the blocker:

- macOS exposes the expected Logitech HID++ long-report collection, usage page `0xff43`, usage `0x0202`.
- OpenLogi opens that interface successfully.
- Accessibility and Input Monitoring are granted to the signed nested `OpenLogiAgent.app`.
- Many HID++ requests return normally in about 15–30 ms.
- A standalone `openlogi list` trace eventually completed the same 35-entry feature walk and correctly identified the mouse, including battery and BTLE transport.

The failure occurs when any one request during the feature-table walk receives no response. The failed slot varies between attempts. `enumerate_features()` returns an error, `probe_features()` discards all features already read, and `probe_direct()` then classifies the Bluetooth-direct mouse as a likely receiver secondary interface:

```text
enumerate_features failed slot=255 error=Channel(Timeout)
[...]

Issue #375, opened 2026-07-12
[Bug]:


chernandezosma:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.19

### Operating system

Linux

### OS version & architecture

Kubuntu 26.04 LTS (Resolute Raccoon) x86_64

### Device model

Logitech G502 HERO (USB), Logitech G915 Wireless RGB Mechanical Keyboard (USB + Bluetooth + Logi Bolt)

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [ ] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

The OpenLogi GUI crashes within milliseconds of launch with a `RefCell already borrowed` panic in the GPUI X11 event loop. The window briefly appears and then disappears.

### Steps to reproduce

1. Install openlogi v0.6.19 via .deb package
2. Enable the systemd user agent: `systemctl --user enable --now openlogi-agent.service`
3. Launch the GUI: `openlogi-gui` or `openlogi`

### Diagnostics report

```markdown
2026-07-12T09:43:37.675738Z  INFO RPC{rpc.deadline=2026-07-12T09:43:47.675692622Z otel.kind="client" otel.name=Agent.protocol_version rpc.trace_id=00}: tarpc::client: SendRequest
2026-07-12T09:43:37.675844Z  INFO RPC{rpc.deadline=2026-07-12T09:43:47.675692622Z otel.kind="client" otel.name=Agent.protocol_version rpc.trace_id=00}: tarpc::client: ReceiveResponse
2026-07-12T09:43:37.675856Z  INFO RPC{rpc.deadline=2026-07-12T09:44:02.675852360Z otel.kind="client" otel.name=Agent.next_pairing rpc.trace_id=00}: tarpc::client: SendRequest
[...]

PR #888, opened 2026-08-24
fix(hid): key the native handle cache on usage pair, not device id alone

### LINKED ISSUE #880: [Bug]: MX Master 4 (BLE-direct) stuck offline — HID++ node probe fails indefinitely on macOS
### Environment
- OpenLogi 0.7.10 (macOS, signed build from today's release)
- macOS with Bluetooth LE HID (no Logi Bolt receiver — mouse paired directly)
- Device: MX Master 4, `direct:046d:b042:serial:2531zaz5cgx8`

### Symptom
Mouse shows "offline" in the GUI device card. `openlogi list` reports
"No Logitech HID++ devices or webcams found." indefinitely.

### What I've verified is NOT the cause
- macOS Bluetooth: mouse shows Connected, battery 100% (system_profiler SPBluetoothDataType)
- TCC permissions confirmed granted: Bluetooth, Input Monitoring, Accessibility
  (Input Monitoring toggle confirmed on in System Settings; agent log shows
  "accessibility granted — installing OS mouse hook")
- No competing official Logi Options+ process running
- Fully quit/relaunched app + `launchctl kickstart -k gui/$UID/org.openlogi.agent`
- Toggled `enabled = true` in config.toml for the device entry — no change
- Already on latest release (0.7.10)

### Debug log (RUST_LOG=debug)
Repeats indefinitely, every ~4s, never recovers:
```
WARN openlogi_device::inventory: node probe keeps failing — retiring its channel before reopen
```

### Config snippet
```toml
[devices."direct:046d:b042:serial:2531zaz5cgx8".identity.model_info.transports]
usb = false
equad = false
btle = true
bluetooth = false
```

Possibly related: #454, #218, #520, #468, #519, #757, #866, #375 — all show the
same "node probe keeps failing" warning across different devices/transports on macOS,
suggesting a shared root cause in the HID++ probe/handshake layer rather than
anything device- or transport-specific.

cristitm76:
## Bug

macOS enumerates one `async_hid::DeviceInfo` per HID **usage collection**, but a multi-collection device shares one `DeviceId::RegistryEntryId` across all of them. My MX Master 4 (BLE-direct, no Bolt/Unifying receiver) exposes 4 collections on this id:

- `{usage_page: 0x0001, usage_id: 0x0002}` — generic desktop / mouse
- `{usage_page: 0x0001, usage_id: 0x0001}` — generic desktop / pointer
- `{usage_page: 0xff43, usage_id: 0x0202}` — vendor HID++ (this is the one we want)
- `{usage_page: 0x000e, usage_id: 0x0001}` — haptics

`NativeBackend::refresh()` (`crates/openlogi-hid/src/transport/native.rs`) cached handles keyed only on `node_id()` = `format!("{:?}", info.id)`, so all 4 collections collapsed onto one `HashMap` entry — last-enumerated wins.

`enumerate_hidpp()` correctly filters and returns the `0xff43/0x0202` `NodeInfo`. But `handle()` looks the OS handle back up by id alone, so it can hand back a *different* collection's handle than the one that was actually enumerated. `open_hidpp_channel()` then computes `long_only` from whichever collection `handle()` returned — on my machine that was the haptics collection, `is_long_only_collection(0x000e, 0x0001)` is `false`, so the channel decided the device supports HID++ *short* reports. It doesn't (BLE-direct is long-only), so the `hidpp` crate skipped up-converting outgoing short messages, and every write went out as report `0x10` — a report id that doesn't exist in this device's HID++ collection. macOS rejects it locally with `kIOReturnNotFound` (`0xE00002F0`) before it ever reaches the radio.

**Symptoms:** the vendor collection opens fine (so `opened HID++ channel` logs), but every read/write after that fails immediately, the feature walk never completes, and the inventory watcher logs `node probe keeps failing — retiring its channel before reopen` in a tight retry loop forever. `openlogi list` / the GUI report no devices found even though the peripheral shows connected at the OS level. This matches #880 exactly, and I suspect at least some of #757, #520, #468, #375 as well, since "node probe keeps failing" on macOS is a symptom of at least two distinct root causes (this one, and the separate known `kIOReturnTimeout` / async-hid#45 issue) currently sharing one log message.

## Fix

Cache handles under a `NodeKey { id, usage_page, usage_id }` instead of `NodeId` alone, so each collection gets its own cache entry and `handle()` returns the one the caller actually asked for.

## Testing

Verified against a Logitech MX Master 4 (VID `046d`, PID `b042`) paired Bluetooth-direct on macOS (Tahoe):

**Before:**
```
WARN openlogi_device::inventory: node probe keeps failing — retiring its channel before reopen
```
repeating indefinitely, `openlogi list` → `No Logitech HID++ devices or webcams found.`

**After:**
```
$ openlogi list
MX Master 4 (—, vid=046d pid=b042)
  └─ slot 255 ● MX Master 4 (mouse, wpid=?, battery=100% full (discharging))
          model_ids=[b042,0000,0000] ext=00 serial=2531ZAZ5CGX8 unit_id=3c28ac40 transports=btle

$ openlogi diag battery
device: MX Master 4 (direct 046d:b042)
  0x1004 UnifiedBattery: percentage=100 level=Full status=Discharging

$ openlogi diag smartshift
✓ SmartShift round-trip OK

$ openlogi diag dpi
✓ DPI round-trip OK
```

Full read/write/readback/restore round trips all pass. Action Ring, gesture routing, and haptic panel button handling also confirmed working end to end with the built `openlogi-overlay` alongside this fix.

Fixes #880

### DIFF
--- a/crates/openlogi-hid/src/transport/native.rs
+++ b/crates/openlogi-hid/src/transport/native.rs
@@ -32,11 +32,30 @@ pub(crate) fn native_backend() -> Arc<dyn HidBackend> {
     Arc::clone(&NATIVE_BACKEND) as Arc<dyn HidBackend>
 }
 
+/// Cache key for one enumerated HID node: its reported id plus the usage pair
+/// that distinguishes the collections sharing that id. See [`NativeBackend`].
+#[derive(Clone, PartialEq, Eq, Hash)]
+struct NodeKey {
+    id: NodeId,
+    usage_page: u16,
+    usage_id: u16,
+}
+
+impl NodeKey {
+    fn of(info: &NodeInfo) -> Self {
+        Self {
+            id: info.id.clone(),
+            usage_page: info.usage_page,
+            usage_id: info.usage_id,
+        }
+    }
+}
+
 /// [`HidBackend`] over `async-hid`.
 #[derive(Default)]
 pub(crate) struct NativeBackend {
     /// OS handles from the most recent enumeration, keyed by the id that
-    /// enumeration reported them under.
+    /// enumeration reported them under *plus its usage pair*.
     ///
     /// `async_hid::Device` is an OS handle, not a value: it cannot be rebuilt
     /// from a [`NodeId`], and re-finding one costs another enumeration. Since
@@ -44,7 +63,22 @@ pub(crate) struct NativeBackend {
     /// the handles from that enumeration is both cheaper and a truer model
     /// than looking them up again. Held behind an `Arc` so an open can borrow
     /// one without keeping the map locked across its await.
-    nodes: Mutex<HashMap<NodeId, Arc<Device>>>,
+    ///
+    /// The usage pair belongs in the key because macOS reports one
+    /// `DeviceInfo` per usage pair while giving them all the *same*
+    /// `DeviceId::RegistryEntryId`. Keying on the id alone collapses every
+    /// collection of a multi-collection device onto one entry and the last
+    /// enumerated one wins, so `handle()` can hand back a different collection
+    /// than the caller enumerated. That mis-sourced `usage_page`/`usage_id`
+    /// then flows into `is_long_only_collection()` in `open_hidpp_channel`,
+    /// which decides whether the channel advertises HID++ *short* report
+    /// support. A BLE-direct mouse (e.g. MX Master 4) is long-only, but if the
+    /// generic-desktop or haptics collection wins the key it is judged
+    /// short-capable, the `hidpp` channel skips up-converting short messages,
+    /// and every write goes out as report `0x10` — a report id that does not
+    /// exist on that device — which macOS rejects with `kIOReturnNotFound`
+    /// (0xE00002F0) before it ever reaches the radio.
+    nodes: Mutex<HashMap<NodeKey, Arc<Device>>>,
 }
 
 impl NativeBackend {
@@ -57,7 +91,7 @@ impl NativeBackend {
             .collect();
         let handles = devices
             .iter()
-            .map(|device| (super::node_id(device), Arc::clone(device)))
+            .map(|device| (NodeKey::of(&super::node_info(device)), Arc::clone(device)))
             .collect();
         *self.nodes.lock().unwrap_or_else(PoisonError::into_inner) = handles;
         Ok(devices)
@@ -68,7 +102,7 @@ impl NativeBackend {
         self.nodes
             .lock()
             .unwrap_or_else(PoisonError::into_inner)
-            .get(&node.id)
+            .get(&NodeKey::of(node))
             .map(Arc::clone)
             .ok_or(BackendError::Disconnected)
     }


cristitm76:
Update: confirmed working after a full macOS reboot, not just in the dev session where I tested it originally.

To be fully transparent about how this was verified — I copied the four patched binaries (`openlogi`, `openlogi-desktop`, and the `openlogi-agent`/`openlogi-overlay` binaries inside their respective LoginItems bundles) directly into the installed `/Applications/OpenLogi.app`, re-signed ad-hoc (`codesign --force --deep --sign -`) so the bundle's signature stays internally consistent, keeping every `CFBundleIdentifier` unchanged (`org.openlogi.openlogi`, `org.openlogi.agent`, `org.openlogi.overlay`). That preserved the existing LaunchAgents/login items, so nothing needed re-registering with launchd.

The only side effect: switching from the original Developer ID signature to an ad-hoc one changes the code identity, so macOS required Input Monitoring to be re-granted to `OpenLogi Agent.app` (Accessibility was already granted at the bundle-ID level and survived). One-time step, done through System Settings → Privacy & Security → Input Monitoring → Go to Folder → the agent's `.app` path (the "+" picker can't browse into a nested LoginItems bundle directly, worth noting for docs/troubleshooting since I hit that same friction earlier in the session for an unrelated reason).

After reboot, with no manual intervention:
```
$ openlogi list
MX Master 4 (—, vid=046d pid=b042)
  └─ slot 255 ● MX Master 4 (mouse, wpid=?, battery=100% full (discharging))
          model_ids=[b042,0000,0000] ext=00 serial=2531ZAZ5CGX8 unit_id=3c28ac40 transports=btle

$ openlogi diag battery
device: MX Master 4 (direct 046d:b042)
  0x1004 UnifiedBattery: percentage=100 level=Full status=Discharging

$ openlogi diag smartshift
✓ SmartShift round-trip OK

$ openlogi diag dpi
✓ DPI round-trip OK
```

Agent process confirmed launched automatically via its login item (`org.openlogi.agent`), not manually started. Action Ring, haptic panel gesture routing, and button bindings all also working end to end (overlay renders and tracks hover via the `observe_action_ring`/`action_ring_hover` RPCs).

Happy to answer any questions about the diagnosis or provide more logs if useful — this was blocking BLE-direct MX Master 4 use entirely on macOS, so glad to have nailed it down.

AprilNEA:
Thanks for the excellent diagnosis and hardware validation, @cristitm76.

#882 has now merged the same usage-pair-aware handle-cache fix, including regression coverage for multiple HID collections sharing one OS node. Since the functional change is now on `master`, this PR is being closed as superseded.

Your contribution remains valuable: it established the exact wrong-collection → wrong report width → `0xE00002F0` failure chain, confirmed that #880 was distinct from #798, and validated the fix extensively on the affected MX Master 4 across reboot, discovery, battery, DPI, SmartShift, Action Ring, haptics, and button routing.

The fix is not included in v0.7.10 and will ship in the next release. Thanks again for documenting this so thoroughly.

Issue #651, opened 2026-08-17
[Feature]: MX Ergo S wheel tilt left/right remapping support


JohndoeZhao:
### Pre-flight checklist

- [x] I searched existing issues and the [Roadmap](https://github.com/AprilNEA/OpenLogi#roadmap), and this isn't already tracked.

### Problem / motivation

The Logitech MX Ergo S has a scroll wheel that can be tilted left and right.

Currently, OpenLogi does not expose wheel tilt-left and wheel tilt-right as independently configurable inputs. As a result, these physical controls cannot be assigned to custom actions.

This limits button remapping compared with Logitech Options+, where the left and right wheel tilt actions can be used for functions such as navigation or other shortcuts.

### Proposed solution

Detect and expose the MX Ergo S scroll wheel tilt directions as two separate button inputs:

Wheel Tilt Left
Wheel Tilt Right

Both inputs should appear in OpenLogi's button remapping options and support the same configurable actions as other remappable mouse buttons.

For example:

Tilt left → Back
Tilt right → Forward

or:

Tilt left → Previous tab
Tilt right → Next tab

The two inputs should be independently configurable rather than being limited to horizontal scrolling.

### Alternatives considered

Using horizontal scrolling as the default behavior partially uses the wheel tilt hardware, but it does not allow the two directions to be customized.

External remapping tools could also be used, but this would require running another utility alongside OpenLogi and would defeat the purpose of having device button configuration handled directly by OpenLogi.

Related issues include #366 for MX Ergo S unsupported features and #100 for similar wheel tilt detection/remapping behavior on the Logitech M500S.

### Related area(s)

- [ ] GUI
- [ ] CLI
- [x] Button actions / remapping
- [ ] DPI
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

JohndoeZhao:
> [OLG-221](https://linear.app/aprilnea/issue/OLG-221)

May I know what this is ?

PR #902, opened 2026-08-24
feat(core,gui): make the main wheel's tilt rebindable (MX anywhere 2s,...)

### LINKED ISSUE #583: [Bug]: Wheel tilt-left/right on MX Anywhere 2S not detected
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.6.26

### Operating system

Windows

### OS version & architecture

Windows 11 26.1000

### Device model

MX Anywhere 2S

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

Wheel tilt not detected and thus not configurable.

### Steps to reproduce

Connect mouse, open OpelLogi.

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.6.26 (release)
- Agent: v0.6.26 (connected)
- IPC protocol: GUI 13 / agent 13
- Inventory: ready
- OS: Windows (x86_64)
- Locale: cs-CZ (UI: follow system)
- Launch at login: yes · Menu bar: yes · Update check: off
- Config: schema 3 · 1 configured device(s) · thumbwheel 14

**Assets**
- Cache path: C:\Users\yoga\.local/share/openlogi\assets · Bundle assets: absent

**Devices (1)**
- MX Anywhere 2S — mouse (codename: Wireless Mobile Mouse MX Anywhere 2S)
  - Connection: Bluetooth (direct) · Online: yes · Battery: 90% (discharging, full)
  - Model: 0b01a (model-ids: b01a/406a/0000, ext-model: 00)
  - Transports: eQuad, BTLE
  - Render: mx_anywhere_2s · direct
```

### `openlogi list` output

```text

```

### Logs

```shell

```

6lv1-blr:
## Summary

Fixes #583, reproduced on the exact setup reported there: an MX Anywhere 2S
connected Bluetooth-direct on Windows 11, where the wheel tilt is neither
detected nor configurable.

Mice whose main wheel tilts left/right expose each direction as its own HID++
reprogrammable control, and Options+ lets users rebind both. OpenLogi showed
neither: the buttons panel had no hotspot for them, so there was nothing to
click.

Two independent gaps caused that:

- `ButtonId` had no variant for either tilt, so no layer could name one.
- `map_slot_name` did not know Logitech's slot names for them
  (`SLOT_NAME_LEFT_SCROLL_BUTTON` / `SLOT_NAME_RIGHT_SCROLL_BUTTON`), even
  though the device metadata OpenLogi already downloads ships a marker for
  each — placed beside the wheel, where the control physically is.

Verified against real hardware before writing any code. An MX Anywhere 2S
reports both controls as divertable:

```
$ openlogi diag controls
device: Wireless Mobile Mouse MX Anywhere 2S (direct 046d:b01a)
  0x0052  0x00a9  0x0131  divertable, raw-xy
  0x0053  0x003c  0x0131  divertable, raw-xy
  0x0056  0x003e  0x0131  divertable, raw-xy
  0x005b  0x003f  0x0131  divertable, raw-xy   <- tilt left
  0x005d  0x0040  0x0131  divertable, raw-xy   <- tilt right
```

HID++ diversion is the only workable path. A tilt reaches the host as a
horizontal scroll, not as a button, so capturing it in the OS hook would also
swallow a trackpad's horizontal scroll — the device-level divert affects only
the mouse the user rebound.

**An untouched tilt stays native.** Each side is seeded with the horizontal
scroll its firmware already performs, and the capture plan diverts a control
only when its binding leaves the default — so a user who never opens the
panel keeps firmware horizontal scrolling exactly as before.

## Changes

**`openlogi-core`**
- `ButtonId::WheelTiltLeft` / `WheelTiltRight`, appended after `HapticPanel`
  to keep the serialized form append-only, with labels `Tilt Left` /
  `Tilt Right`.
- Defaults are `HorizontalScrollLeft` / `HorizontalScrollRight` — matching the
  firmware behaviour is what keeps an unbound tilt out of the divert set.

**`openlogi-device`**
- CIDs `0x005b` / `0x005d` added to `DIVERTABLE_STANDARD_BUTTONS`. Arming
  already filters on what a device's own `getCtrlIdInfo` reports, so the
  entries are inert on a wheel that does not tilt, and press dispatch needed
  no change — the rising-edge path is per-CID and generic.

**`openlogi-desktop`**
- `map_slot_name` maps the two tilt slots, giving each a hotspot at the
  coordinates Logi publishes. Neither is an OS-hook button nor a HID++
  gesture source, so the picker offers them as plain buttons with no
  swipe-mode entry.

**`openlogi-ui`**
- `Tilt Left` / `Tilt Right` added to all 21 locale catalogues, at the same
  position, with per-language translations.

**Tests**
- `wheel_tilt_defaults_to_the_scroll_its_firmware_already_does` — pins the
  default on both sides, since a different seed would silently divert an
  untouched tilt.
- `bound_wheel_tilt_is_diverted_but_an_untouched_one_stays_native` — the
  capture-plan half of that contract.
- `wheel_tilt_slots_map_to_their_own_controls` — the slot names, which is what
  was missing.

No wire change: `ButtonId` does not cross the IPC boundary (bindings are read
from `config.toml` by the agent, not shipped over tarpc), so
`PROTOCOL_VERSION` is unchanged. The wire-format goldens are green as-is.

## Testing

Run on Windows 11, `x86_64-pc-windows-msvc`, rustc 1.98.0, on the final tree:

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass |
| `cargo test --workspace` | pass |
| `cargo test -p openlogi-ipc --test wire_format` | pass (13 tests) |
| `cargo test -p openlogi-desktop i18n` | pass |
| `cargo xtask release check-publish` | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent` | fails on `openlogi-permissions` only — see below |
| `cargo build --release -p openlogi-desktop -p openlogi-overlay -p openlogi-agent -p openlogi --target x86_64-pc-windows-msvc` | pass |

**Not run on this host** (`cargo xtask ci` reported 3 passed, 6 skipped): `shell`
(no shellcheck/shfmt), `MSRV (cargo check)` (linux/macos only), `tests (linux)`,
`tests (macos)`, `cargo-deny` (not installed), `wasm (portable crates)` (target
not installed). None of these are claimed green.

**Two pre-existing Windows-only failures, unrelated to this diff:**

1. The rustdoc gate fails on `openlogi-permissions`: its crate-level docs link
   to `PermissionStatus::Unknown`, a type declared
   `#[cfg(any(target_os = "macos", target_os = "linux"))]`. On Windows the type
   does not exist, so the link cannot resolve. That crate is untouched here and
   has no dependencies at all on Windows (both dependency blocks are cfg-gated
   away), so this diff cannot reach it. Excluding it, the gate passes. CI's
   rustdoc job runs on Linux and never sees this.
2. `cargo xtask ci` fails its own `publish closure` job on Windows: the job
   shells out to `cargo xtask release check-publish`, whose nested cargo tries
   to relink the running `xtask.exe`, which Windows locks. Run standalone, the
   same job passes (see the table above).

**Hardware verification: partial.** The control-ID dump above is from a real
MX Anywhere 2S, and the release build runs with the agent seeing the device.
Binding a tilt to an action and confirming it fires has **not** been
runtime-tested end to end. Screenshots of the two new hotspots are still to be
added.

Fixes #583

Nothing here keys off a model list: a tilt appears when the device itself
reports the CIDs as divertable and its metadata ships the two slots. So this
should also cover the MX Anywhere 2 reported in #583's thread, and likely #651
(MX Ergo S) and #100 (M500S) — none of those three were verified, no hardware
for any of them. A `openlogi diag controls` dump from one of those mice would
settle it.

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -155,6 +155,35 @@ mod tests {
     }
 
     #[test]
+    fn bound_wheel_tilt_is_diverted_but_an_untouched_one_stays_native() {
+        // The main wheel's tilt scrolls horizontally in firmware, so the
+        // default binding must leave it native — diverting an untouched tilt
+        // would silently kill horizontal scrolling. Binding one side to a real
+        // action is what arms its `0x1b04` divert.
+        let mut cfg = Config::default();
+        cfg.set_binding(
+            "2b01a",
+            ButtonId::WheelTiltLeft,
+            Binding::Single(Action::PrevTab),
+        );
+
+        let plan = plan_for_device(&cfg, "2b01a", route(), None, 0);
+        assert!(
+            plan.divert_buttons
+                .contains(&(0x005b, ButtonId::WheelTiltLeft)),
+            "a bound tilt must be diverted, or the binding can never fire: {:?}",
+            plan.divert_buttons
+        );
+        assert!(
+            !plan
+                .divert_buttons
+                .iter()
+                .any(|&(_, button)| button == ButtonId::WheelTiltRight),
+            "the untouched right tilt must keep its native horizontal scroll"
+        );
+    }
+
+    #[test]
     fn haptic_panel_gestures_when_promoted() {
         // The MX Master 4 haptic panel is a HID++ gesture source: promoting it
         // into gesture mode must arm the raw-XY gesture divert, exactly like
--- a/crates/openlogi-core/src/binding/button.rs
+++ b/crates/openlogi-core/src/binding/button.rs
@@ -64,19 +64,31 @@ pub enum ButtonId {
     /// (Logi metadata slot `ASSIGNMENT_NAME_SHOW_RADIAL_MENU`, HID++ CID
     /// `0x01a0`). A separate physical control from [`ButtonId::GestureButton`];
     /// captured over HID++ like it, and eligible as the gesture owner.
+    HapticPanel,
+    /// Tilting the main wheel left — `0x1b04` CID `0x005b` ("Left Scroll"),
+    /// Logi metadata slot `SLOT_NAME_LEFT_SCROLL_BUTTON`. A distinct control
+    /// from the thumb wheel: it is a plain divertable button, not a rotation,
+    /// and it lives on the main wheel of mice like the MX Anywhere 2S.
+    WheelTiltLeft,
+    /// Tilting the main wheel right — `0x1b04` CID `0x005d` ("Right Scroll"),
+    /// Logi metadata slot `SLOT_NAME_RIGHT_SCROLL_BUTTON`. Counterpart to
+    /// [`ButtonId::WheelTiltLeft`].
+    ///
     /// Declared last: the TOML config and any serialized form encode the
     /// variant identifier / index, so new buttons are append-only.
-    HapticPanel,
+    WheelTiltRight,
 }
 
 impl ButtonId {
     /// Every rebindable button in declaration (physical front-to-side) order —
     /// the iteration source for default-binding seeding and the popover
     /// trigger list.
-    pub const ALL: [ButtonId; 11] = [
+    pub const ALL: [ButtonId; 13] = [
         ButtonId::LeftClick,
         ButtonId::RightClick,
         ButtonId::MiddleClick,
+        ButtonId::WheelTiltLeft,
+        ButtonId::WheelTiltRight,
         ButtonId::Back,
         ButtonId::Forward,
         ButtonId::DpiToggle,
@@ -135,6 +147,8 @@ impl ButtonId {
             ButtonId::LeftClick => "Left Click",
             ButtonId::RightClick => "Right Click",
             ButtonId::MiddleClick => "Middle Click",
+            ButtonId::WheelTiltLeft => "Tilt Left",
+            ButtonId::WheelTiltRight => "Tilt Right",
             ButtonId::Back => "Back",
             ButtonId::Forward => "Forward",
             ButtonId::DpiToggle => "DPI Toggle",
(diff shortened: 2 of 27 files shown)


AprilNEA (approved):
LGTM. The implementation looks correct for the MX Anywhere 2S.

As a small follow-up, we could also support [MX Ergo S metadata](https://cdn.jsdelivr.net/npm/@logi-assets/pointing@0.1.0/v1/devices/mx_ergo_s/metadata.json) and MX Ergo. They use `SLOT_NAME_SCROLL_LEFT` / `SLOT_NAME_SCROLL_RIGHT` for the same `c91` / `c93` controls. Adding these as aliases in `map_slot_name`, with a test for both naming variants, would extend this fix to those models as well.

ilkondr:
looks like this should work for the M590 as well. 
Here’s the output I get from `openlogi diag controls`:
```
device: M585/M590 Multi-Device Mouse (direct 046d:b01b)
  0x0050  0x0038  0x0011  -
  0x0051  0x0039  0x0011  -
  0x0052  0x003a  0x0171  divertable, raw-xy
  0x0053  0x003c  0x0171  divertable, raw-xy
  0x0056  0x003e  0x0171  divertable, raw-xy
  0x005b  0x003f  0x0171  divertable, raw-xy
  0x005d  0x0040  0x0171  divertable, raw-xy
  ```

filipstojakovic:
on MacOS this side scroll is not configurable on MX 2S

PR #910, opened 2026-08-24
fix(mouse): compose thumb-wheel preset labels from translated action names

NatsUIJM:
## Summary
Split out of #764 at review request (@AprilNEA): the preset-label localization fix travels alone so the zoom feature PR stays focused.

The thumb-wheel preset rows rendered flat `"X / Y"` strings that are keys in **no** locale file — every preset row fell back to English regardless of app language. This composes each row label from its two sides' already-translated action names instead:

- per-action names carry translations for all locales, so composition reuses that reviewed coverage instead of adding ~21 locales × 11 presets of new keys;
- `CycleDpi` fires one action in both directions and renders as a single undivided name;
- callers (`picker.rs`, `view.rs`) drop their now-redundant `tr!` wrapping since `label()` returns a localized `String`.

## Testing
- new unit test: two-action presets contain the " / " separator; CycleDpi equals the single translated action name with no separator (locale-independent shape assertions);
- `cargo clippy -p openlogi-desktop --all-targets` clean; all 12 thumbwheel-related desktop tests pass locally.

#764 will rebase on top of this once it lands (its preset entries then need no label() change at all — `pair()` + composition cover them).

### DIFF
--- a/crates/openlogi-desktop/src/features/mouse/picker.rs
+++ b/crates/openlogi-desktop/src/features/mouse/picker.rs
@@ -127,7 +127,7 @@ pub(crate) fn thumbwheel_picker<T: 'static>(
         .enumerate()
         .map(|(idx, preset)| {
             let selected = current == Some(preset);
-            let label = tr!(preset.label());
+            let label = preset.label();
             let observer = observer.clone();
             let popover = popover.clone();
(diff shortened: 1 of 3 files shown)


AprilNEA (approved):
LGTM, thanks!

PR #944, opened 2026-08-24
fix(agent): pace thumb-wheel action repeats from the sensitivity slider

### LINKED ISSUE #887: [Bug]: Thumbwheel set to control volume ignores sensitivity
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.7.10

### Operating system

Windows

### OS version & architecture

11, x86_64

### Device model

MX Master 3s

Logi Bolt receiver

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

I have set the thumbwheel to control the volume, and no matter how much I set the sensitivity of the thumbwheel later, the volume only goes up or down by 2. I would like to at least reduce/increase a quarter of the volume in just one swipe and I expected that I could adjust that with that setting.

### Steps to reproduce

1. Bind the thumbwheel to Volume Up / Down
2. Go to "Pointer" settings tab and adjust the sensitivity

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

4ni1ak:
## Summary

A thumb wheel bound to Volume Up/Down moved two steps per swipe no matter where
the sensitivity slider sat.

Fixes #887

## Root cause

The slider does reach the action path — `action_threshold()` drops as it rises —
but it is not the binding constraint:

```rust
pub fn action_threshold(self) -> i32 {
    (2 * i32::from(Self::DEFAULT) - i32::from(self)).max(1)   // DEFAULT = 14
}
```

That floors at one increment once the slider passes 27 of its 1–100 range, so
the whole top ~70% asks for the same threshold. Every fire was then gated by a
fixed 200 ms cooldown, capping a swipe at five fires a second whatever the
setting said — and a ~400 ms swipe at exactly the two steps the report
describes.

## Change

Scale the cooldown by the same slider, next to the threshold it complements:

```rust
pub fn action_cooldown(self) -> Duration
```

- `DEFAULT` → **200 ms**, the value it always had, so an untouched install is
  bit-for-bit unchanged.
- Bottom of the range → 400 ms: still deliberate, one flick one fire.
- Top of the range → 25 ms: a sustained spin repeats.

The floor keeps a fast spin from flooding the action queue; the ceiling keeps
the low end deliberate rather than unusable.

No locale changes: the GUI copy already promised this behaviour — *"Scales the
thumb wheel's horizontal scroll speed and how readily custom wheel actions
trigger."*

## Note on one-shot actions

Above the default this also lets a non-repeating action (App Exposé, say)
re-fire within one long spin. That is the slider doing what it says rather than
a regression, and the default is untouched — but if you would rather the
cooldown only relax for repeatable actions, that needs a "repeatable" notion on
`Action` that does not exist yet, and I would rather raise it than invent one
here.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo test -p openlogi-agent-core watchers::gesture
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings      # RUSTFLAGS=-D warnings
cargo test --workspace
cargo xtask ci wasm
```

All green, including the existing `custom_action_fires_on_threshold_then_
respects_cooldown`, which pins the default path and needed no change.

Four new tests:
- `raising_sensitivity_shortens_the_gap_between_repeats` — and that the top of
  the range shares one floored threshold, so the cooldown is the only thing left
  to move.
- `the_default_keeps_its_historical_gap` — pins 200 ms.
- `a_sustained_spin_repeats_at_max_sensitivity`
- `the_same_spin_stays_paced_at_the_default`

The repeat test was checked against the old fixed cooldown: it fires **3** times
per half-second there versus ≥10 after, so it pins the defect rather than the
implementation.

**Not runtime-tested on hardware.** No thumb-wheel device here — the mouse on
this desk is a Signature M650, which has none — so the timing was exercised
through `advance()` with a synthetic clock rather than a real spin. To confirm:
bind the thumb wheel to Volume Up/Down, push the sensitivity slider to the top,
and one swipe should move the volume well past two steps; at the default it
should behave exactly as it does today.

**Not run on this host:** `tests (macos)`, `cargo-deny`, macOS clippy.


OutshineIssue:
Any update on this?

4ni1ak:
Still open and green — CI passes, no unresolved review comments. Ready for a maintainer look whenever there's time.

ozkansoyturk:
Tested on hardware: MX Master 3 (Unifying), Windows 11, thumb wheel on Volume Up/Down. Cherry-picked 630781fb onto today's master (applies cleanly, `cargo test -p openlogi-agent-core wheel` passes). Volume now follows a fast spin smoothly at sensitivity 100, default 14 behaves as before. Works for me.

Note: the branch as-is is ~155 commits behind master and rejects a current config (`unsupported schema_version 7`), so it needs a rebase before merging

4ni1ak:
Thanks for the hardware verification! Already rebased onto current master since then (`e939efc2`, current head) — no longer 155 commits behind, applies cleanly against today's schema.

Issue #730, opened 2026-08-20
[Device]: G502 X PLUS (wireless+USB) cant rebind buttons


AsterionXeity:
### Pre-flight checklist

- [x] I quit **Logi Options+** before testing.
- [x] I searched existing issues for my device.

### Device model

G502 X PLUS Wireless Gaming Mouse

### Connection

Unifying receiver

### What's the problem?

Device is detected but a feature doesn't work

### `openlogi list` output

```text
Lightspeed Receiver (89FC3921, vid=046d pid=c547)
  └─ slot 1 ● G502 X PLUS (mouse, wpid=4099, battery=98% full (discharging))
          model_ids=[4099,c095,0000] ext=00 serial=— unit_id=21bed667 transports=usb+equad
```

### `openlogi diag features` output

```text
(UNIFYING)
device: G502 X PLUS (slot 1 on receiver 89FC3921)
   idx      id  ver
     0  0x0000  v2
     1  0x0001  v2
     2  0x0003  v4
     3  0x0005  v0
     4  0x1d4b  v0
     5  0x0020  v0
     6  0x1004  v3
     7  0x2201  v2
     8  0x2121  v0
     9  0x8071  v2
    10  0x8081  v2
    11  0x8100  v0
    12  0x8110  v0
    13  0x8060  v0
    14  0x1500  v0
    15  0x00d0  v3
    16  0x1802  v0
    17  0x1803  v0
    18  0x1806  v8
    19  0x1811  v0
    20  0x1805  v0
    21  0x1830  v0
    22  0x1875  v0
    23  0x1861  v1
    24  0x1890  v7
    25  0x18a1  v0
    26  0x1801  v0
    27  0x1e00  v0
    28  0x1e22  v0
    29  0x1eb0  v0
    30  0x18b1  v0
    31  0x18c0  v0
  (32 feature entries)

(USB WIRED)
device: G502 X PLUS (direct 046d:c095)
   idx      id  ver
     0  0x0000  v2
     1  0x0001  v2
     2  0x0003  v4
     3  0x0005  v0
     4  0x1d4b  v0
     5  0x0020  v0
     6  0x1004  v3
     7  0x2201  v2
     8  0x2121  v0
     9  0x8071  v2
    10  0x8081  v2
    11  0x8100  v0
    12  0x8110  v0
    13  0x8060  v0
    14  0x1500  v0
    15  0x00c2  v0
    16  0x00d0  v3
    17  0x1802  v0
[...]

Issue #392, opened 2026-07-14
[Device]: Logitech G502 Hero SE


eGOoOGle:
### Pre-flight checklist

- [x] I quit **Logi Options+** before testing.
- [x] I searched existing issues for my device.

### Device model

Logitech G502 Hero SE

### Connection

Wired (USB cable)

### What's the problem?

Device is detected but a feature doesn't work

### `openlogi list` output

```text
Unable to configure the mouse like others
```

### `openlogi diag features` output

```text

```

exico91:
I have the same problem but with the G502 Hero (no SE).
I can adjust the DPI and RGB but I cannot bind any button

pxlnvdr:
Not issue related but does it connect via bolt or wired only?

exico91:
Mine is a G502 Hero, is wire only, not detachable. (maybe is an SE after all after reading the lsusb output)

Some infos from lsusb, maybe are useful:

```
 lsusb -v -d 046d:c08b

Bus 005 Device 005: ID 046d:c08b Logitech, Inc. G502 SE HERO Gaming Mouse
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x046d Logitech, Inc.
  idProduct          0xc08b G502 SE HERO Gaming Mouse
  bcdDevice           27.03
  iManufacturer           1 Logitech
  iProduct                2 G502 HERO Gaming Mouse
  iSerial                 3 1660396F3931
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x003b
    bNumInterfaces          2
    bConfigurationValue     1
    iConfiguration          4 U127.03_B0010           
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower              300mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      1 Boot Interface Subclass
      bInterfaceProtocol      2 Mouse
      iInterface              0 
        HID Device Descriptor:
          bLength                 9
[...]

Issue #886, opened 2026-08-24
[Feature]: Support Logitech gaming devices — become a full G Hub alternative


erek303:
### Pre-flight checklist

- [x] I searched existing issues and the [Roadmap](https://github.com/AprilNEA/OpenLogi#roadmap), and this isn't already tracked.

### Problem / motivation

OpenLogi is a great local-first replacement for **Logi Options+**, but Logitech's *gaming* peripherals are managed by a separate app: **G Hub**. Users who own both productivity and gaming gear (e.g., an MX Master for work plus a G502 X / G Pro X Superlight 2 for gaming) still have to keep G Hub installed alongside OpenLogi.

That partially defeats OpenLogi's core promise: G Hub requires a Logitech account for some features, runs a heavy Electron app plus multiple background services, and pushes cloud profile sync — exactly the things OpenLogi frees us from on the Options+ side.

Receiver-level detection for several Lightspeed receivers already works (#512), but the feature surface on top is still Options+-oriented, so gaming mice are recognized without exposing their gaming capabilities.

### Proposed solution

Extend OpenLogi's gaming-device feature set so it can fully replace G Hub, the same way it replaces Options+. Roughly in priority order:

1. **Gaming mouse basics** — remapping for all G-buttons (including G-Shift), multi-stage DPI (per-stage DPI value, e.g. 5 stages, with optional on-device color indication), and report/polling rate switching.
2. **Onboard memory profiles** — read/write on-device profiles so settings persist on other machines without any daemon (on top of #459).
3. **LIGHTSYNC RGB** — per-zone colors and effects, per-profile (on top of #351).
4. **Remaining Lightspeed receivers** — tracked in #512.
5. Later: G-series keyboards (G-keys/macros) and headsets.

It would also help to have a single "Gaming devices" roadmap/tracking item so the scattered gaming-related PRs and issues (#351, #369, #459, #730, #392, #877) have one home.

### Alternatives considered

- **G Hub itself** — what we're trying to avoid (account, telemetry, resource usage).
- **Logitech Onboard Memory Manager** — Windows-only, limited device coverage, no longer actively extended.
- **piper/ratbag (Linux)** — partial gaming-mouse support (DPI, buttons), Linux-only, no per-app profiles or richer features.

### Related area(s)

- [ ] GUI
- [ ] CLI
- [x] Button actions / remapping
- [x] DPI
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Configuration (TOML)
[...]

PR #949, opened 2026-08-24
fix(core,gui,ipc,assets): give G-series mice a working Buttons panel

### LINKED ISSUE #877: [Bug]: G309 LIGHTSPEED buttons tab not showing
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.7.10

### Operating system

Windows

### OS version & architecture

11 25H2

### Device model

G309 + LIGHTSPEED receiver

Other / not sure

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

The G309 buttons tab only shows on new installs and then when turning the mouse on it's gone and doesn't come back. Also the mouse doesn't show a picture in the UI.

### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.7.10 (release)
- Agent: v0.7.10 (connected)
- IPC protocol: GUI 26 / agent 26
- Inventory: ready
- OS: Windows (x86_64)
- Config: schema 4 · 3 configured device(s) · thumbwheel 14

**Assets**
- Cache path: C:\Users\InterClaw\.local/share/openlogi\assets · Bundle assets: absent

**Devices (3)**
- G309 — mouse (codename: G309)
  - Connection: Logi Unifying receiver · Online: yes · Battery: 100% (discharging, full)
  - Capabilities: buttons=no, pointer=yes, lighting=no
  - DPI: 400 dpi (range 100–25600, 383 steps)
  - Model: 0b03c (wpid: 40b1, model-ids: b03c/40b1/0000, ext-model: 00)
  - Transports: eQuad, BTLE
  - Render: ⚠️ none (silhouette) · Slot 1
- MX Keys S — keyboard (codename: MX KEYS S)
  - Connection: Logi Bolt receiver · Online: yes · Battery: 55% (charging, full)
  - Capabilities: buttons=yes, pointer=no, lighting=no
  - Model: 0b378 (wpid: b378, model-ids: b378/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_keys_s · Slot 1
- M585/M590 — mouse (codename: M585/M590)
  - Connection: Logi Unifying receiver · Online: yes · Battery: 90% (discharging, full)
  - Capabilities: buttons=yes, pointer=no, lighting=no
  - Model: 5b01b (wpid: 406b, model-ids: b01b/406b/0000, ext-model: 05)
  - Transports: eQuad, BTLE
  - Render: m585 · Slot 1

**Receivers (3)**
- Logi Bolt Receiver (VID 046d / PID c548)
- Lightspeed Receiver (VID 046d / PID c547)
- Unifying Receiver (VID 046d / PID c52b)

Image

Image

### Steps to reproduce

1. Plug in the USB dongle.
2. Install OpenLogi.
3. Turn on the mouse.

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

yuzi-co:
## Summary

G-series mice get no Buttons panel at all, because they expose no ReprogControls.
A G502 LIGHTSPEED reports `0x8100` OnboardProfiles and `0x8110` MouseButtonFilter
and nothing in the `0x1b0x` family — verified on hardware over both the Lightspeed
receiver and USB. `Capabilities::buttons` therefore read false, `tabs_for` withheld
the Buttons and Actions Ring tabs, and the mouse had no way to bind anything from
the GUI, even though its middle/back/forward are ordinary HID buttons the input
hook already remaps host-side.

Two commits: the capability model, then the asset-depot parsing that makes the
resulting panel actually usable.

**Capabilities.** `buttons` now answers "does a Buttons panel appear" and covers
the gaming control tables as well as ReprogControls. A new `no_button_diversion`
answers "what goes in it" — without ReprogControls nothing beyond the OS-visible
three can be diverted, so the model filters its hotspots down to those and drops
the orphaned labels with them.

The new flag is stored as a *restriction* rather than a capability so
`#[serde(default)]` lands on the old behaviour by itself. `Capabilities` is
persisted to `config.toml` and the probe cache as well as sent over IPC, and
builds before this change set `buttons` only for ReprogControls devices — exactly
the ones that are *not* hook-only. Phrased the other way round, the first launch
after upgrading would have read every stored MX identity as hook-only and stripped
its DPI, gesture and thumb-wheel hotspots. No config migration or schema bump is
needed as a result.

The Keys panel needs the opposite question and now asks it: the function-row
remapper has no OS-hook path, it diverts every key over `0x1b04`. A gaming
keyboard reports `buttons` from the same tables a G502 does, so gating Keys on
`buttons` would open a panel whose capture session bails on the missing feature
and re-arms roughly once a second.

**Asset depots.** With the panel restored it rendered zero clickable targets,
because gaming depots speak a different dialect than the Options+ depots the mouse
model was written against, differing in three ways at once: no
`device_buttons_image` entry (the side render is keyed `device_side`), no
`slotName` on any assignment (only a `slotId` like `g502wireless_g4_m1`), and
markers in absolute origin pixels rather than the 0..100 percentage the existing
translation divides by. Any one of those alone empties the hotspot list.

`_g<N>_` maps for N in 1..=5 — the five standard HID mouse buttons in order. The
G502's own artwork confirms the ordering: its side render has "G5" printed on the
forward thumb button and "G4" on the rear one. G6 and up are model-specific and
stay unmapped rather than guessed; they are handled in firmware, emit no HID event
for the hook to intercept, and would need `0x8100` OnboardProfiles writes to
rebind. An absent `slotName` is what tells the two marker conventions apart, so
the percentage path is untouched for every depot that already worked — no
magnitude heuristic involved.

### Known limitations

- **Middle click is not reachable on these mice.** Its marker lives on the front
  render and the panel draws one view, so a G-series mouse gets back/forward only.
  Worth a follow-up (a two-view panel); it affects any depot that splits markers
  across `device_image` and `device_side`.
- **G6+ remain unconfigurable**, and not because of this change: those controls
  are handled in firmware and reach neither the hook nor `0x1b04`. Supporting them
  means implementing `0x8100` OnboardProfiles, which OpenLogi carries only as a
  registry name today.

## Changes

**openlogi-core**
- `Capabilities::buttons` covers `0x8100`/`0x8110` alongside `0x1b00`–`0x1b04`.
- New `Capabilities::no_button_diversion` plus a `can_divert_buttons()` accessor.
- `diag` names the restriction when it applies — the first thing to check on a
  "my extra button does nothing" report.

**openlogi-ipc**
- `PROTOCOL_VERSION` 31 → 32; inventory golden regenerated.

**openlogi-assets**
- `Assignment::slot_id` is now deserialized.
- `assignments()` falls back to the `device_side` entry; new `buttons_origin()`
  returns that same entry's origin — a gaming depot's views differ in width (the
  G502's front render is 1391, its side 936), so measuring against the first entry
  scales every hotspot off its control.

**openlogi-desktop**
- Buttons/Actions Ring tabs gate on `buttons`; Keys gates on `can_divert_buttons()`.
- Hotspot filtering on both the asset and synthetic paths, with orphaned labels
  dropped alongside.
- `map_slot_id` for depots that ship no `slotName`, and marker normalization for
  the absolute-pixel convention.
- `buttons_image_for` resolves `device_side` when `device_buttons_image` is absent.

**openlogi-fixture / openlogi-agent**
- The shared canonical fixture profile gains a synthetic gaming mouse on Bolt
  slot 5 (G-series capabilities, no `0x1b0x` features), so the mock agent reaches
  this path without G-series hardware. Fixture replay/record tests adjusted for
  the extra device.

## Testing

Rebased onto master at `b9c8fede` and re-run on a Windows host:

```
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings     # RUSTFLAGS=-D warnings
cargo test --workspace --no-fail-fast
```

fmt and clippy are clean. `cargo test --workspace` has one failure, and it is
neither new nor this branch's: xtask's `ci_yml_runs_what_this_runner_runs`, which
reports that `ci.yml` and the xtask job table have drifted on the rustdoc job's
flags. It fails identically on a docs-only branch cut from the same master, and
this diff touches no file under `xtask/` or `.github/`.

The wire golden passes. Master had moved to v31, so this branch's bump was
renumbered to v32.

Not run on this host, and named as not run rather than claimed green: the Linux
and macOS CI lanes, MSRV, cargo-deny, the wasm portability job, and rustdoc —
`openlogi-permissions` has a `PermissionStatus` doc link that dangles on Windows
because the type is macOS/Linux-gated, which is what #970 fixes.

**Hardware verification** — tested on a real G502 LIGHTSPEED over the Lightspeed
receiver:

- `openlogi diag features` confirms the premise: `0x8100`, `0x8110`, `0x2201`,
  `0x2121`, `0x8070`, `0x1001`, and nothing in the `0x1b0x` family.
- `openlogi diag controls` fails with "device does not expose HID++ feature
  `0x1b04`", as expected.
- The agent's capture session reports
  `gesture_sources=0 dpi_buttons=0 buttons=0 thumbwheel=false` — measured live,
  matching what `from_feature_ids` derives, with no error and no re-arm loop.
- The Buttons panel renders the depot's side view with hotspots landing on the
  buttons physically labelled G4 and G5 in Logitech's own artwork.

Not verified on hardware: the G-series *keyboard* Keys-tab gate (no G-series
keyboard available), and the stale-config path on a real upgrade — that one is
covered by a unit test asserting a capability table with no `no_button_diversion`
key still reads as divertable.

The hardware runs above predate the rebase. Only the checks under Testing were
re-run on `44b19925`.

image

The reviewer can reach this path without hardware via
`cargo run -p openlogi-agent --bin openlogi-agent-mock`, which now serves a synthetic gaming mouse on Bolt slot 5.

Fixes #877
Addresses #392 and #730 in part — those devices gain a working Buttons panel with
back/forward, but not the full button set, which needs OnboardProfiles support.

### DIFF
--- a/crates/openlogi-agent/src/bin/mock_agent.rs
+++ b/crates/openlogi-agent/src/bin/mock_agent.rs
@@ -19,9 +19,10 @@
 //! Scripted behavior:
 //!
 //! - A Bolt receiver with an online mouse (DPI + SmartShift + battery that
-//!   drains ~1%/minute), an offline mouse, and a lighting-capable keyboard,
-//!   plus one directly-attached mouse — covering every panel and both route
-//!   kinds without hardware.
+//!   drains ~1%/minute), an offline mouse, a lighting-capable keyboard and a
+//!   G-series mouse (gaming control tables, no ReprogControls), plus one
+//!   directly-attached mouse — covering every panel and both route kinds
+//!   without hardware.
 //! - A standalone Litra light whose power / brightness / temperature writes
 //!   persist, and a `camera_active` flag that flips every 30s so the
 //!   camera-linked light rendering has something to follow.
@@ -73,6 +74,10 @@ const RECEIVER_UID: &str = "MOCK-BOLT-01";
 const MOUSE_SLOT: u8 = 1;
 const OFFLINE_SLOT: u8 = 2;
 const KEYBOARD_SLOT: u8 = 3;
+/// A G-series mouse: the gaming control tables and no ReprogControls, so it
+/// gets a Buttons panel drawn down to what the OS hook can see. Scripted here
+/// because that path is otherwise unreachable without G-series hardware.
+const GAMING_MOUSE_SLOT: u8 = 4;
 const MOCK_TORQUE: TunableTorque = match TunableTorque::try_new(50) {
     Ok(value) => value,
     Err(_) => panic!("valid mock SmartShift torque"),
@@ -312,6 +317,17 @@ impl State {
             },
         );
         settings.insert(
+            GAMING_MOUSE_SLOT,
+            DeviceSettings {
+                dpi: Some(DpiState {
+                    current: Dpi::new(3200),
+                    capabilities: DpiCapabilities::new((100u16..=25600).step_by(100).collect())?,
+                }),
+                smartshift: None,
+                lighting: false,
+            },
+        );
+        settings.insert(
             DIRECT_DEVICE_INDEX,
             DeviceSettings {
                 dpi: Some(DpiState {
@@ -324,7 +340,7 @@ impl State {
         );
         Ok(Self {
             paired_extra: Vec::new(),
-            next_slot: KEYBOARD_SLOT + 1,
+            next_slot: GAMING_MOUSE_SLOT + 1,
             settings,
             pairing: None,
             phase: None,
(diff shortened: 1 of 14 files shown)


yuzi-co:
Rebased onto master (`8c80c6e`). One conflict, in the wire contract: #960 landed `Action::HoldShortcut` as **v28**, which this branch had also claimed. Renumbered this branch to **v29** (`Capabilities::no_button_diversion`), bumped `PROTOCOL_VERSION`, and moved the `protocol_version_is_pinned` literal with it so the bump stays visible in the same diff, as that test intends.

`cargo test -p openlogi-ipc --test wire_format` is green (14/14) — the goldens themselves needed no regeneration, only the pin.

yuzi-co:
Pushed `193ff847`. The capability fixtures are real dumps now rather than invented
ones, from a G705 and a G713 reported on #886.

- `gaming_keyboard_gets_no_keys_tab` used `[0x8100, 0x8110, 0x8070, 0x1001]`. A real
  G713 carries `0x8100` **without** `0x8110`, which is the stricter case for a gate
  that reads either table. It still derives `buttons: true, no_button_diversion:
  true`, so the Keys panel stays hidden — proved on hardware now instead of assumed.
- New `a_second_gaming_mouse_derives_the_same_way`: a G705 over a Lightspeed
  receiver lists `0x8100`/`0x8110` and nothing in the `0x1b0x` family, and
  `openlogi diag controls` fails against it exactly as it does on a G502. The same
  derivation on a second device.

Tests only, no behaviour change. On Windows: `cargo fmt --all -- --check` and
`cargo clippy --workspace --all-targets -- -D warnings` clean, and
`cargo test --workspace --no-fail-fast` green on every target except the
pre-existing xtask `ci_yml_runs_what_this_runner_runs` drift.

Unrelated to this PR but worth recording somewhere: the G713 also exposes `0x8010`,
which the feature registry names `GamingGKeys` and nothing implements. A gaming
keyboard's G-keys are unreachable today by any path, and a remapper for them would
target `0x8010` rather than `0x1b04` — so the Keys gate here stays correct either
way.

Yaphet2015:
Really need this.Thanks for your work

PR #957, opened 2026-08-25
feat(linux): ship a Flatpak bundle with each release

### LINKED ISSUE #371: [Feature]: Flatpak distribution / Publish on Flathub
### Pre-flight checklist

### Problem / motivation

OpenLogi is currently unavailable on atomic distributions (Silverblue, Bluefin, Bazite, secureblue, etc.) which are the future of Linux desktop.

The current way to install OpenLogi is not straightforward for average desktop users (at least for Fedora and its derivatives).

### Proposed solution

Please distribute this great application as a flatpak, the universal Linux format and publish it on Flathub to drastically improve "discoverability" and ease of installation for average desktop users.

### Alternatives considered

_No response_

### Related area(s)

- [x] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [ ] Configuration (TOML)
- [x] Auto-update
- [x] Other

### Additional context

_No response_

AalmanSadath:
## Summary

Adds a `.flatpak` bundle per architecture to the release assets, beside the
`.deb`/`.rpm`/`.pkg.tar.zst` already there. One new job in the existing build
matrix, and four small edits to `release.yml` to attach what it produces.

It deliberately stops there. A bundle is a one-time install: there is no remote
behind it, so `flatpak update` has nothing to check. Closing that needs a
repository to host, sign and keep alive, which is a bigger decision than adding
a build artifact, #767 proposes exactly that, and this PR is the subset that
does not commit you to any of it.

## Changes

**`packaging/linux/flatpak/`** (new)

- `org.openlogi.OpenLogi.yml` — freedesktop 25.08 runtime, `llvm20` SDK
  extension (openlogi-camera pulls `v4l2-sys-mit`, whose build script runs
  bindgen and dlopens libclang; the base SDK ships none), and rustup rather than
  `Sdk.Extension.rust-stable`, because the extension trails `rust-version` and
  cargo refuses outright when it does. Source is `type: dir` of the checkout, so
  a release builds exactly the tagged tree.
- `org.openlogi.OpenLogi.metainfo.xml` — AppStream, required for software
  centres to list the app. There is none in the tree today.
- `org.openlogi.OpenLogi.desktop` — Flatpak requires the entry to be named for
  the application ID, which `packaging/linux/desktop/openlogi.desktop` is not.

Permissions worth reviewing: `--device=all` is the only one covering both
`/dev/hidraw*` and `/dev/uinput` — `--device=input` covers neither. It grants a
path to those nodes, not DAC permission; the host still needs
`70-openlogi.rules`, exactly as the `.deb` does. `--share=network` is for the
device render images. Solaar carries the same `--device=all` on Flathub for the
same reason.

**`.github/workflows/build.yml`**

- New `linux-flatpak` job, matrix `amd64`/`arm64` on the same runners as
  `linux-packages`, wired into `results` and the `workflow_call` outputs.
- Separate from `linux-packages` on purpose: it compiles the workspace again
  inside the Flatpak SDK, where the runner's rust-cache cannot reach, so folding
  them together would lose the `.deb` whenever the Flatpak build failed.
- AppStream is stamped from the tree — version from the workspace, date from the
  commit — never from an input, so a rebuild of an old tag cannot claim to have
  been released the day it was rebuilt.

**`.github/workflows/release.yml`**

Downloads `OpenLogi-linux-flatpak-*`, adds `*.flatpak` to `SHA256SUMS`, adds
`flatpak` to the best-effort detection loop, and adds the conditional entry to
softprops' `files`. It rides the existing degradation path: `publish` gates only
on `macos_result`, so a failed Flatpak leg costs the bundle, not the release.

**`README.md`** — the Flatpak bundle in the Linux section, with the host udev
setup as one command.

## Testing

Dispatched `Release` on the branch (publish skipped, `ref_type != 'tag'`):

- `linux-flatpak (amd64)` — pass
- `linux-flatpak (arm64)` — pass
- `linux-packages` — pass; macOS/Windows legs fail in a fork for want of signing
  secrets, which is expected and unrelated

Locally, from the repository root:

```sh
flatpak-builder --user --disable-rofiles-fuse --force-clean \
  --repo=repo --default-branch=stable \
  build packaging/linux/flatpak/org.openlogi.OpenLogi.yml
flatpak build-bundle --runtime-repo=https://dl.flathub.org/repo/flathub.flatpakrepo \
  repo out.flatpak org.openlogi.OpenLogi stable
appstreamcli validate packaging/linux/flatpak/org.openlogi.OpenLogi.metainfo.xml
desktop-file-validate packaging/linux/flatpak/org.openlogi.OpenLogi.desktop
```

Both exit 0 in ~10 minutes; metainfo and desktop validate. Re-imported the
bundle and confirmed the contents: all four binaries, the stamped version,
`70-openlogi.rules`, every icon size, and the runtime repo embedded.

Installed the amd64 CI artifact on Fedora 44 — `flatpak list` shows it at
0.7.10.  device access depends on the host udev rules, which that machine already had.

The `release.yml` half is not exercised by any of the above, because `publish`
needs a tag *and* a successful macOS leg, which a fork cannot produce. I ran its
shell against a fabricated `dist/` instead: with bundles present both land in
`SHA256SUMS` and `flatpak=true`; with the leg failed, checksums still generate
and `flatpak=false`, so `fail_on_unmatched_files` does not trip.

## Notes for review

- **CI will not build this on the PR.** `should-build` returns `false` for
  `pull_request` without the `needs: build` label, so a green check here has not
  run the new job. Worth labelling before judging it.
- `publish` has `needs: build`, so a release now waits for the Flatpak legs even
  though it does not depend on them. Measured ~15 minutes cold on
  `ubuntu-latest`; `timeout-minutes: 45` is a hang bound, not an estimate.
- The bundle's origin ends up named after the file, so the release filename is
  user-visible in `flatpak list`, not only on the release page.

Fix #371

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,9 @@ on:
       linux_result:
         description: Result of the Linux packages build job.
         value: ${{ jobs.results.outputs.linux }}
+      linux_flatpak_result:
+        description: Result of the Linux Flatpak bundle build job.
+        value: ${{ jobs.results.outputs.linux_flatpak }}
   # PR builds are opt-in: add the `needs: build` label to produce unsigned
   # installers for testing. `closed` / unlabel cancel an in-flight run via
   # concurrency (jobs below no-op on those actions).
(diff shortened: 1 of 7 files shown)


davidbudnick:
Duplicate of #767 closing.

Issue #556, opened 2026-08-10
[Feature]: KWin/Plasma native-Wayland frontmost-window backend


recchia:
### Pre-flight checklist

- [x] I searched existing issues and the Roadmap, and this isn't already tracked.

### Problem / motivation

#191 added Wayland-native frontmost-window backends for wlroots compositors (`zwlr_foreign_toplevel_management_v1`) and GNOME Shell (companion D-Bus extension), with X11/XWayland as the universal fallback. KDE Plasma/KWin has neither: it isn't wlroots-based, and there's no KWin-side backend equivalent to the GNOME extension. So `detect_session_kind()` picks Wayland, both Wayland candidates fail to initialize, and `frontmost_bundle_id()` falls through to the X11 backend — which only sees XWayland windows. Native-Wayland KDE apps (anything not running under XWayland) report `None`, so per-app profiles silently don't fire for them.

This matches real user reports: #355 was filed by a KDE Plasma/Wayland user (separate touchpad-grab bug, being fixed in #356), and #191's own compositor-coverage table lists "KDE/KWin, others" as X11/XWayland-fallback-only, a known gap at merge time.

### Proposed solution

Add a `FrontmostSource` candidate for KWin, following the same pattern as the two existing Wayland backends in `crates/openlogi-hook/src/linux.rs`. KWin has no public Wayland protocol for this either, so the practical options are:

- A KWin script (using the KWin scripting API's `workspace.activeWindow`) that exports the focused window's resource class over a D-Bus service, mirroring the GNOME Shell extension's `org.openlogi.Frontmost` approach.
- Or investigate whether KWin's existing `org.kde.KWin` D-Bus interface / `kwin_wayland` already exposes enough (e.g. via `queryWindowInfo` or similar) to avoid shipping a companion script at all.

Whichever backend lands, it should return KWin's native window identifier and slot into `wayland_candidates()` alongside the wlroots and GNOME backends, falling through to X11/XWayland exactly as today when it can't initialize.

### Alternatives considered

- Leave KDE on the X11/XWayland fallback — this is the current behavior, but means per-app profiles don't work for native-Wayland apps on the second-most-common Linux desktop after GNOME.

### Related area(s)

- [x] Per-application profiles
- [x] Other (Linux/KDE Wayland support)

### Additional context

- Wayland backend dispatch: `crates/openlogi-hook/src/linux.rs` (`wayland_candidates()`, `FrontmostSource`)
[...]

PR #998, opened 2026-08-25
feat(linux): ship the GNOME Shell companion extension in the packages

### LINKED ISSUE #555: [Feature]: Package the GNOME Shell frontmost-window extension via nfpm
### Pre-flight checklist

- [x] I searched existing issues and the Roadmap, and this isn't already tracked.

### Problem / motivation

#191 added a GNOME Shell companion extension (`crates/openlogi-hook/gnome-shell-extension/openlogi-frontmost@openlogi.dev/`) so per-app profiles can work on GNOME/Wayland, where Mutter offers no protocol/portal for the focused window's `WM_CLASS`. Today the extension isn't part of any Linux package — `packaging/linux/nfpm.yaml` has no `contents:` entry for it. GNOME/Wayland users have to manually copy the extension files into `~/.local/share/gnome-shell/extensions/` and log out/in, per the README in that directory. That's a rough first-run experience and easy to miss.

This was flagged as an open question in #191 itself: extension distribution (bundle-and-document vs. extensions.gnome.org vs. auto-install) was left unresolved, with a note that #173/#179 (nfpm packaging) — the blocker for the bundle-and-document approach — have since merged.

### Proposed solution

Add the extension to `packaging/linux/nfpm.yaml` as a `contents:` entry, installed to `/usr/share/gnome-shell/extensions/openlogi-frontmost@openlogi.dev/` (matching the system-wide GNOME extension install path), so `.deb`/`.rpm`/`.pkg.tar.zst` installs ship it automatically. Update `crates/openlogi-hook/gnome-shell-extension/README.md` and the Linux install docs accordingly — users would still need to `gnome-extensions enable` and log out/in (Wayland can't hot-reload the shell), but wouldn't need to manually copy files first.

Publishing to extensions.gnome.org is a larger, separate effort (review process, versioning independent of OpenLogi releases) and can stay a follow-up.

### Alternatives considered

- Auto-install from the app itself (GUI or agent writes the extension files on first run) — more magic, and still requires a logout/login plus manual `gnome-extensions enable`, so the packaging approach covers most of the friction with less complexity.
- Publish to extensions.gnome.org — desirable long-term but has its own review/versioning process independent of this packaging fix.

### Related area(s)

- [x] Per-application profiles
- [x] Other (Linux packaging)

### Additional context

- Extension source: `crates/openlogi-hook/gnome-shell-extension/openlogi-frontmost@openlogi.dev/`
- Current manual install docs: `crates/openlogi-hook/gnome-shell-extension/README.md`
- Packaging descriptor: `packaging/linux/nfpm.yaml`
- Originating PR: #191

4ni1ak:
## Summary

The GNOME Shell companion extension was in the tree but in no package, so
per-app profiles on GNOME/Wayland began with the user finding
`crates/openlogi-hook/gnome-shell-extension/`, copying two files into
`~/.local/share/gnome-shell/extensions/`, and logging back in — before any of it
worked.

Fixes #555

## Changes

Installed to `/usr/share/gnome-shell/extensions/openlogi-frontmost@openlogi.dev/`
from the same four places the icon goes, so no packaging path is left behind:

- `packaging/linux/nfpm.yaml` — `.deb` / `.rpm` / `.pkg.tar.zst`
- `packaging/linux/package.nix` — the derivation, plus two `test -f` lines in
  its install check
- `packaging/linux/install.sh` — with the enable instructions printed after
- `packaging/linux/uninstall.sh` — removes the directory

**Installed, not enabled.** GNOME loads only extensions the user opted into, and
Wayland cannot reload the shell in place, so `gnome-extensions enable` and a new
session stay the user's two steps. What the packages remove is the copying.

**Docs.** The extension's README no longer opens with the manual copy — it says
the packages already place the files and keeps the copy for source builds.
`docs/INSTALL-linux.md` gains a short section covering the same, and says
plainly that nothing else depends on it: button remaps, DPI and SmartShift work
without the extension, and X11/XWayland sessions resolve the focused window with
no extension at all.

Publishing to extensions.gnome.org stays a follow-up, as the issue says — its
review and versioning run independently of OpenLogi's releases.

### Since the first review

- The uninstaller no longer runs `rm -rf` on the extension directory. It removes
  the two files the installer wrote and drops the directory only when nothing
  else is left in it, so an extension installed there by a distro package
  survives an OpenLogi uninstall.
- Adopted @hikarisakamoto's follow-up commit, authorship intact: the nfpm
  post-install script now prints the same `gnome-extensions enable` hint
  `install.sh` already prints. They verified the packaging side with nfpm
  2.46.3 — the files land in all three formats and the hint reaches the pacman
  `.INSTALL`.

**Not run here:** the `shell` (shellcheck + shfmt) and `typos` CI jobs — neither
tool nor a Nix shell is available on this machine, so the scripts were only
syntax-checked with `bash -n`. CI covers both.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo fmt --all -- --check
cargo test -p xtask                                     # 59 passed
bash -n packaging/linux/install.sh packaging/linux/uninstall.sh
```

All green, including `packaged_bins_cover_every_binary_nfpm_installs`, which
this leaves alone — the new entries are data, not binaries.

The nfpm manifest was also parsed and every `src` resolved against the working
tree: the only paths that do not exist are the three `target/release/` binaries
CI builds before packaging, exactly as before this change. Both new entries
resolve, at mode `0644`.

**Not verified:** no package was built or installed here — `nfpm` is not on this
host, and neither is `nix`, so the derivation edit is reviewed rather than
evaluated. To confirm: build the package, install it, and check
`/usr/share/gnome-shell/extensions/openlogi-frontmost@openlogi.dev/` holds both
files.

**Not verified either:** the end-to-end GNOME path. This desk runs KDE Plasma on
Wayland, where the extension is neither loaded nor needed (#556 tracks the KWin
gap separately), so what is tested here is that the files ship — not that
enabling them makes per-app profiles fire. That part is unchanged by this PR.

**Not run on this host:** `tests (macos)`, `cargo-deny`, macOS clippy, and the
`shell` job — shellcheck and shfmt are not installed, so the two shell diffs
were reviewed by hand and pass `bash -n`; both new blocks are 2-space indented
per `.editorconfig` with every expansion quoted.


hikarisakamoto:
Nice work — this was the missing piece for per-app profiles on GNOME Wayland. I rebased the branch onto current master and added one small follow-up commit on top: the deb/rpm/pacman post-install script now prints the gnome-extensions enable hint, matching what install.sh already prints (no uninstall-side disable — a root script can't touch per-user dconf, and GNOME tolerates an enabled-but-missing UUID).
The result is at [hikarisakamoto/OpenLogi:adopt/pr-998-gnome-extension](https://github.com/hikarisakamoto/OpenLogi/tree/adopt/pr-998-gnome-extension) with your commit and authorship intact — I don't have write access to update this PR directly, so a maintainer can push that onto this branch (edit access is enabled), or @4ni1ak  can pull it in.
Verified with nfpm 2.46.3: the extension files land in all three package formats, and the hint shows up in the pacman .INSTALL.

4ni1ak:
Thanks for double-checking this! Looks like your print-hint commit is already in this branch — same author, same patch (`557d0a56`), just carried forward through a couple of rebases since. This branch also has one more commit beyond that: `6789cd35`, scoping the uninstaller's `rm -rf` down to only the two files the installer wrote, so it doesn't delete extension files another install method left behind. Nothing to pull in from your branch that isn't already here — appreciate you verifying with nfpm though, good to have that confirmed independently.

4ni1ak:
Fixed the package-ownership finding, pushed `1efb546f`. `uninstall.sh` now checks `dpkg -S` / `rpm -qf` / `pacman -Qo` before removing either extension file — if a package owns it, it's left in place for that package's own removal instead of getting deleted out from under the package manager's record.

Also rebased onto current master while I was in there (no content change beyond that).

Not run through `shellcheck`/`shfmt` locally — neither is on this machine outside the devenv shell. `sh -n` syntax-checks clean; will watch the CI `shell` job.

4ni1ak:
Valid follow-up, fixed in `9a41e357`. `install.sh` now runs the same package-ownership check before writing either extension file — if a distro package already owns the path, the standalone installer leaves it alone instead of overwriting it. Symmetric with `uninstall.sh`'s check now.

PR #1004, opened 2026-08-25
fix(linux): open the Actions Ring at the cursor on Wayland

### LINKED ISSUE #774: [Bug]: The Actions Ring always opens in the center of the desktop and not at the current cursor position
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.7.4

### Operating system

Linux

### OS version & architecture

Ubuntu 26.04 LTS, GNOME Version 50, Wayland

### Device model

MX Master 3S + Logi Bolt receiver

Logi Bolt receiver

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

Image

### Steps to reproduce

1. Move the cursor from the center of the desktop
2. Click the button mapped to the "Actions Ring" built-in action, the Actions Ring always opens in the center of the desktop

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

AalmanSadath:
## Summary

On Wayland the Actions Ring opened in the centre of the screen instead of under
the pointer. Both halves of the existing placement are unavailable to an
ordinary Wayland client:

- **The cursor cannot be read.** No protocol reports the global pointer position,
  so `openlogi_hook::cursor_position()` — X11 only — has no Wayland answer. Under
  XWayland it is worse than nothing: it returns a stale position pinned to a
  screen corner, which looks like a real reading.
- **The window cannot place itself.** An `xdg_toplevel` has no say in its own
  origin, so the computed bounds were discarded even when they were correct.

`WindowKind::LayerShell` would solve both, but it needs `zwlr_layer_shell_v1`,
which Mutter does not implement — so it cannot be the fix for GNOME.

Instead the ring gets a second placement. On Wayland the window covers a display
and the panel is drawn at the pointer *inside* it, using the position the
compositor delivers with pointer-enter — the one report of the cursor a Wayland
client does get. One window per display, since nothing here can know which
display holds the cursor: `wl_pointer.enter` reaches exactly the surface under
the pointer, so the compositor settles it.

macOS, Windows and X11 keep the positioned window unchanged. The path is chosen
with `gpui::guess_compositor()`, GPUI's own backend selection, so the two cannot
drift apart.

Two details worth flagging for review:

- **Maximized, not fullscreen.** Both cover the display, but a compositor
  presents a fullscreen surface against black rather than against the desktop,
  so a transparent one renders as a black screen. Observed on Mutter; maximizing
  gets the same area without entering that presentation mode.
- **The clamp measures against the window's viewport, not the display's bounds.**
  GPUI reports display bounds in logical pixels and the pointer in device pixels,
  which disagree under fractional scaling — a 1.5x display reports 1128x752 while
  its window reports 1692x1128. Clamping against display bounds would be wrong on
  every fractionally-scaled setup.

## Changes

**`crates/openlogi-overlay`**

- `ring.rs`: new `Placement` enum — `Window` (today's behaviour) and
  `AtPointer { center }` (Wayland). The panel is extracted into
  `RingView::panel()` so both placements render the identical element.
  `ring_windows()` returns the window set for one invocation;
  `covering_window_options()` builds the display-covering window; `panel_origin()`
  is the pure placement rule.
- `session.rs`: `close_ring_windows()`. A ring is now several windows, so
  activating a slot must close all of them — the other displays' windows are
  transparent but still cover their display and would go on swallowing input.
  Keyed on the session id so a late teardown cannot close a ring that opened
  after it.
- `main.rs`: opens the window set, treating the ring as showing if any window
  opened.

Side effect: outside clicks now dismiss the ring on Wayland. The click-away
monitor is macOS-only, so that never worked elsewhere; the covering window means
the click reaches our own root handler.

## Testing

```
cargo xtask ci
# 10 passed, 0 failed, 2 skipped
# PASS rustfmt, typos, publish closure, shell, clippy, MSRV (cargo check, linux)
#      rustc 1.98, rustdoc (non-GUI crates), tests (linux), cargo-deny,
#      clippy (windows) proxy
# SKIP tests (macos) — this host is linux
# SKIP wasm (portable crates) — missing wasm32-unknown-unknown std
```

```
cargo clippy -p openlogi-overlay --all-targets -- -D warnings
cargo test -p openlogi-overlay      # 20 passed
```

Four new tests cover `panel_origin`: window-sized, pointer-not-yet-known,
centred on the pointer, and clamped against an edge.

**Runtime-tested on hardware** — GNOME 49 / Wayland, 1.5x fractional scaling,
MX Master 3S over Bluetooth, `ShowActionsRing` bound to the gesture button.
The ring opens centred on the cursor, clamps to stay whole when raised against a
screen edge, and dismisses on an outside click.

**Not tested:** macOS and Windows (unchanged code path), X11 (unchanged code
path), and KDE/wlroots. Those compositors do implement `zwlr_layer_shell_v1`,
which GPUI already exposes as `WindowKind::LayerShell` and which would be a
cleaner fit than a maximized window — worth a follow-up rather than widening
this change.

Fixes #774

### DIFF
--- a/crates/openlogi-overlay/src/main.rs
+++ b/crates/openlogi-overlay/src/main.rs
@@ -29,8 +29,10 @@ use tracing_subscriber::EnvFilter;
 use openlogi_core::action_ring::DISPLAY_LIFETIME;
 
 use crate::agent::{Ipc, OverlayCommand, spawn_ipc};
-use crate::ring::{RingView, ring_window_options};
-use crate::session::{ClickAwaySession, claim_the_role, spawn_click_away_dismissal};
+use crate::ring::{RingView, ring_windows};
+use crate::session::{
+    ClickAwaySession, claim_the_role, close_ring_windows, spawn_click_away_dismissal,
+};
 
 fn main() -> Result<()> {
     tracing_subscriber::fmt()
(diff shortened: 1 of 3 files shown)


yokalak:
Confirmed this reproduces on Hyprland (0.56.2, Wayland native, not XWayland).

**Root cause**: `hyprctl clients -j` shows the ring window (class `openlogi-action-ring`) has real floating, workspace, and pinned state, meaning it's a normal `xdg_toplevel` window, not an `xdg_popup`. That lines up with this PR's diagnosis: a toplevel has no way to read the cursor position or ask to be placed anywhere specific.

On a tiling compositor this is worse than just wrong position. Hyprland's layout takes full ownership of the window like it would any other app, so it can end up genuinely tiled into your workspace instead of just off center.

**Worth testing before #1004 merges**: its fix works by covering the screen with a maximized window and drawing the ring at the pointer inside it. That's confirmed working on GNOME. But tiling compositors like Hyprland and Sway don't always treat maximize as take over the whole screen for a normal window. It might just get tiled into a slot like anything else, unless something explicitly floats it first. If that happens, the math placing the ring at the cursor would still be correct, it would just be drawn inside a smaller, wrong sized window. Worth double checking that the covering window actually fills the monitor on Hyprland and Sway.

**In the meantime**: I put together a Hyprland specific stopgap, a window rule plus a small background script that reads Hyprland's own IPC events and moves the ring under the cursor: https://github.com/yokalak/openlogi-hyprland-ring

Not a replacement for the real fix, just something to use until #1004 lands.

PR #1018, opened 2026-08-26
fix: make configured button bindings actually fire

elmcrest:
# fix: make configured button bindings actually fire

Three independent defects that share one shape: a binding the GUI displays was
unreachable at runtime, so the mouse did something else and nothing in the logs
said why. Found while debugging "back/forward works in Firefox but not Safari" on
an MX Master 3 over a Unifying receiver, macOS 26.6, German keyboard layout.

Each commit stands alone and can be reviewed or dropped on its own.

---

## 1. `fix(core)` — thumb buttons were seeded with an action they could never run

`default_binding` seeded `Back`/`Forward` with `BrowserBack`/`BrowserForward`.
The capture plan diverts a control only when its binding *leaves* that default
(`capture_plan.rs`), so selecting `BrowserBack` matched the default, the button
was never diverted, and the raw HID button 4 went to the OS instead. Firefox
consumes buttons 4/5 itself and looked fine; Safari ignores them, so it got
nothing. The single setting meant to fix Safari was the one that guaranteed it
would stay broken.

Observable as `buttons=0` in `control capture active` while the config clearly
carried `Back = "BrowserBack"`.

Fixed by seeding `MouseBack`/`MouseForward` — what the firmware actually sends,
and the pairing `is_native_click` already treats as native. This is the same
reasoning the wheel-tilt defaults spell out twenty lines above; Back/Forward
just did not follow it.

An untouched thumb button behaves exactly as before. The browser actions become
selectable for the first time.

## 2. `fix(macos)` — browser navigation pressed the wrong physical key

With the button finally diverted, Safari still did not move. `hid_usage_to_macos`
resolves a HID usage to a **fixed ANSI** virtual key, so `⌘[` pressed whatever
character that ANSI position carries. Resolved against the active layout with
`UCKeyTranslate`:

```
kVK_ANSI_LeftBracket -> 'ü'   (German)
```

So the agent pressed `⌘Ü` and no browser navigated. Nothing in the injector
consults the keyboard layout — no `UCKeyTranslate`, no
`TISCopyCurrentKeyboardLayout`.

Fixed by using `⌘←` / `⌘→`, which every macOS browser accepts for history
navigation and which carry no layout variance. The Linux backend already used
`Alt+Left` for the same reason.

The new test pins the *property*, not the value: browser navigation must resolve
inside the named navigation block `0x4a..=0x52`, so an edit back to a punctuation
chord fails instead of silently regressing one keyboard layout.

Also removed the comment claiming Safari is served upstream via
`ax_navigate_browser()`. That function has **no call site anywhere in the
workspace** — only its definition and a re-export. The comment sent me chasing an
AX path that does not exist. The function itself is left in place; wiring it up or
deleting it is a separate decision.

## 3. `fix(agent)` — a gesture button bound to Mission Control ran the mouse's app switcher

Same default-equality trap, but worse, because a gesture source has no native
path to fall back to. `capture_plan.rs` says it directly: *"The HID++ gesture
sources never reach the OS hook."* Skipping the divert therefore does not leave
the button "native" — it leaves the **firmware's** behavior in place.

`default_binding(GestureButton)` is `MissionControl`, so a gesture button bound
to Mission Control was never diverted, and the mouse's built-in app switcher ran
while the GUI showed Mission Control.

The haptic panel already carried exactly the needed exception. Extended it to
both gesture sources: divert unless the binding is `Action::None`, which is the
one binding that genuinely asks for the firmware's behavior.

### Behavior change worth a reviewer's attention

This inverts `gestures_off_default_gesture_button_stays_native`, which asserted
the opposite:

> With gestures off and no explicit binding, the gesture button keeps its native
> HID behavior — same contract as the standard buttons.

That premise does not hold. A standard button's default action *is* what its
firmware sends (explicitly so after commit 1); a gesture source's is not. Under
the old contract "native" meant the built-in app switcher while the GUI showed
Mission Control.

The consequence reaches past the reported symptom: a gesture button with gestures
off and *no* explicit binding is now captured too, and runs its displayed action.
Scoping it narrower is not possible in this filter — an explicit
`Single(MissionControl)` and a default-derived one are the same value by the time
`plan_for_device` sees them, since `bindings` is the effective map, not the stored
one. Distinguishing them would be a data-model change.

---

## Verification

- `cargo test --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items` (GPUI crates excluded, per the documented gate)

Every new test was confirmed to fail without its fix, not just to pass with it:

| Test | Failure without the fix |
|---|---|
| `thumb_button_bound_to_a_browser_action_is_diverted` | `divert_buttons` held only `[(416, HapticPanel)]` |
| `browser_navigation_uses_layout_invariant_keys` | `BrowserBack presses usage 0x2f, whose ANSI position carries a different character on every keyboard layout` |
| `single_bound_gesture_button_is_diverted_even_at_its_default_action` | gesture button absent from `divert_buttons` |

Confirmed on hardware (MX Master 3, macOS 26.6, German layout): back/forward now
work in Safari *and* still work in Firefox. `control capture active` tracks each
fix landing — `buttons=0` before, `buttons=2` once the thumb buttons could be
diverted, `buttons=3` once the gesture source was diverted too.

No wire types changed — `Action` gains no variant and none are reordered, so
`PROTOCOL_VERSION` is untouched.

Cross-platform lints (`aarch64-unknown-linux-musl`, Windows) were **not** run
locally; the target is not installed on this machine. `linux.rs` and `windows.rs`
are untouched and no shared names or signatures changed, but that is reasoning,
not a green run — CI's word is the one that counts.

## Deliberately not included

- **The layout bug's root cause.** `hid_usage_to_macos` is a positional ANSI map
  with no layout translation, and commit 2 only steers browser navigation clear
  of it. `Undo`/`Redo` are still broken on any layout that moves `Z`: measured on
  German, `⌘Z` presses `⌘Y`. The real fix is resolving named shortcuts through
  the active layout (`TISCopyCurrentKeyboardLayoutInputSource` +
  `UCKeyTranslate`), which would also let commit 2's arrows be reconsidered.
  `CustomShortcut` should arguably keep positional semantics — there the user
  meant a key, not a character. That is Carbon FFI in `openlogi-inject` and wants
  its own PR.
- **`ax_navigate_browser`.** Left as-is, minus the misleading comment.
- **A channel-scoped LaunchAgent label.** `LABEL` is hardcoded
  `org.openlogi.agent`, so a dev build and a production install fight over one
  autostart plist — whichever reconciled last owns it. Unrelated to these three
  fixes.

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -184,6 +184,37 @@ mod tests {
     }
 
     #[test]
+    fn thumb_button_bound_to_a_browser_action_is_diverted() {
+        // The thumb buttons send mouse buttons 4/5 in firmware, which no
+        // Safari-style host consumes — reaching a browser action means the
+        // agent has to inject it, and that needs the button diverted. Seeding
+        // the default with `BrowserBack` made that unreachable: picking it
+        // matched the default, so the button stayed native and the action could
+        // never fire. Binding one side must arm its `0x1b04` divert.
+        let mut cfg = Config::default();
+        cfg.set_binding(
+            "2b023",
+            ButtonId::Back,
+            Binding::Single(Action::BrowserBack),
+        );
+
+        let plan = plan_for_device(&cfg, "2b023", route(), None, 0);
+        assert!(
+            plan.divert_buttons.contains(&(0x0053, ButtonId::Back)),
+            "a thumb button bound to a browser action must be diverted, or the binding can never \
+             fire: {:?}",
+            plan.divert_buttons
+        );
+        assert!(
+            !plan
+                .divert_buttons
+                .iter()
+                .any(|&(_, button)| button == ButtonId::Forward),
+            "the untouched forward button must keep its native button 5"
+        );
+    }
+
+    #[test]
     fn haptic_panel_gestures_when_promoted() {
         // The MX Master 4 haptic panel is a HID++ gesture source: promoting it
         // into gesture mode must arm the raw-XY gesture divert, exactly like
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -81,10 +81,20 @@ pub fn plan_for_device(
         .filter(|(_, button)| !oshook.contains_key(button))
         .filter(|(_, button)| {
             bindings.get(button).is_some_and(|action| {
-                // The panel's default is ShowActionsRing, which must be
-                // diverted to open the ring. Action::None means "leave native
-                // firmware haptics alone", so treat None as the only non-divert.
-                if *button == ButtonId::HapticPanel {
+                // A gesture source has no OS-visible native path — the hook
+                // never sees its CID — so "equals the default" cannot mean
+                // "leave it native" the way it does for a standard button.
+                // Skipping the divert there leaves the *firmware's* behavior
+                // in place instead of the bound action, which is how a gesture
+                // button bound to its own default action (Mission Control) ran
+                // the mouse's built-in app switcher instead. Action::None is
+                // the one binding that does ask for the firmware's behavior
+                // (native haptics on the panel), so it stays the sole
+                // non-divert.
+                if GESTURE_SOURCE_BUTTONS
+                    .iter()
+                    .any(|&(_, source)| source == *button)
+                {
                     *action != Action::None
                 } else {
                     *action != default_binding(*button)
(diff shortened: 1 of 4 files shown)


Rick0525:
Independent confirmation of the commit 3 symptom, from a second setup: MX Master 3 over a Unifying receiver (slot 1), macOS 15.7.3 (Sequoia), Apple Silicon.

With gestures off on the gesture button and `GestureButton = "MissionControl"` in the config (the effective default), an agent built from `master` at `b32ae08`, which does not include this PR, armed its capture as:

```
2026-08-28T03:19:08.703511Z  INFO openlogi_device::session::gesture: control capture active index=1 gesture_sources=0 dpi_buttons=1 buttons=2 thumbwheel=false
```

`gesture_sources=0` despite the displayed Mission Control binding; the `buttons=2` are `Back` and `Forward`, both bound off their defaults. **Pressing the gesture button runs the mouse's built-in app switcher while the GUI shows Mission Control.**

Device-side evidence for the "the firmware's behavior stays in place" half: reading `getCidReporting` over HID++ with OpenLogi fully stopped (the same read-only example described in #1097) shows the gesture button control `0x00c3` (`AppSwitchGesture`) at `diverted = false`, and `getCtrlIdInfo` reports its native task `0xa9`, Gesture Button Navigation. The switcher that runs is that firmware task, undisturbed, exactly as this PR describes.

One structural observation, for after this lands rather than as a change request. `default_binding` answers two different questions with one value: what the panel shows on first run, and what the firmware does when a control is left alone. The capture plan only needs the second answer, and the hook layer already has it as `is_native_click`; commit 1 leans on exactly that pairing to justify the new seeds. Today the criterion works wherever the two answers happen to coincide (the clicks, the tilts, the thumb buttons after commit 1) and breaks wherever they cannot (the gesture button, whose untouched behavior is the firmware app switcher, never a nameable `Action`), which is why commit 3 has to special-case it. If `plan_for_device` asked the `is_native_click` question directly, divert unless the bound action is the very event the control already emits on its own, the two meanings would stop sharing one table: seeds stay a GUI concern, editing one can no longer silently flip capture behavior (which is how commit 1's bug arose), and commit 3's rule follows from the general case, with `Action::None` staying as the explicit request for firmware behavior. It also sidesteps the data-model change discussed at the end of the PR body: this criterion never needs to know who set the value, so an explicit `Single(MissionControl)` and a default-derived one can stay indistinguishable.

Happy to send that as a follow-up PR once this is in, if wanted. Nothing above needs to hold this PR up; the three fixes here already cover the user-visible breakage, including the case in this comment.

PR #1020, opened 2026-08-26
feat(camera): expose pan, tilt, gain and backlight compensation

### LINKED ISSUE #663: Camera: Pan, Tilt, Gain, and BacklightCompensation have no CameraControl variant, so no backend can expose them
While probing a Logitech C930e on Windows I found four device-supported UVC controls that `CameraControl` has no variant for. They are not missing from one backend — the shared vocabulary in `crates/openlogi-camera/src/controls.rs` has no name for them, so no backend can expose them.

Reporting the device's own numbers rather than a wishlist. From an independent DirectShow probe (raw property ids, no OpenLogi code in the path):

```
[IAMCameraControl]
  Pan                    min=-10   max=10    step=1  default=0    caps=0x2
  Tilt                   min=-10   max=10    step=1  default=0    caps=0x2
  Roll                   UNSUPPORTED
  Zoom                   min=100   max=400   step=1  default=100  caps=0x2   (mapped)
  Exposure               min=-11   max=-2    step=1  default=-5   caps=0x3   (mapped)
  Iris                   UNSUPPORTED
  Focus                  min=0     max=255   step=5  default=0    caps=0x3   (mapped)

[IAMVideoProcAmp]
  Brightness / Contrast / Saturation / Sharpness                             (mapped)
  Hue                    UNSUPPORTED          <- so `Tint` is correctly absent here
  Gamma                  UNSUPPORTED
  ColorEnable            UNSUPPORTED
  WhiteBalance           min=2000  max=7500  step=1  default=4000 caps=0x3   (see #662)
  BacklightCompensation  min=0     max=1     step=1  default=0    caps=0x2
  Gain                   min=0     max=255   step=1  default=0    caps=0x2
```

So: **Pan, Tilt, BacklightCompensation, Gain** are supported by the hardware and unreachable through OpenLogi. Pan/tilt is a headline feature of this camera — zoom already works, so the panel shows one third of a PTZ control set.

## The seam already exists and the compiler enforces it

Worth saying because it makes the smaller half of this cheap: `controls.rs` is the OS-generic vocabulary, and all three backend mappings are exhaustive with no catch-all arm (`uvc.rs`, `uvc_linux.rs`, `uvc_windows.rs`). Adding a variant to `CameraControl` therefore fails to compile until every backend maps it — no silent per-OS gaps.

## Two clean halves, very different costs

**Gain and BacklightCompensation are cheap and symmetric.** All three platforms model them as ordinary scalar properties:

| control | UVC selector | V4L2 | DirectShow |
|---|---|---|---|
| Gain | `PU_GAIN_CONTROL` | `V4L2_CID_GAIN` | `VideoProcAmp_Gain` (9) |
| BacklightCompensation | `PU_BACKLIGHT_COMPENSATION_CONTROL` | `V4L2_CID_BACKLIGHT_COMPENSATION` | `VideoProcAmp_BacklightCompensation` (8) |

**Pan/Tilt is not symmetric, and the cost lands entirely on macOS.** Windows exposes `CameraControl_Pan` and `CameraControl_Tilt` as two independent properties, and V4L2 has `PAN_ABSOLUTE` / `TILT_ABSOLUTE` — but raw UVC carries both in a single `CT_PANTILT_ABSOLUTE_CONTROL` whose payload is two 4-byte signed values. `uvc.rs`'s `Payload` enum is `U16 | I16 | U32` — there is no 8-byte variant — so two generic controls would have to share one selector with a read-modify-write of half the struct.

I would suggest splitting those: Gain + BacklightCompensation as a small change, Pan/Tilt as its own with the `Payload`/UVC work scoped properly. Happy to do either or both if the shape is agreed — I have a C930e to test the Windows side against, but no macOS hardware, so the UVC half would need someone who does.

## Environment

- Windows 11 Pro 25H2 (26200.8973), x86_64
- Logitech Webcam C930e (`046d:0843`)
- Built from `master` at `822c6e4`

yuzi-co:
## Summary

Four UVC controls the hardware supports were unreachable because the shared vocabulary in `crates/openlogi-camera/src/controls.rs` had no name for them — so no backend could expose them, and a PTZ camera showed only its zoom slider. This adds all four, in two commits split along the line #663 recommends: pan/tilt, whose cost lands entirely on the raw-UVC backend, and gain/backlight compensation, which every platform models identically.

Persistence needed no change. `CameraControls` is a name-keyed `BTreeMap`, so the four names slot in beside the existing entries and older configs stay valid. The CLI resolves control names from `CameraControl::ALL` and picked them up for free.

## Changes

- **`openlogi-camera`** — `CameraControl` gains `Pan`, `Tilt`, `Gain` and `BacklightCompensation`. Windows maps them to `CameraControl_Pan`/`_Tilt` and `VideoProcAmp_BacklightCompensation`/`_Gain`; Linux to `V4L2_CID_PAN_ABSOLUTE`/`TILT_ABSOLUTE`/`GAIN`/`BACKLIGHT_COMPENSATION`.

  macOS is the asymmetric one: raw UVC carries both axes in a single 8-byte `CT_PANTILT_ABSOLUTE_CONTROL` (`dwPanAbsolute` then `dwTiltAbsolute`), so `Payload` gains a `PanTilt(Axis)` variant addressing one word of that pair, and the read buffer widens to the new maximum payload. A write reads the live pair, replaces one axis and sends both back — a bare 4-byte write would recentre the other axis. A failed read is surfaced rather than written over with a zeroed sibling word, so a transport error cannot silently recentre an axis the caller never named.

- **`openlogi-desktop`** — labels for the four, and pan/tilt join zoom/focus/exposure in the Lens section; gain and backlight compensation sit with the image controls.

- **`openlogi-ui`** — `Pan`, `Tilt`, `Gain`, `Backlight compensation` in all 22 locale catalogs.

- **`openlogi-cli`** — the hand-maintained control-name list in `camera set --help` now names all four.

- **README** — the camera feature bullet lists them.

Built-in profiles are deliberately untouched: Streaming and Video call encode a look, and baking a pan value into a shared preset would swing the camera off-centre when you pick one. Reset still recentres both axes, because it walks the device's reported defaults.

## Testing

Gate on Windows 11 Pro (10.0.26200), `RUSTFLAGS="-D warnings"`, on this branch
rebased onto current `master` (`d5e2b880`):

- `cargo fmt --all -- --check` - pass
- `cargo test -p openlogi-ui locale` - pass (catalog parity: key set, order, placeholders)
- `cargo clippy -p openlogi-camera -p openlogi-cli -p openlogi-desktop -p openlogi-ui --all-targets -- -D warnings` - pass
- `cargo test -p openlogi-camera -p openlogi-cli -p openlogi-ui` - pass
- `cargo test -p openlogi-desktop` - pass (includes the `i18n` key-resolution tests)

Not runtime-tested on hardware since the rebase - the camera behaviour was verified
before it, and nothing in the re-port changes runtime logic.

### Rebase note: locale catalogs moved to semantic TOML keys

`master` migrated `locales/*.yml` to `locales/*.toml` with semantic keys (#1169),
which deleted every file this branch's locale hunks applied to. The strings are
re-authored rather than rebased:

- `Pan` -> `camera.pan`, `Tilt` -> `camera.tilt` (after `camera_controls_unavailable`)
- `Gain` -> `camera.gain`, `Backlight compensation` -> `camera.backlight_compensation`
  (after `low_light_compensation`)

`control_label()` resolves all four. Note `camera_controls_unavailable` reads
"This camera has no adjustable image controls." on `master`; that wording wins over
the branch's "exposes no" - a copy change, same key.

All 23 catalogs keep the human translations this branch already carried - they were
lifted out of the old `.yml` files before those were deleted, not refilled with
English. `cargo test -p openlogi-ui locale` is the check that they line up.

Fixes #663


yuzi-co:
**Update — #662 has merged, and this branch no longer depends on it.**

Why the merge-base moved after the review on `a9adbee9`: #662 was **squash**-merged
as `8d666e32`, which gave master a new SHA for the same one-line edit this branch
was carrying as `1ce05151`. The same patch under two identities is a conflict, so
the PR went to CONFLICT roughly twelve minutes after that review, through no change
on this side.

Rebasing onto the new master resolved it, and git dropped the duplicate by patch-id
on its own:

```
warning: skipped previously applied commit 1ce05151
```

What is left is three commits on `8d666e32` — pan/tilt, gain/backlight, README —
and the `uvc_windows.rs` diff against master is now purely additive, with
`VPA_WHITE_BALANCE = 7` inherited rather than carried:

```rust
+const VPA_BACKLIGHT_COMPENSATION: i32 = 8;
+const VPA_GAIN: i32 = 9;
+const CC_PAN: i32 = 0;
+const CC_TILT: i32 = 1;
```

The force-push that did this dismissed the review on `a9adbee9`, so re-approval has
to land on `8b8cf3e3`. Worth avoiding a re-run on the old SHA: CI concurrency here
is per branch with `cancel-in-progress`, so approving a stale run cancels the one on
the current head.

Local gate on the rebase, Windows: `cargo fmt --all -- --check` and
`cargo clippy --workspace --all-targets -- -D warnings` clean,
`cargo test --workspace --no-fail-fast` green on every target except the
pre-existing xtask `ci_yml_runs_what_this_runner_runs` drift.

<details>
<summary>Original note, when this branch was still stacked on #662</summary>

Rebased onto #662 rather than master, so this branch now carries the white-balance
fix as its first commit.

The two conflicted, because they touch the same constant block. `windows-0.61`'s
DirectShow module has `VideoProcAmp_ColorEnable = 6`, `VideoProcAmp_WhiteBalance = 7`,
`VideoProcAmp_BacklightCompensation = 8`, `VideoProcAmp_Gain = 9`. The `8` and `9`
this PR adds only line up if white balance is `7`, so leaving it at `6` put a hole
at 7 in the middle of a contiguous run:

```rust
const VPA_SHARPNESS: i32 = 4;
const VPA_WHITE_BALANCE: i32 = 7;
const VPA_BACKLIGHT_COMPENSATION: i32 = 8;
const VPA_GAIN: i32 = 9;
```

Merge order works either way. Take #662 first and this rebases onto master with that
commit dropping out on its own, or squash this one and close #662 as included.

The `CC_*` values are unaffected and all match `CameraControlProperty`: Pan 0,
Tilt 1, Zoom 3, Exposure 4, Focus 6.


</details>

davidbudnick (dismissed):

davidbudnick (dismissed):

PR #1029, opened 2026-08-26
feat: make thumb-wheel horizontal scroll direction reversible

### LINKED ISSUE #619: Thumb wheel horizontal scroll can't be reversed: default binding pair is inverted vs. native scroll direction
# Thumb wheel horizontal scroll can't be reversed: default binding pair is inverted vs. native scroll direction

**Device:** MX Master 4 (Bolt receiver) · **OS:** macOS 15.7.9 · **Version:** v0.6.27 (official cask)

## Summary

On the MX Master 4, the thumb wheel's **native** (undiverted) scroll direction is **up-rotation → scroll left, down-rotation → scroll right**. OpenLogi's default binding pair says the opposite:

https://github.com/AprilNEA/OpenLogi/blob/v0.6.27/crates/openlogi-core/src/binding/defaults.rs#L36-L37

```rust
ButtonId::ThumbwheelScrollUp => Action::HorizontalScrollRight,
ButtonId::ThumbwheelScrollDown => Action::HorizontalScrollLeft,
```

Because the wheel is only diverted when something differs from defaults (`thumbwheel_armed`, [`watchers/gesture.rs#L100`](https://github.com/AprilNEA/OpenLogi/blob/v0.6.27/crates/openlogi-agent-core/src/watchers/gesture.rs#L100): sensitivity ≠ `DEFAULT_THUMBWHEEL_SENSITIVITY` or a thumbwheel binding ≠ default, computed in [`capture_plan.rs#L96`](https://github.com/AprilNEA/OpenLogi/blob/v0.6.27/crates/openlogi-agent-core/src/capture_plan.rs#L96)), this inversion is invisible in the default state — native scroll is what users feel, and it's fine. But it surfaces as two real bugs:

### 1. Scroll direction cannot be reversed at all

Swapping the two bindings (`ThumbwheelScrollUp = "HorizontalScrollLeft"`, `ThumbwheelScrollDown = "HorizontalScrollRight"`) is non-default → the wheel is diverted → `advance()` injects sign from the action ([`gesture.rs#L474-L484`](https://github.com/AprilNEA/OpenLogi/blob/v0.6.27/crates/openlogi-agent-core/src/watchers/gesture.rs#L474-L484)) → up injects Left, down injects Right — **which is exactly the native direction**. Verified on hardware: the swapped config feels identical to the default config. Meanwhile the mapping that *would* reverse it (up→Right/down→Left, injected) equals the defaults, so it never diverts and never takes effect. Net: no combination of scroll bindings changes the felt direction, and the GUI (correctly, given the model) exposes no reverse option.

### 2. Moving the sensitivity slider silently flips scroll direction

Changing `thumbwheel_sensitivity` off 14 — with untouched default bindings — arms the divert by itself, switching the user from native (up→left) to injected defaults (up→right). A user who only wanted faster/slower scrolling gets their horizontal scroll direction reversed as a side effect.

## Repro

1. MX Master 4, default config: rotate thumb wheel up → content scrolls left (native).
2. Set only `thumbwheel_sensitivity = 15`, restart agent: rotate up → content scrolls **right** (bug 2).
3. Set bindings Up→`HorizontalScrollLeft` / Down→`HorizontalScrollRight` (the "swap"): rotate up → content scrolls left — same as step 1, no reversal possible (bug 1).

## Proposed fix

Swap the two defaults in `defaults.rs`:

```rust
ButtonId::ThumbwheelScrollUp => Action::HorizontalScrollLeft,
ButtonId::ThumbwheelScrollDown => Action::HorizontalScrollRight,
```

(+ the GUI's default pair and affected tests, + the doc comment above the defaults.) Then:

- injected-default matches native → sensitivity changes no longer flip direction;
- the binding swap becomes non-default → diverts → genuinely reverses, giving users a working (if TOML-only) reverse today;
- a GUI "reverse direction" toggle becomes a trivial follow-up (it just writes the swapped pair).

Caveats worth a maintainer call: I've verified the native direction on one device (MX Master 4 via Bolt, macOS — the injected CGEvent and native scroll both pass through the same system scroll-inversion handling, so macOS "natural scrolling" doesn't affect the relative comparison). If other thumb-wheel devices share the firmware convention this is universal; and existing configs that explicitly wrote the old default pair would start diverting after the swap (behavior-preserving direction-wise, but they lose native-scroll latency — may warrant a migration touch).

Happy to send the PR if the direction (pun intended) looks right to you.

**Workaround for anyone hitting this now:** keep the default bindings and set `thumbwheel_sensitivity` to 15 (forces the diverted path, whose defaults are the reverse of native; ~5% faster than 14, imperceptible).

dalxds:
## Summary

Makes the thumb wheel's horizontal scroll direction reversible, as tracked in #619.

Two pieces, because the preset alone cannot work: the wheel is only diverted when a thumb-wheel setting leaves its default, and the current defaults (`ThumbwheelScrollUp → HorizontalScrollRight`, down → left) are the opposite of the wheel's native firmware direction (up → left, verified on an MX Master 4 over Bolt on macOS — see #619). With that seed, the swapped binding pair diverts into the native direction (a felt no-op), and no combination of scroll bindings changes the felt direction. Swapping the defaults to match native fixes that, also fixes #619's bug 2 (a sensitivity change silently flipping scroll direction), and makes a "Horizontal Scroll (Reversed)" preset a working, honest option.

## Changes

- `openlogi-core`:
  - `binding/defaults.rs`: swap the two thumb-wheel scroll defaults to the native firmware direction (up → `HorizontalScrollLeft`, down → `HorizontalScrollRight`), with the comment rewritten to state the constraint (the old comment asserted the opposite direction as the firmware's).
  - `config.rs` / `config/file.rs`: schema v7 plus a one-time load migration. A pre-v7 device whose thumb-wheel trio sits entirely on the old defaults — the pair the GUI's "Horizontal Scroll" preset wrote — scrolled natively (undiverted); left as written it would now divert into a felt reversal, so the pair is rewritten onto the new defaults and keeps feeling exactly as it did. Devices with any other thumb-wheel binding were already diverted with each explicit direction injecting the action it names, which the swap does not change, so they are left as written.
  - Tests for the new defaults, the migration, its customized-trio edge, and the version gate (a v7 file carrying the reversed pair must survive reloads).
- `openlogi-desktop`: new "Horizontal Scroll (Reversed)" thumbwheel preset; the plain "Horizontal Scroll" preset now writes the (new) default pair so picking it keeps the wheel undiverted.
- `openlogi-ui`: the preset's label key added to all 22 locale catalogs (best-effort translations; Crowdin can refine).
- `docs`: `CONFIGURATION.md` migration note and current schema version (the text still said `5`), `config.example.toml` bumped to v7.

## Caveats

- Native direction is verified on one device (MX Master 4 / Bolt / macOS). If some thumb-wheel firmware scrolls the other way, its untouched-default feel is unchanged either way (undiverted), but the two presets' labels would be swapped for it.
- Per-app overlay entries holding the old default pair are deliberately not migrated: pre-v7 they were inert (equal to defaults → no divert), post-v7 they become an active per-app reversal matching what they literally say. Migrating them felt-preservingly would need per-app effective-trio analysis; can follow up if you want it.
- No wire-format change: only default values moved; no serde shapes or enum variants changed, so no `PROTOCOL_VERSION` bump.

## Testing

- `cargo clippy --workspace --exclude openlogi-desktop --exclude openlogi-overlay --all-targets -- -D warnings` (with `RUSTFLAGS="-D warnings"`)
- `cargo test --workspace --exclude openlogi-desktop --exclude openlogi-overlay` — includes the `openlogi-ipc` wire-format goldens and the `openlogi-ui` locale parity test
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent`
- `cargo check -p openlogi-hidpp -p openlogi-device --target wasm32-unknown-unknown` and `cargo check -p openlogi-core --no-default-features --target wasm32-unknown-unknown`

Not run locally (this host has Command Line Tools only, so GPUI's Metal shaders don't build): `openlogi-desktop` / `openlogi-overlay` clippy and tests — that includes the updated `ThumbwheelPreset` unit tests and the desktop i18n coverage test; CI's Linux clippy and macOS `--all-targets` tests cover them. Also not run locally: typos, cargo-deny, MSRV, Windows clippy, shell (no shell changes; no dependency changes).

Not runtime-tested on hardware. To verify on an MX Master 4 (or any 0x2150 device): (1) default config — thumb wheel should keep scrolling natively, and moving the sensitivity slider off 1× should no longer flip the direction; (2) pick "Horizontal Scroll (Reversed)" — the direction should genuinely flip; (3) pick "Horizontal Scroll" — back to native. A pre-upgrade config that had picked "Horizontal Scroll" should load feeling unchanged (migration), and a v7 file with the reversed pair must keep it across restarts.

No screenshot: the new picker row reuses the existing preset row rendering (label + existing chevrons icon); I could not build the GUI on this host.

Fixes #619

### DIFF
--- a/crates/openlogi-core/src/binding/defaults.rs
+++ b/crates/openlogi-core/src/binding/defaults.rs
@@ -57,13 +57,17 @@ pub fn default_binding(button: ButtonId) -> Action {
                       not because the control stays native like the keyboard arm below"
         )]
         ButtonId::Thumbwheel => Action::None,
-        // The thumb wheel scrolls horizontally by default: rotating it produces
-        // continuous horizontal scroll, with "up" → right and "down" → left.
-        // The wheel watcher renders these two actions as smooth, sensitivity-
-        // scaled scrolling rather than the discrete per-press burst a button
-        // would get (see `watchers::gesture`).
-        ButtonId::ThumbwheelScrollUp => Action::HorizontalScrollRight,
-        ButtonId::ThumbwheelScrollDown => Action::HorizontalScrollLeft,
+        // The thumb wheel scrolls horizontally in firmware: "up" (forward)
+        // scrolls left and "down" scrolls right. As with the tilt above, the
+        // seed must match that native direction — once anything arms the
+        // divert, the injected scroll follows these actions, so a mismatched
+        // seed makes the swapped pair a felt no-op and flips the felt
+        // direction whenever something else (a sensitivity change) diverts the
+        // wheel. The wheel watcher renders these two actions as smooth,
+        // sensitivity-scaled scrolling rather than the discrete per-press
+        // burst a button would get (see `watchers::gesture`).
+        ButtonId::ThumbwheelScrollUp => Action::HorizontalScrollLeft,
+        ButtonId::ThumbwheelScrollDown => Action::HorizontalScrollRight,
         ButtonId::GestureButton => Action::MissionControl,
         ButtonId::HapticPanel => Action::ShowActionsRing,
         // Keyboard keys stay on their native firmware function until the user
--- a/crates/openlogi-core/src/binding/tests.rs
+++ b/crates/openlogi-core/src/binding/tests.rs
@@ -533,6 +533,22 @@ fn wheel_tilt_defaults_to_the_scroll_its_firmware_already_does() {
     }
 }
 
+#[test]
+fn thumbwheel_defaults_to_the_scroll_its_firmware_already_does() {
+    // Same constraint as the wheel tilt above, with a sharper failure mode: a
+    // seed that doesn't match the wheel's native direction (up → left) makes
+    // the swapped binding pair a felt no-op and turns any other divert cause
+    // (a sensitivity change) into a silent direction flip.
+    assert_eq!(
+        default_binding(ButtonId::ThumbwheelScrollUp),
+        Action::HorizontalScrollLeft
+    );
+    assert_eq!(
+        default_binding(ButtonId::ThumbwheelScrollDown),
+        Action::HorizontalScrollRight
+    );
+}
+
 // ── Effect classification ─────────────────────────────────────────────────
 //
 // `Action::effect()` is the platform-neutral IR `openlogi-inject`'s three
--- a/crates/openlogi-core/src/config.rs
+++ b/crates/openlogi-core/src/config.rs
@@ -49,6 +49,13 @@ use settings::GestureOwner;
 /// persisted shape or enum vocabulary changes; readers inspect this value
 /// before consuming the rest of the file.
 ///
+/// v7 swaps the two thumb-wheel scroll defaults to the wheel's native firmware
+/// direction (up → left, down → right). A pre-v7 file whose thumb-wheel
+/// bindings sit on the old defaults — the pair the GUI's "Horizontal Scroll"
+/// preset wrote — is rewritten onto the new defaults
+/// (see `Config::migrate_thumbwheel_native_direction`) so it keeps scrolling
+/// natively instead of starting to divert into a felt reversal.
+///
 /// v6 adds threshold-based `{ short = ..., long = ... }` button bindings.
 ///
 /// v5 also drops the transport prefix from `direct:` keys: `direct:046d:c08d:unit:6be9d300`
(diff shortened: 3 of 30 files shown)


AprilNEA:
Thanks for taking this on and for documenting the migration concerns so carefully.

After reviewing the direction model, I don’t think we should merge this implementation. Thumb-wheel polarity is device-specific (`0x2150 default_dir`), while this PR changes a global default based on one device. That would leave opposite-polarity devices with the same bug and also change the behavior of existing per-app configurations.

The fix needs to normalize device polarity at the capture boundary and represent scroll inversion explicitly, rather than infer it from swapped default action bindings. That is a broader redesign across the device, agent, and config layers, so reshaping this PR would effectively mean rewriting it.

I’m closing this PR while keeping #619 open to track the requirement and the replacement implementation. The investigation and tests here were still very helpful—thank you.

PR #1037, opened 2026-08-26
refactor(agent): split launch_agent into per-platform autostart modules

AprilNEA:
> Stacked on #1031 (`feat/agent-launchd-lifecycle`); will be rebased and retargeted to `master` once that merges. The diff shown here is the refactor alone.

## Summary

`launch_agent.rs` interleaved three platforms' constants, reconcile logic, helpers, and tests through ~15 `#[cfg(target_os = …)]` gates in one 366-line file. This splits it into the shape the workspace already uses for its larger platform surfaces (`inject.rs` → `inject/{macos,linux,windows}.rs`, hook likewise): one module per OS behind a single `cfg` at the declaration, a thin facade owning the dispatch. Pure move — each platform file carries its mechanism, its logging, and its tests unchanged; no behavior differs.

The convention itself is now recorded in `.claude/rules/rust.md`'s module-layout section: per-platform modules for compile-time-exclusive divergence (each file implements the same function names, so a missing one fails that platform's compile — the same guarantee a trait would give); a trait only where implementations genuinely coexist (runtime backend selection, test doubles, or a cross-crate seam like `HidBackend`); single small divergent functions stay inline. The new `launch_agent/{linux,windows}.rs` files are also added to `.claude/rules/cross-platform.md`'s explicit watch list.

The module is also renamed `launch_agent` → `autostart`: the old name described the 2026-06 implementation (a hand-written macOS LaunchAgent plist) that no longer exists — the macOS arm only migrates legacy plists away, and the Linux/Windows arms never were LaunchAgents. All path references (CODEOWNERS, the rule files, brand.rs docs) follow; the GUI's unrelated local `fn launch_agent` (the "start the agent binary" verb) keeps its name.

## Changes

- `openlogi-agent`: `launch_agent.rs` becomes the facade (module doc + `reconcile` dispatch + the no-backend fallback arm); `launch_agent/macos.rs` (legacy-plist migration), `launch_agent/linux.rs` (systemd user unit + its 9 tests), `launch_agent/windows.rs` (HKCU Run key). Each platform module exposes `pub(super) fn reconcile(enabled: bool)` and owns its error logging.
- `.claude/rules/rust.md`: the platform-module vs trait convention, with the in-tree examples.
- `.claude/rules/cross-platform.md`: the two new non-macOS files join the explicit list.

## Testing

- `cargo test -p openlogi-agent` (21 tests) and `cargo clippy -p openlogi-agent --all-targets -- -D warnings` — green on macOS.
- `cargo xtask ci clippy-windows` — pass (covers `launch_agent/windows.rs` and the facade on the Windows lane).
- Linux lane: not cross-lintable from this machine (the devenv toolchain carries no Linux target std; the musl recipe in cross-platform.md assumes rustup). Per that rule's option 2 the moved file was hand-audited instead: `linux.rs` is a byte-identical move of code CI built green twice today, and the only new Linux-visible surface is the module declaration, the `reconcile` wrapper, and the consolidated imports — each checked against its uses. CI's Linux clippy/tests are the authority.
- No runtime behavior change intended; not runtime-tested beyond the macOS test suite.


AprilNEA:
Folded into #1031 directly (commits d3b757a2 + b9dbb380) — keeping the lifecycle change and its module reorganization in one PR.

PR #1047, opened 2026-08-26
feat(gui): add precise camera control entry

vinayaksodar:
## Summary
Add precise numeric entry for UVC camera image controls so users can set exact values when slider positioning is difficult.

## Changes
- Make each displayed camera control value clickable.
- Open a numeric editor with the device-reported supported range.
- Validate exact values and reuse the existing slider-release write, auto-mode takeover, and persistence path.

## Testing
- `git diff --check`
- Rust formatting and compilation were not run because Cargo is unavailable in the development shell.
- Not runtime-tested on camera hardware.

### DIFF
--- a/crates/openlogi-desktop/src/features/camera/controls.rs
+++ b/crates/openlogi-desktop/src/features/camera/controls.rs
@@ -19,7 +19,10 @@ use gpui::{
 };
 use gpui_base::Button as BaseButton;
 use gpui_component::{
-    IconName, Selectable as _, h_flex,
+    IconName, Selectable as _, WindowExt as _,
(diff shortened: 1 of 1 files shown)


vinayaksodar:
This is vibe coded but is just a minor ui fix so should be fairly easy to review, the problem with sliders is it is very hard to set values exactly, I struggled a lot to set exposure value to exactly 300 to fix flickering in webcam.

davidbudnick (dismissed):

vinayaksodar:
Added the single semicolon that caused clippy to fail, all other check had passed

davidbudnick (approved):

PR #1069, opened 2026-08-27
fix(camera): surface startup failures and retry preview

### LINKED ISSUE #1019: Camera: start_stream reports success for a camera another app holds, so the preview shows a silent black box
`openlogi_camera::start_stream` returns `Ok(CameraStream)` for a camera another application is already streaming. Setup succeeds — the Media Foundation source activates and the media type negotiates — and only the first `ReadSample` fails, on the reader thread, where nothing surfaces it. The stream then runs forever at zero fps.

The GUI preview cannot distinguish that from a live stream whose first frame has not arrived, so it renders its placeholder indefinitely: the user gets a black box instead of "camera is in use by another application".

## Measured

Logitech C930e (`046d:0843`), Windows 11 Pro 26200, built from `master` at `afb10a4c`:

| camera state | `start_stream` | frames in 5 s |
|---|---|---|
| idle | `Ok` in 188 ms | 44 (~8.8 fps), 1280x720 |
| held by the Windows Camera app | `Ok` in 169 ms | **0** |

In the busy case the first `ReadSample` returns `MF_E_HW_MFT_FAILED_START_STREAMING` (`0xC00D3704`).

## Not a media-type problem

A reader that sets no native media type and accepts the in-use 1920x1080 format fails identically, so this is not the `SetCurrentMediaType` mode switch in `open_reader`. Windows refuses this camera to a second client at the platform level — two unrelated apps cannot share it either — so the fix is reporting the condition, not working around it.

## Suggested fix

`reader_thread` already has a setup channel it uses to report failures before `pump_frames` starts. Pulling the first sample inside the setup window, or forwarding the first `ReadSample` error through that channel, would let `start_stream` fail with a new `CaptureError::InUse` variant that the preview can render as a real message.

## Separately

The idle figure is worth its own look: `open_reader` scores native formats by width alone, so it picks an uncompressed mode, and the C930e caps YUY2 720p at ~10 fps where MJPEG would deliver 30. The preview is running at a third of the frame rate the hardware offers.

yuzi-co:
## Summary

Windows capture startup now reports success only after a decoded frame has actually been stored. Successful `ReadSample` calls with no sample, stream ticks, and malformed buffers do not establish readiness. This fixes the zero-frame false success behind #1019.

The desktop displays startup errors and retries after two seconds while the preview remains selected. This adopts the contributor's two commits onto the fetched master baseline [e041973e](https://github.com/AprilNEA/OpenLogi/commit/e041973e1237bfd3c9ee259c6a610e4360766af5), preserving their authorship and the newer CameraPreview lifecycle/tests from #1182 and #1322. The checks below apply to this branch, not to a synthetic merge with later master commits.

`MF_E_HW_MFT_FAILED_START_STREAMING` means unavailable hardware resources, not proof that another application owns the camera. The new wording preserves that uncertainty. See Microsoft's [ReadSample contract](https://learn.microsoft.com/en-us/windows/win32/api/mfreadwrite/nf-mfreadwrite-imfsourcereader-readsample) and [IMFMediaSource::Start remarks](https://learn.microsoft.com/en-us/windows/win32/api/mfidl/nf-mfidl-imfmediasource-start).

## Changes

- **openlogi-camera:** wait for actual frame publication within the existing five-second setup budget; handle ERROR/ENDOFSTREAM flags as bitmasks; honor cancellation before/after blocking reads; preserve Media Foundation/COM teardown and timeout detachment. Classify resource shortage as `ResourcesUnavailable` and access denial as `AccessDenied`. Conversion/lock/unlock failures and poisoned storage return errors rather than claiming readiness.
- **openlogi-desktop:** `StartFailed` owns the target, typed error and pending retry task. Repeated renders remain idempotent. Retargeting, stopping and success cancel retries; revoked access moves back to `AwaitingAccess`. Error changes notify the view. Resource, permission and generic failures have distinct messages; long notes have centered wrapping and horizontal padding. Simple error matching stays inline.
- **openlogi-ui:** add the same two keys in all 23 catalogs, preserving current master translations for existing messages. New translations describe another application's use as a possibility, not a diagnosis.
- **Cargo.lock:** update rustls 0.23.40 → 0.23.45 and its required rustls-webpki dependency 0.103.13 → 0.103.15. This resolves the cargo-deny failure for [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285), published September 14, 2026. No advisory exemption was added; GPUI and gpui-component pins are unchanged.

No manifest, IPC type, protocol version, macOS capture backend or signing change. No WinRT SharedReadOnly fallback or capture-performance refactor.

## Testing

Final-tree Linux orb checks used `RUSTFLAGS="-D warnings"`:

- `cargo fmt --all -- --check` — passed.
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — passed.
- `cargo test --workspace --locked` — 1451 passed, 0 failed, 2 existing ignored documentation examples (`openlogi-hidpp` and `openlogi-hidpp-derive`). Includes 203 desktop tests.
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent --locked` — passed.
- `cargo test -p openlogi-ui locale --locked` — catalog parity passed.
- `cargo test -p openlogi-desktop i18n --locked` — desktop key resolution passed.
- `cargo xtask ci clippy-windows` — 1 passed, 0 failed, 0 skipped. This is the ring-free cross-lint proxy, not native full-workspace Windows validation; it excludes the desktop.
- `cargo xtask ci cargo-deny` — 1 passed, 0 failed, 0 skipped with cargo-deny 0.20.2 after the TLS dependency update. Advisories, bans, licenses and sources all passed.
- Focused `cargo test -p openlogi-desktop features::camera::preview --locked` — 10 passed. Covers the 1999ms/2000ms boundary, repeated failures then success, changed-error notification, target-specific deadlines, stop/drop and permission revocation, plus existing stream/texture lifecycle contracts.
- Eight new Windows-specific tests were cross-compiled/linted, **not executed**. They cover frame-publication readiness, null/short samples, asymmetric BGRA rows, one-shot acknowledgement, EOF/error flags, cancellation, caller disconnection, poisoned storage and HRESULT classification.

Native GPUI rendering was inspected in light and dark themes using injected capture results through the existing private capture seam. All four representative panels rendered legible, centered notes without clipping. The temporary gallery fixture was removed afterward. This verifies appearance, not Windows capture or real-camera operation.

Native GPUI light theme with injected capture results

Native GPUI dark theme with injected capture results

Not runtime-tested on hardware. No macOS or Windows runtime tests were available in the orb. Other CI jobs (native Windows/macOS matrix, MSRV, typos, wasm, shell and packaging) were not run locally as part of this update; old-head CI/Greptile results are not evidence for this revision.

### Hardware smoke test

1. On Windows with a Logitech camera, open the preview while idle; verify image delivery and normal teardown/LED release when leaving the tab.
2. Hold the camera in another application, then open the preview. Verify a startup failure replaces indefinite “Starting preview…”. Resource-unavailable wording applies when the driver returns the corresponding HRESULT.
3. Release the camera in the other application without interacting with OpenLogi. Verify recovery after the retry interval plus capture setup time.
4. While failure/retry is pending, change devices or leave the preview. Verify the old target is not reopened and teardown releases the camera.
5. Deny camera access, check the permission guidance, restore access and verify recovery. Repeat permission-change and tab/device teardown checks on macOS.

### Limits retained

Capture startup remains synchronous. The existing five-second wait now includes first-frame delivery; this does not make the GUI's startup call asynchronous. Blocking native calls remain cooperatively cancellable and may outlive the caller's timeout. A successful startup proves initial delivery, not ongoing stream liveness after later failures. Shared camera access and post-start stream recovery remain outside this change.

Fixes #1019


yuzi-co:
Good catch on the retry, fixed in `edbcb890`.

The mechanism is slightly worse than described. `set_target` does return early on
an unchanged target, but even without that early return the retry would never
have fired: the parent calls `set_target` from `render`, and the only thing
driving renders on this view is the repaint pump, which exists only while
`stream.is_some()`. A preview that fails to start stops being rendered, so
nothing asks again. It needed a timer, not a relaxed guard.

So a failed start now arms a reopen pump on a 2 s interval, which stops as soon
as a stream comes up, the target changes, or the target goes away. Leaving the
tab drops the task, because the parent passes `None` on every render where the
camera tab is not the active one, so the polling cannot outlive the thing that
is on screen.

`start_stream` was doing three jobs, so it is now split: `open_stream` attempts
one open and records why it failed, `pump_frames` starts the repaint pump, and
`start_stream` sequences them. The retry loop calls `open_stream` rather than
`start_stream` deliberately, since it must not assign `retry_task` while that
field still holds the task it is running inside.

Windows gate on `edbcb890`: `cargo fmt --all -- --check` clean,
`cargo clippy -p openlogi-desktop --all-targets -- -D warnings` exit 0,
`cargo test -p openlogi-desktop` green (184/184). Still not runtime-tested on
hardware; the check that matters is starting the preview with the camera held by
another application, then quitting that application and watching the image
appear within a couple of seconds without touching the UI.

davidbudnick (dismissed):

AprilNEA (approved):

PR #1124, opened 2026-08-28
fix(hid): end capture when liveness writes fail

### LINKED ISSUE #1125: [Bug]: Windows: SmartShift remap stops until agent restart
### Component

CLI / background agent

### OpenLogi version

0.8.1

### Operating system

Windows 11 Home 64-bit, build 26200

### Device & connection

Logitech MX Master 3, Bluetooth direct (no receiver). No Logi Options+ package or process found.

### What happened?

An OpenApplication action bound to the SmartShift/DpiToggle button (CID 0x00c4) works right after an agent restart, then stops working after extended runtime. When it fails, native SmartShift resumes.

Observed during a failure:

- `getCidReporting` on CID 0x00c4 reports `diverted=false` — the agent's diversion was dropped.
- Manually writing `diverted=true` succeeds and sticks, but the button is swallowed and no action runs.
- The agent itself stays responsive: inventory IPC answers and `openlogi diag dpi --device "MX Master"` succeeds.

The device/command paths remain healthy while the gesture-capture consumer is dead. From reading the source, the capture session's liveness check treats transport errors other than Timeout/NoResponse as "healthy", so the dead session is never torn down and state is never re-armed. A proposed fix is under review in #1124.

The exact trigger and timing are unknown; it appeared after extended runtime.

### Steps to reproduce

Not deterministic — observed to appear after extended runtime.

1. Windows 11, MX Master 3 over Bluetooth-direct, OpenLogi v0.8.1 agent running.
2. Bind an action (e.g. OpenApplication) to the SmartShift/DpiToggle button; confirm it fires right after an agent restart.
3. Leave the machine running until the bound action stops firing and native SmartShift resumes.
4. Power-cycling the mouse (10 s off/on) does not restore capture; restarting the agent does.

### Diagnostics / logs

Not captured.

ikchifo:
## Summary

Fix a liveness classification that leaves Windows Bluetooth gesture capture dead until the agent restarts.

## Changes

- `openlogi-device`: treat non-timeout channel failures as dead while preserving the two-strike silence policy and device-reply handling.
- `openlogi-device`: add scripted regression coverage for live, silent, feature-error, and write-dead channels.

## Testing

- `RUSTFLAGS="-D warnings" cargo fmt --all -- --check`
- `cargo clippy -p openlogi-device -p openlogi-hid -p openlogi-agent -p openlogi-agent-core --all-targets -- -D warnings`
- `cargo test -p openlogi-device -p openlogi-hid -p openlogi-agent -p openlogi-agent-core` (416 passed)
- A v0.8.1 backport recovered after a 30-second MX Master 3 power cycle on Windows. The current-master branch was not runtime-tested on hardware.

Fixes #1125

### DIFF
--- a/crates/openlogi-device/src/session/gesture.rs
+++ b/crates/openlogi-device/src/session/gesture.rs
@@ -53,12 +53,58 @@ use crate::thumbwheel::{self, Thumbwheel, WheelDirection, WheelResolution};
 
 /// How often the capture session pings its device to prove the channel still
 /// delivers input reports. Cheap: one HID++ round-trip per interval.
+///
+/// Tests shrink this so the real-time response wait (a paused tokio clock
+/// cannot govern the channel's `futures_timer` deadline) still runs fast.
+#[cfg(not(test))]
 const LIVENESS_PING_INTERVAL: std::time::Duration = std::time::Duration::from_secs(20);
+#[cfg(test)]
+const LIVENESS_PING_INTERVAL: std::time::Duration = std::time::Duration::from_millis(50);
 
 /// Consecutive all-silent pings after which the capture channel is declared
 /// dead. Two, so one ping lost to transient receiver congestion (which does
 /// happen under pointer load) doesn't churn the session.
 const LIVENESS_PING_STRIKES: u8 = 2;
+
+/// The data byte that marks the session's own liveness ping on the wire
+/// (`root.ping(0x5a)`), so a scripted transport can fail exactly this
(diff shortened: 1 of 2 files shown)


ikchifo:
Superseded by #1137 (merged), which introduced the same liveness classification — silent pings stay on the two-strike path, device replies prove delivery, and other channel failures end the capture session immediately — alongside event-aware probe deferral. Closing in its favor; leaving #1125 open for hardware/release verification.

Issue #1141, opened 2026-08-29
Smooth scroll hardware report (MX Master 4, macOS): synthetic output bypasses OS wheel acceleration — fast spins travel less than with smoothing off


isleofgreg:
## Summary

PR #975 shipped the smooth-scroll motion model with the note *"Not runtime-tested on hardware. Verify with `smooth_scroll = true`"* — this is that hardware verification, from an MX Master 4 on macOS (Sequoia 15.7, Mac Studio), wheel in permanent free-spin mode.

The plumbing works exactly as designed: per-source state, one balanced Began/Changed/Ended lifecycle, fail-open under pressure, absolute-timestamp evaluation — all solid, and `smooth_scroll = false` remains clean passthrough. But the **motion model itself makes scrolling feel much worse than either native or the commercial reference** (SmoothScroll.app): subjectively "too slow — takes too many turns of the wheel," and the measurements below show why.

## Measurements

Method: a listen-only CGEventTap at the annotated session tap (i.e. observing what applications actually receive), summing `pointDeltaAxis1` pixels — the OS fills this with its acceleration curve applied for line events, and it's the literal pixel distance for continuous events, so it's apples-to-apples "how far the page traveled." Same physical gestures repeated across three configurations, same machine/mouse/page.

**Fast gesture (three hard free-spin flicks, ~5 s):**

| Configuration | Output events | Distance delivered | Peak px/event |
|---|---|---|---|
| Native (no smoothing anywhere) | 201 line events (1,550 raw ticks) | **16,608 px** | 93 |
| OpenLogi `smooth_scroll = true` | 631 phased continuous | **9,640 px** | 128 |
| SmoothScroll.app 1.7.6 (defaults) | 180 continuous | **61,961 px** | 677 |

**Slow gentle scrolling (~10–12 s):** native ≈ 1,200 px, OpenLogi ≈ 3,000 px, SmoothScroll ≈ 6,100 px (slow gestures are harder to reproduce exactly, so treat these as indicative; the fast row is the controlled one).

Two things fall out of the fast row:

[...]

PR #1156, opened 2026-08-30
fix(macos): write the scroll point delta last so pixel precision survives

isleofgreg:
## Summary

Hardware-verification follow-up to #1141 (first of the fix series proposed there).

A scroll `CGEvent` keeps one canonical distance. `set_continuous_axis` writes the pixel/point delta first and the coarse integer **line** field after it, which makes the line count canonical: at delivery the point delta is re-derived from it at 8 points per line. Every smooth-scroll frame under 10 posted points reaches applications as **zero**, and the rest quantize to 8-point steps.

Measured on macOS 15.7 (MX Master 4, free-spin wheel, listen-only session-tap logger): a slow animated stream lost **77%** of its distance, a medium one 45%. This is most of the reason `smooth_scroll = true` currently travels less than native (#1141's 0.58× number).

The fix is a reorder: write the line and fixed-point fields first, the point delta last, so it stays authoritative. Verified with isolated synthetic-event experiments that the mangle follows field write order, not posting location (HID vs session tap both show it when the line field is written last).

## Test plan

- `cargo fmt --all --check`, `cargo clippy -p openlogi-inject --all-targets -- -D warnings` (aarch64-darwin + x86_64-pc-windows-gnu), `cargo test -p openlogi-inject`, `cargo check --workspace` — all green.
- On hardware: with this reorder, sub-10-point frames arrive intact and slow smooth-scroll streams deliver their full distance (same logger, before/after).

### DIFF
--- a/crates/openlogi-inject/src/inject/macos.rs
+++ b/crates/openlogi-inject/src/inject/macos.rs
@@ -715,9 +715,15 @@ fn set_continuous_axis(
     const POINTS_PER_LINE: i64 = 10;
     const FIXED_POINT_SCALE: i64 = 1 << 16;
     let points = i64::from(points);
-    event.set_integer_value_field(point_field, points);
+    // Write order is load-bearing: a scroll CGEvent keeps one canonical
+    // distance, and writing the coarse integer *line* field makes it the
+    // canonical value — the point delta is then re-derived from it at 8
+    // points per line, discarding pixel precision (measured on macOS 15.7:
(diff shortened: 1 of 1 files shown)


isleofgreg:
Rebased onto master (a92aa43b, v0.8.6) after the 09-19 refactor split moved scroll injection out of `inject/macos.rs`. No functional change: the write-order fix now lives in `crates/openlogi-inject/src/inject/macos/scroll.rs` (`set_continuous_axis`). Previous head for reference: eac9349a.

PR #1157, opened 2026-08-30
fix(scroll): post phaseless continuous scroll events on macOS

isleofgreg:
## Summary

Hardware-verification follow-up to #1141 (second of the fix series proposed there; independent of #1156 — they touch different functions and merge in any order).

Stamping `kCGScrollWheelEventScrollPhase` (Began/Changed/Ended) onto the synthetic smooth-scroll frames declares a trackpad **gesture** to the system, and applications act on that:

- **Rubber-band elastic overscroll** appears on plain wheel scrolling (AppKit/WebKit enable it for gesture streams) — reproduced in Brave and Apple Mail's sidebar.
- **WebKit gesture latching starves sites that scroll from JS wheel handlers** — reproduced on Fastmail's virtualized mailbox list in Safari, which stopped scrolling entirely. The `Began` frames even go out with zero delta, exactly what a latch keys off.

Native continuous wheels and established smooth-scroll utilities emit **phaseless** continuous events (captured SmoothScroll.app 1.7.6 output: phase 0 on all events). Switching our macOS injection to phaseless fixed both problems on hardware immediately, with no change in delivered distance.

The engine's balanced phase lifecycle stays the cross-platform runtime contract (Linux/Windows already inject distance only); the phase is simply no longer forwarded into the CGEvent. Zero-distance frames now have nothing left to carry, so they're dropped at the boundary instead of flooding the stream.

Known trade-off, accepted deliberately: phaseless output also means no elastic bounce in apps where trackpad users get it — identical to how the commercial reference utilities behave.

## Test plan

- `cargo fmt --all --check`, `cargo clippy -p openlogi-inject --all-targets -- -D warnings` (aarch64-darwin + x86_64-pc-windows-gnu), `cargo test -p openlogi-inject`, `cargo check --workspace` — all green.
- On hardware (macOS 15.7, MX Master 4): rubber-band gone in Brave/Mail, Fastmail scrolls normally, no regressions across several days of daily use.

### DIFF
--- a/crates/openlogi-inject/src/inject.rs
+++ b/crates/openlogi-inject/src/inject.rs
@@ -395,10 +395,13 @@ pub fn post_scroll(delta: ScrollDelta) {
 
 /// Lifecycle phase of one synthetic smooth-scroll frame.
 ///
-/// macOS forwards this state to the scroll-wheel event so applications see a
-/// balanced continuous gesture. Linux and Windows have no equivalent field;
-/// there the phase is retained by the runtime contract but only the frame's
-/// distance is injected.
+/// Part of the runtime contract — the motion engine emits a balanced
+/// lifecycle — but no backend forwards it today: injected frames carry only
+/// distance. macOS deliberately posts *phaseless* continuous events (matching
+/// wheel semantics); stamping `kCGScrollWheelEventScrollPhase` instead
+/// declares a trackpad gesture, which enables rubber-band overscroll and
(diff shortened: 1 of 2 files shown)


isleofgreg:
Rebased onto master (a92aa43b, v0.8.6) after the 09-19 refactor split. No functional change: the phaseless `post_smooth_scroll` now lives in `crates/openlogi-inject/src/inject/macos/scroll.rs`; the `inject.rs` doc-comment change is as before. Previous head for reference: 8162bcf4.

Issue #399, opened 2026-07-16
[Feature]: Hold a bound key while the button is held (push-to-talk)


theoephraim:
### Pre-flight checklist

- [x] I searched existing issues and the [Roadmap](https://github.com/AprilNEA/OpenLogi#roadmap), and this isn't already tracked.

### Problem / motivation

Every button action fires as a one-shot tap on button-down — the release edge isn't used. That means there's no way to bind a key that stays *held* while you hold the mouse button, which is what push-to-talk in Slack/Discord/Zoom listens for. Binding a `CustomShortcut` sends a single tap, so the mic opens and immediately closes.

### Proposed solution

A hold variant of the custom-shortcut action: synthesize the key-down on press and the matching key-up on release, so the chord reads as held for as long as the button is down.

The capture side already supports this — the OS hook reports both edges on all three platforms; it's the injector that only emits atomic down+up pairs.

### Alternatives considered

Options+ has no equivalent as far as I can tell. Workaround today is binding PTT to a keyboard key and not using the mouse for it.

Worth noting the main risk: a key-down that never gets its key-up jams that key system-wide, so the interrupted-capture and rebind-mid-hold paths need to force-release.

### Related area(s)

- [x] Button actions / remapping
- [x] Configuration (TOML)

### Additional context

Happy to open a PR — I have this working locally against `master`.

PR #1161, opened 2026-08-30
feat(core): add an action that holds modifiers and taps the key once

### LINKED ISSUE #1098: [Feature]: An action that holds the modifiers and taps the main key once, for Cmd+Tab / Alt+Tab style switchers
### Pre-flight checklist

### Problem / motivation

Binding a mouse button to the application switcher does not work today, and the two existing shortcut actions each fail it in a different way:

| action | modifiers | main key | result with `Cmd+Tab` |
|---|---|---|---|
| `CustomShortcut` | tapped | tapped | switcher flashes and closes; Cmd is not held |
| `HoldShortcut` (#960, v0.8.0) | held | **held** | switcher runs to the end of the list and parks there |
| `HoldModifier(u8)` proposed in #652 | held | none | nothing taps Tab |
| **missing** | **held** | **tapped once** | what the switcher actually needs |

The switcher UI needs the modifier to stay down (so the overlay stays open and releasing commits the choice) while the main key is tapped exactly once. `HoldShortcut` was built for push-to-talk (#399), where holding the whole chord is the correct behaviour; this is not a bug in it, just a shape it does not cover.

Measured on an MX Master 3 (macOS 15.7.3), listening on `kCGSessionEventTap`, holding the bound button for about three seconds:

| | Cmd held | Tab |
|---|---|---|
| `HoldShortcut` bound to `Cmd+Tab` | 2683 ms | **held the whole time** |
| the same chord typed on a keyboard | 3143 ms | tapped, 5 ms |

### Proposed solution

A new `Action` variant that holds the combo's modifiers for the lifetime of the button press and taps the combo's main key once at press time: the third leg of the family `CustomShortcut` and `HoldShortcut` already form.

Sketch, following the existing conventions:

- `Action::TapKeyHoldingModifiers(KeyCombo)`, appended at the end of the enum. The `config.toml` side is position-independent (variant names are the external tags), but the IPC side encodes bincode variant indices, so appending is the append-only recipe v28 already followed when `Action::HoldShortcut` was added, together with the `PROTOCOL_VERSION` bump and the `cargo test -p openlogi-ipc --test wire_format` update that implies.
- Reuses the press lifecycle `HoldShortcut` already has (`try_hidpp_button_down` / `try_hidpp_button_up`), so no new plumbing.
- `crates/openlogi-inject/src/inject.rs::held_keys()` currently pushes the main key into the held list unconditionally (`keys.push(HeldKey::Key(combo.key()))`); the new variant would hold only the modifiers and emit a balanced down/up for the key.
- Platform-neutral, like every other `Action`: `KeyCombo` already stores abstract modifier bits and the injection layer maps them per platform, so a Windows user picks `Alt+Tab` (the Option bit) and a macOS user picks `Cmd+Tab`. No `cfg(target_os)` in the enum.
- The GUI label can be generated from the combo the way `HoldShortcut`'s `Hold {combo}` label already is, for example `Hold ⌘, tap Tab`.

**Scope, honestly.** This covers one family of interactions: UIs that stay open while a modifier is held and commit on release. On macOS that is Cmd+Tab and Cmd+\`; on Windows and Linux, Alt+Tab; in some editors, Ctrl+Tab for the recent-files switcher. Outside that family I cannot name a use for it. It is a narrower action than `HoldShortcut`, not a more general one.

### Alternatives considered

- **Fixing `HoldShortcut` to tap the key:** no. Push-to-talk (#399, the feature it was built for) needs the key held; changing it would break that.
- **`HoldModifier(u8)` from #652:** solves half of it; nothing taps the key.
- **Sending the chord repeatedly:** advances the switcher on its own and does not keep the overlay open.

### Related area(s)

- [x] GUI
- [ ] CLI
- [ ] DPI
- [ ] SmartShift
- [x] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

### Additional context

The `Action` hold axis is already being extended one special case at a time: `HoldShortcut` (#960, merged), `HoldModifier` (#652, proposed), `HoldMouseButton6` (PR #1055, open), hold-to-repeat over HID++ (PR #467, open), plus #1046 and #1054 asking for held mouse buttons for CAD orbit. This is the keyboard-chord leg of the same axis.

It would also close a request that is currently filed away. #792 asked:

> add the ability to setup a custom shortcut for example cmd+ tab o something else

and was closed as a duplicate of #573. #573 is about shortcuts not being assignable to mouse buttons at all, and #1018 is about assigned shortcuts not firing; but even with both of those fixed, `Cmd+Tab` still will not work, because no existing action can express the hold-and-tap shape. As far as I can tell nobody has pointed that out yet.

Rick0525:
## Summary

Binding a mouse button to the application switcher cannot be expressed today. The switcher needs the modifiers held for the lifetime of the press while the ordinary key is tapped exactly once, and neither existing shortcut action has that shape: `CustomShortcut` releases the modifiers immediately, so the switcher opens and closes again at once, and `HoldShortcut` holds the ordinary key down too, so key repeat walks the switcher to its last window and parks there.

`Action::TapKeyHoldingModifiers(KeyCombo)` is the third leg of that family. It holds the chord's modifiers for as long as the physical button is down and taps the chord's ordinary key once on the press edge, which is what a hand does on a keyboard.

Measured over `kCGSessionEventTap` with the new `hold_chord` example, holding `Shift+F13` (a chord no system hotkey claims, so the reading is about the injection alone) for about three seconds under each shape, with a real keyboard press of `Cmd+Tab` for scale:

| | modifier held | ordinary key |
|---|---|---|
| `WholeChord`, the shape of `HoldShortcut` | 3023.1 ms | 3005.8 ms, held the whole time |
| `ModifiersOnly`, the shape of `TapKeyHoldingModifiers` | 3024.0 ms | 1.0 ms, tapped once |
| `Cmd+Tab` typed on a keyboard, held about as long | 3143 ms | 5 ms |

The behavioural consequence, the same example driving `Cmd+Tab` end to end, with the recently used order pinned so one step always lands on the same app:

```
holding 1000 / 3000 / 8000 ms, starting from the same app
  ModifiersOnly: one step every time
  WholeChord:    overshoots, and lands further the longer it is held
```

`HoldShortcut` is not wrong here. It was built for push to talk (#399), where holding the whole chord is the correct behaviour. This is a shape it does not cover, not a bug in it.

Fixes #1098.

## Changes

**`openlogi-core`**

- `Action::TapKeyHoldingModifiers(KeyCombo)`, appended at the end of the enum so the bincode variant indices of the existing variants do not move.
- `Action::held_combo` becomes `Action::held_chord`, returning `(&KeyCombo, HeldKeys)` instead of `&KeyCombo`. The new `HeldKeys { WholeChord, ModifiersOnly }` makes "which keys stay down" a fact the action owns rather than something the injection layer infers from the variant.
- `Action::effect` folds the new variant into `HoldShortcut`'s `Effect::HeldKey` arm, so the one-shot fallback is unchanged: a dispatcher with no release context still degrades the action to a balanced tap of the whole chord, exactly as it does for `HoldShortcut`.
- `KeyCombo::rendered_label` is split into `rendered_modifiers() -> Option<String>` and `rendered_key() -> String`, so an action that treats the two halves differently can label them separately. `rendered_label` keeps its output exactly as before, now composed from the two.
- Catalog derives: the new variant joins `HoldShortcut` in `Category::Editing` and in the keyboard icon arm. It is parameterized, so like `CustomShortcut` and `HoldShortcut` it is not a picker row.

**`openlogi-inject`**

- `press_hold(combo, keys)` takes the new `HeldKeys`. `ModifiersOnly` presses the modifiers, emits a balanced down/up for the ordinary key on the press edge, and releases the modifiers on every terminal outcome the way `HoldShortcut` already does (release, cancellation, invalidated binding, shutdown).
- The tap takes no ownership in the reference counted held-output map, so a modifier shared with another live hold is unaffected.
- New `examples/hold_chord.rs`: a manual smoke test that holds one chord for a fixed time under either shape, so the two can be measured side by side against a live event stream. `press_hold` had no equivalent of `execute`'s `inject_action` example.

**`openlogi-agent-core`**

- The two `held_combo()` call sites follow the rename. No behaviour change for `HoldShortcut`.

**`openlogi-ipc`**

- `PROTOCOL_VERSION` 29 to 30, with the version-history line, and the pinned assertion in `tests/wire_format.rs` updated.

**`openlogi-desktop` and `openlogi-ui`**

- Label: `Hold Cmd, tap Tab`, generated from the combo the way `HoldShortcut`'s `Hold {chord}` label already is. A chord with no modifier renders as `Tap {key}`.
- Icon: the keyboard glyph, alongside the other chord actions.
- Two new keys in all 23 locale files, at the same position as in `en.yml`, plus the matching assertions in the i18n suite. `be.yml` carries the English source text rather than a guessed translation, so its Belarusian arrives from Crowdin, which the header of that file names as the place to edit.

**`docs`**

- `CONFIGURATION.md` gains a paragraph next to the `HoldShortcut` one, and `config.example.toml` shows the switcher binding.

## Testing

`cargo xtask ci` on macOS 15.7.3, aarch64, rustc 1.98.0: 11 passed, 0 failed, 1 skipped.

- Pass: `rustfmt`, `typos`, `publish closure`, `shell`, `clippy`, `MSRV (cargo check)`, `rustdoc (non-GUI crates)`, `tests (macos, aarch64)`, `cargo-deny`, `clippy (windows) proxy`, `wasm (portable crates)`.
- Skipped, and therefore not claimed: `tests (linux)`, which needs a Linux host. CI's Intel macOS leg and its native Windows clippy job are likewise not covered here; the Windows job above is the cross lint the runner substitutes for it.
- `cargo test -p openlogi-ipc --test wire_format` and `cargo test -p openlogi-desktop i18n` both pass, since the wire format and the locale files both changed.
- Five new unit tests inside the `tests (macos, aarch64)` job above: `held_chord` returns the modifiers-only shape where `HoldShortcut` returns the whole chord and `CustomShortcut` returns nothing, a chord with no modifier labels as `Tap F5` rather than an empty hold, the variant round trips through TOML, and two in `openlogi-inject` pin the held-output ownership claim above from both sides, that a modifiers-only hold never owns its ordinary key while a whole-chord hold of the same chord still does. Four existing tests grew an arm for the new variant, including the persisted variant-name list and the effect lowering.

Verified on hardware, an MX Master 3 on macOS 15.7.3 with `Back = { TapKeyHoldingModifiers = "Cmd+Tab" }`, the button pressed by hand three times, watching `kCGSessionEventTap`. Both of the ways this mouse reaches the machine were measured on the same build, because they are different paths into the agent: over Bluetooth the HID++ notification arrives on the BLE link, over a Unifying receiver it arrives from a device slot on the dongle.

Over Bluetooth:

| press | Cmd held | Tab tapped | result |
|---|---|---|---|
| short | 105.0 ms | 0.1 ms | one step |
| about 1.2 s | 1224.6 ms | 0.1 ms | one step |
| about 2.6 s | 2580.2 ms | 0.1 ms | one step |

Over a Unifying receiver:

| press | Cmd held | Tab tapped | result |
|---|---|---|---|
| short | 152.1 ms | 0.1 ms | one step |
| about 1.7 s | 1747.4 ms | 0.1 ms | one step |
| about 3.6 s | 3646.9 ms | 0.2 ms | one step |

The tap stays between 0.1 and 0.2 ms on every press of both runs, while the modifier tracks the finger across a 25x range of hold durations over Bluetooth and a 24x range over the receiver. The frontmost application alternates between the two most recently used windows, which is one step every time rather than a walk down the list. The same measurement against `HoldShortcut` overshoots by more the longer the button is held. No native button event from the diverted button appears anywhere in either capture, so what is measured is the whole path from the HID++ notification to the injected chord.

The GUI label, with the three chord actions side by side on one device so the wording can be compared. Back is the new action, Forward is `HoldShortcut`, and Middle Click is `CustomShortcut`. The button inspector spells the new one out in full:

screenshot-tap-key-while-holding-modifiers

That screenshot is the dev GUI against `openlogi-agent-mock`, so the device in it is the mock MX Master 3S rather than hardware. The behaviour measurements above are from a real MX Master 3.

Not covered: Windows and Linux are lint-checked only, not run. The action is platform neutral by construction, since `KeyCombo` stores abstract modifier bits and the injection layer maps them per platform, so a Windows user binding `Alt+Tab` exercises the same code path.

### DIFF
--- a/crates/openlogi-agent-core/src/runtime.rs
+++ b/crates/openlogi-agent-core/src/runtime.rs
@@ -37,15 +37,15 @@ struct HeldShortcuts {
 
 impl HeldShortcuts {
     fn start(&mut self, press: &PressToken, action: &Action) -> bool {
-        let Some(combo) = action.held_combo() else {
+        let Some((combo, keys)) = action.held_chord() else {
             return false;
         };
         match self.by_press.entry(press.clone()) {
             std::collections::hash_map::Entry::Occupied(mut held) => {
-                held.get_mut().replace(combo);
+                held.get_mut().replace(combo, keys);
             }
             std::collections::hash_map::Entry::Vacant(slot) => {
-                slot.insert(openlogi_inject::press_hold(combo));
+                slot.insert(openlogi_inject::press_hold(combo, keys));
             }
         }
         true
--- a/crates/openlogi-agent-core/src/runtime/hook.rs
+++ b/crates/openlogi-agent-core/src/runtime/hook.rs
@@ -405,7 +405,7 @@ fn handle_key(
     };
 
     info!(keycode, action = %action.label(), "key → executing bound action");
-    let queued = if action.held_combo().is_some() {
+    let queued = if action.held_chord().is_some() {
         let queued = dispatcher.try_hook_key_down(keycode, &action);
         if queued {
             HELD_KEYS.with_borrow_mut(|keys| {
--- a/crates/openlogi-core/src/binding.rs
+++ b/crates/openlogi-core/src/binding.rs
@@ -24,7 +24,7 @@ mod value;
 #[cfg(test)]
 mod tests;
 
-pub use action::{Action, WorkflowStep};
+pub use action::{Action, HeldKeys, WorkflowStep};
 pub use action_ring::{
     ActionRingConfig, ActionRingEntry, ActionRingIcon, ActionRingLayout, ActionRingSlot,
     RingAction, RingActionError,
--- a/crates/openlogi-core/src/binding/action.rs
+++ b/crates/openlogi-core/src/binding/action.rs
@@ -187,6 +187,37 @@ pub enum Action {
     /// cancellation and shutdown. Dispatchers without a release context must
     /// degrade this action to a balanced tap rather than leave keys held.
     HoldShortcut(KeyCombo),
+    /// Hold the chord's modifiers for the lifetime of the physical press while
+    /// tapping its ordinary key exactly once, the way a hand does on a
+    /// keyboard.
+    ///
+    /// This is the shape an application switcher needs. Cmd+Tab steps one
+    /// window forward and leaves the switcher on screen for as long as Cmd is
+    /// down. Neither existing variant can express it:
+    /// [`Action::CustomShortcut`] releases Cmd immediately, so the switcher
+    /// closes again at once, and [`Action::HoldShortcut`] holds Tab down too,
+    /// so key repeat walks the switcher to its last window and parks there.
+    ///
+    /// Lifecycle-aware runtimes emit the modifier down edges plus one tap of
+    /// the ordinary key when the press starts, and the modifier up edges for
+    /// every terminal outcome. Dispatchers without a release context degrade
+    /// this to a balanced tap of the whole chord, exactly as they do for
+    /// [`Action::HoldShortcut`].
+    TapKeyHoldingModifiers(KeyCombo),
+}
+
+/// Which of a held chord's keys stay down for the lifetime of the press.
+///
+/// Both shapes hold the chord's modifiers; they differ in the ordinary key.
+/// See [`Action::held_chord`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum HeldKeys {
+    /// Modifiers and ordinary key alike stay down
+    /// ([`Action::HoldShortcut`]).
+    WholeChord,
+    /// The modifiers stay down and the ordinary key is tapped once as the
+    /// hold starts ([`Action::TapKeyHoldingModifiers`]).
+    ModifiersOnly,
 }
 
 /// One step in a [`Action::Workflow`]. A workflow is a `Vec<WorkflowStep>`
@@ -313,6 +344,12 @@ macro_rules! derive_action_core {
                     Action::Workflow(steps) => format!("Workflow ({} steps)", steps.len()),
                     Action::OpenApplication(target) => format!("Open {}", target.display_name()),
                     Action::HoldShortcut(combo) => format!("Hold {}", combo.rendered_label()),
+                    Action::TapKeyHoldingModifiers(combo) => match combo.rendered_modifiers() {
+                        Some(modifiers) => {
+                            format!("Hold {modifiers}, tap {}", combo.rendered_key())
+                        }
+                        None => format!("Tap {}", combo.rendered_key()),
+                    },
                 }
             }
 
@@ -328,7 +365,8 @@ macro_rules! derive_action_core {
                     | Action::RunAppleScript(_)
                     | Action::RunShellCommand(_)
                     | Action::Workflow(_)
-                    | Action::HoldShortcut(_) => Category::Editing,
+                    | Action::HoldShortcut(_)
+                    | Action::TapKeyHoldingModifiers(_) => Category::Editing,
                     Action::SetDpiPreset(_) => Category::Dpi,
                     Action::OpenApplication(_) => Category::System,
                 }
@@ -362,11 +400,13 @@ for_each_unit_action!(derive_action_core);
 
 impl Action {
     /// The chord whose output must remain down until the originating press
-    /// ends, or `None` for an instantaneous action.
+    /// ends, together with which of its keys stay down. `None` for an
+    /// instantaneous action.
     #[must_use]
-    pub fn held_combo(&self) -> Option<&KeyCombo> {
+    pub fn held_chord(&self) -> Option<(&KeyCombo, HeldKeys)> {
         match self {
-            Self::HoldShortcut(combo) => Some(combo),
+            Self::HoldShortcut(combo) => Some((combo, HeldKeys::WholeChord)),
+            Self::TapKeyHoldingModifiers(combo) => Some((combo, HeldKeys::ModifiersOnly)),
             _ => None,
         }
     }
--- a/crates/openlogi-core/src/binding/action_ring/icon.rs
+++ b/crates/openlogi-core/src/binding/action_ring/icon.rs
@@ -317,7 +317,8 @@ macro_rules! derive_action_icon {
                     Action::CustomShortcut(_)
                     | Action::TypeText(_)
                     | Action::Workflow(_)
-                    | Action::HoldShortcut(_) => Self::Keyboard,
+                    | Action::HoldShortcut(_)
+                    | Action::TapKeyHoldingModifiers(_) => Self::Keyboard,
                     Action::RunAppleScript(_) | Action::RunShellCommand(_) => Self::Terminal,
                     Action::OpenApplication(_) => Self::Applications,
                 }
--- a/crates/openlogi-core/src/binding/effect.rs
+++ b/crates/openlogi-core/src/binding/effect.rs
@@ -41,6 +41,12 @@ pub enum Effect<'a> {
     /// A user-recorded chord whose output is held by a lifecycle-aware
     /// runtime. A one-shot executor treats this as [`Effect::Key`] so direct
     /// dispatch remains balanced when no matching release can arrive.
+    ///
+    /// Which keys the runtime actually holds is the action's business, not
+    /// the backend's: [`Action::HoldShortcut`] holds the whole chord while
+    /// [`Action::TapKeyHoldingModifiers`] holds only the modifiers. Both
+    /// degrade to the same balanced chord tap here, so the split lives in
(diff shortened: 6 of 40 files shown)


Rick0525:
Greptile's P1 on `crates/openlogi-inject/src/inject.rs` is correct. I measured it rather than reasoning about it, so here is the reading, plus one design question I would rather not answer on my own.

## What happens

A `ModifiersOnly` hold taps its ordinary key outside the reference counted `HeldOutput::owners` map on purpose: the tap owns nothing, and its own down and up cancel out. The cost is that it also cannot see anyone else's ownership. If another live `WholeChord` hold already holds that same ordinary key down, the tap's up edge releases it.

## Measurement

macOS 15.7.3 (Sequoia), this branch at `04e9c68`, a throwaway example that starts one hold, starts a second hold one second later, and drops them in that order one second apart. Edges read off `kCGSessionEventTap`. The chords use `F13` because no system hotkey claims it, so the reading is about the injection alone.

The first hold is a `WholeChord` in every run, so its ordinary key should stay down for the full three seconds. Only the second hold changes:

| The second hold            | Same ordinary key as the first | The first hold's ordinary key stayed down for |
|----------------------------|--------------------------------|-----------------------------------------------|
| modifiers only, so it taps | yes                            | 1002.7 ms, cut short                          |
| modifiers only, so it taps | no                             | 3017.3 ms, the full hold                      |
| whole chord, so it holds   | yes                            | 3013.8 ms, the full hold                      |

Only the first row breaks. Sharing an ordinary key is not enough on its own, as the third row shows, and a tap is not enough on its own, as the second row shows. It takes both.

The edges of the failing run, `Shift+F13` held whole while `Ctrl+F13` taps. Modifiers appear as `kCGEventFlagsChanged` lines carrying the whole set held at that moment, ordinary keys appear as key down and key up, and `keycode=105` is `F13`:

```
2606.0 ms  flags changed               Shift
2629.5 ms  key down       keycode=105  Shift
3629.5 ms  flags changed               Ctrl Shift
3630.7 ms  key down       keycode=105  Ctrl Shift
3632.2 ms  key up         keycode=105  Ctrl Shift
4629.7 ms  flags changed               Shift
5641.2 ms  key up         keycode=105  Shift
5641.5 ms  flags changed               (cleared)
```

The first hold starts at 2606.0 ms and takes `F13` down at 2629.5 ms. The second hold starts at 3629.5 ms, and its tap follows at once: down at 3630.7 ms, up at 3632.2 ms. That up edge is the one that releases the `F13` the first hold still owns. The second hold ends at 4629.7 ms, and the flags go back to `Shift` alone, so the other modifier is untouched. The first hold ends at 5641.2 ms and emits an `F13` up for a key that has already been up for two seconds.

Past the truncation itself, the ownership map and the real key state disagree from 3632.2 ms onward, which is why the first hold's release emits that unpaired up.

The third run rules out the map itself. With both holds whole and sharing the ordinary key, no `F13` edge is emitted when the second hold starts or ends: the count goes 1 to 2 and back without crossing zero, and the key is held for the full 3013.8 ms. Modifiers are correct in every run, the failing one included, where the second hold's release withdraws only its own `Shift` and leaves the other modifier down. The reference counting works as designed. The tap is what bypasses it.

One loose end in the second run, which does not bear on the result: only the up edge of its `F14` tap reaches the tap, while the `F13` tap in the first run shows both edges. `F14` is a display brightness key on this machine, so its down edge is likely consumed earlier. The criterion here is how long the first hold's ordinary key stays down, and that is fully captured in every run.

To be explicit about scope: this is a new interaction that the PR introduces, since `ModifiersOnly` did not exist before it. It takes two buttons bound to hold family actions whose chords share an ordinary key, held at the same time, for example `HoldShortcut = "Ctrl+Tab"` on one and `TapKeyHoldingModifiers = "Cmd+Tab"` on another. The unit test `a_modifiers_only_hold_never_owns_its_ordinary_key` asserts only that the ordinary key stays out of the map, so it passes either way.

## The question

Consulting `owners` before the tap's up edge is a small change in one place. What the tap should emit when the key is already owned is the part I do not want to decide alone:

1. Skip the tap entirely. The other hold stays intact and the press produces nothing.
2. Emit the down and skip the up. The other hold stays intact, the application sees a key down shaped like an auto repeat, and the owner's later release still balances it.

The second option rests on a premise I have not measured: whether an application acts on a repeat down for a key it already sees as held. I can measure that before proposing either one.

I am glad to fix this in this PR, or to leave it for a follow up, whichever fits the project better.

randaljunior:
When assigning middle click to DPI Toggle, holding the button does not work—it only registers a single click.

This appears to be a bug rather than a missing feature.

PR #1167, opened 2026-08-30
feat(gui): enable supported button gestures

### LINKED ISSUE #664: [Bug]: MX Anywhere 3S DPI Toggle does not have a gesture option
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.7.1

### Operating system

Windows

### OS version & architecture

Windows 11

### Device model

MX Anywhere 3S

### Affected area(s)

- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

The DPI Toggle Button does not have an option to enable gestures unlike the original in logi Options.

### Steps to reproduce

1. Open App -> MX Anywhere 3S -> DPI Toggle
2. Scroll through list and see that there is no "gestures" option

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.7.1 (release)
- Agent: v0.7.1 (connected)
- IPC protocol: GUI 17 / agent 17
- Inventory: ready
- OS: Windows (x86_64)
- Config: schema 4 · 1 configured device(s) · thumbwheel 14

**Assets**
- Cache path: C:\Users\Lenovo\.local/share/openlogi\assets · Bundle assets: absent

**Devices (1)**
- MX Anywhere 3S — mouse (codename: MX Anywhere 3S)
  - Connection: Bluetooth (direct) · Online: yes · Battery: 50% (discharging, good)
  - Model: 2b037 (model-ids: b037/0000/0000, ext-model: 02)
  - Transports: BTLE
  - Render: mx_anywhere_3s · direct
```

### `openlogi list` output

```text

```

### Logs

```shell

```

### LINKED ISSUE #806: [Feature]: allow the DPI / ModeShift button to carry the gesture role

juan-apa:
## Summary

Enable gesture mode for every supported non-primary, non-wheel mouse control. This builds on #722, which moved Back/Forward gesture capture onto the device-specific HID++ raw-XY path. It adds the same capability for DPI / ModeShift, exposes the gesture editor consistently in the desktop UI, and makes the exclusion policy explicit for left/right click and all wheel-associated controls.

## Changes

- **Core:** centralize the gesture-source policy around Back, Forward, DPI Toggle, Gesture Button, and Haptic Panel; explicitly exclude primary clicks, Middle Click, wheel tilt, and thumbwheel controls.
- **Agent/device:** request raw-XY capture for every known DPI / ModeShift CID, avoid overwriting gesture arming with plain DPI diversion, and leave controls native when the device does not advertise the required raw-XY capability.
- **GUI:** offer **Use gestures** only for supported controls in the default profile, reject unsupported gesture-mode writes in the state layer, and cover the exact allowed/excluded button sets with tests.

## Screenshots

- Forward inspector with **Use gestures** available.
openlogi-forward-inspector-window

- Forward five-direction gesture editor after enabling gesture mode.
openlogi-forward-gestures-window

- DPI Toggle inspector with **Use gestures** available.
openlogi-dpi-inspector-window

## Testing

- `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets -- -D warnings`
- `RUSTFLAGS="-D warnings" cargo test --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent`
- `cargo xtask ci wasm`
- Ran `openlogi-agent-mock` with the development desktop app and verified that Forward exposes gesture mode, renders the five-direction editor, and DPI Toggle exposes the same gesture option.
- Not runtime-tested on physical hardware.

Supersedes #862, incorporating its DPI / ModeShift gesture-capture work into this broader supported-button implementation.

Refs #722
Fixes #664
Fixes #806

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -16,6 +16,7 @@ use openlogi_core::bindings::{button_bindings_for, hidpp_gesture_maps_for, oshoo
 use openlogi_core::config::{Config, ThumbwheelSensitivity};
 use openlogi_core::device_order::PhysicalDeviceKey;
 use openlogi_hid::DeviceRoute;
+use openlogi_hid::reprog_controls::DPI_MODE_SHIFT_CIDS;
 use openlogi_hid::session::gesture::{
     CaptureSpec, DIVERTABLE_STANDARD_BUTTONS, GESTURE_SOURCE_BUTTONS,
 };
@@ -116,14 +117,21 @@ pub fn plan_for_device(
     // gesture at once, each armed with its own raw-XY divert (the capture
     // target below derives the CIDs to divert from this map's keys).
     let gesture_bindings = hidpp_gesture_maps_for(config, Some(config_key));
-    let divert_gesture_buttons = if os_mouse_hook_available {
-        DIVERTABLE_STANDARD_BUTTONS
-            .into_iter()
-            .filter(|(_, button)| side_gesture_bindings.contains_key(button))
-            .collect::<Vec<_>>()
-    } else {
-        Vec::new()
-    };
+    let mut divert_gesture_buttons = Vec::new();
+    if os_mouse_hook_available {
+        divert_gesture_buttons.extend(
+            DIVERTABLE_STANDARD_BUTTONS
+                .into_iter()
+                .filter(|(_, button)| side_gesture_bindings.contains_key(button)),
+        );
+    }
+    if gesture_bindings.contains_key(&ButtonId::DpiToggle) {
+        divert_gesture_buttons.extend(
+            DPI_MODE_SHIFT_CIDS
+                .into_iter()
+                .map(|cid| (cid, ButtonId::DpiToggle)),
+        );
+    }
     // The HID++ gesture sources never reach the OS hook, so a non-default
     // single binding on one is deliverable only via a plain HID++ divert — but
     // only while the source is NOT in gesture mode (the raw-XY gesture divert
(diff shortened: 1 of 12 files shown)


AprilNEA (changes requested):
Two issues need fixing before merge:

1. DPI gesture mode ignores per-app single-action overrides. The capture plan builds the HID++ gesture map from global bindings even when its single-action map has an app override. I reproduced this through the real dispatcher: a per-app No action override still emits the global Actions Ring request. Gesture selection and arming need to use the effective binding for the active app.

2. The GUI offers DPI gestures without checking that the control supports raw XY. For a divertable control without raw-XY support, enabling gestures skips both gesture capture and ordinary DPI capture, silently losing the existing software action. The UI/state boundary should reject this unsupported mode and retain the binding. This case was verified with a simulated HID++ device, not a specific physical model.

The reporting-write matrix for all three DPI CIDs passed, so I did not reproduce duplicate arming or raw-XY overwrite. I will follow up with fixes and regression coverage on this branch.

AprilNEA (approved):

PR #1168, opened 2026-08-30
test(macos): isolate workspace notification observers

juan-apa:
## Summary

Stop the macOS tray lifecycle tests from affecting each other through the process-global `NSWorkspace` notification center. Under the default parallel Rust test harness, one test could deliver session notifications to the other test's observer and leave its device-I/O gate suspended.

## Changes

- Extract notification registration into a helper that accepts its notification center and workspace object.
- Keep production registration on `NSWorkspace`'s notification center.
- Give each tray lifecycle test a private `NSNotificationCenter`, preserving selector-level coverage without cross-test delivery.

## Testing

- Reproduced the original failure on the fifth repeated default-parallel tray test run.
- Ran the two tray tests 100 consecutive times under the default parallel harness after the fix.
- `RUSTFLAGS="-D warnings" cargo clippy -p openlogi-agent --all-targets -- -D warnings`
- `RUSTFLAGS="-D warnings" cargo test -p openlogi-agent`
- `RUSTFLAGS="-D warnings" cargo test --workspace --all-targets`
- Not runtime-tested on hardware; this is a test-isolation change and does not alter lifecycle transitions.

Refs #1142
Unblocks #1167

### DIFF
--- a/crates/openlogi-agent/src/tray.rs
+++ b/crates/openlogi-agent/src/tray.rs
@@ -37,7 +37,7 @@ use objc2_app_kit::{
     NSWorkspaceWillSleepNotification,
 };
 use objc2_core_graphics::{CGDisplayIsAsleep, CGMainDisplayID};
-use objc2_foundation::{NSNotification, NSString};
+use objc2_foundation::{NSNotification, NSNotificationCenter, NSString};
 use openlogi_core::brand::{self, DeeplinkCommand};
 use openlogi_core::config::AppIcon;
 use openlogi_hid::DeviceIoSignal;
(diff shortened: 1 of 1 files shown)


yuzi-co:
The two macOS test failures I've been hitting aren't area-specific — worth landing this for everyone.

Same assertion on three unrelated branches within the last hour: #1250 (`ci/narrow-nix-ci-triggers`, touches no Rust at all), and #970 / #942 (both mine, neither touches `openlogi-agent`). All three: `tray::tests::startup_stays_suspended_when_the_display_is_already_asleep`, `assertion failed: gate.allows_io()` at `tray.rs:578`.

#1236 proposes a `Mutex<()>` around the two tests instead. Both work, but this one looks like the better fix to me: a private `NSNotificationCenter` per test makes cross-test delivery structurally impossible and keeps the tests parallel, where the mutex leaves the global center shared and turns isolation into a convention every future test in the module has to remember.

Verification favours this one too — 100 consecutive runs here after reproducing the failure, versus #1236 not being runnable on its author's machine at all. For an intermittent race a single green CI run isn't much signal either way.

Issue #893, opened 2026-08-24
[Bug]: "Cmd" modifier in CustomShortcut doesn't trigger the Windows key on Windows


BrantleyF:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

7.10

### Operating system

Windows

### OS version & architecture

11

### Device model

Lift Vertical Ergonomic Mouse

### How is the device connected?

Bluetooth (direct, no receiver)

### Affected area(s)

- [ ] Device discovery / detection
- [x] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [x] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

Docs describe `CustomShortcut` as storing a "platform-neutral chord" using modifiers `shift`, `control`/`ctrl`, `option`/`alt`, `command`/`cmd`. On Windows, I'd expect `cmd` to map to the Windows key (analogous to how Electron's `CommandOrControl` works), since Win+Down is the native "minimize active window" shortcut.

In practice:
```toml
DpiToggle = { CustomShortcut = "Cmd+Down" }
```
sends only the `Down` key — the Cmd/Win modifier is silently dropped. Windows key doesn't seem to be injectable via CustomShortcut at all on Windows currently.

### Proposed solution
On Windows, map `cmd`/`command` in a `CustomShortcut` chord to a synthesized Windows-key (`VK_LWIN`) keypress, so cross-platform chords like `Cmd+Down` behave consistently on Windows.

### Steps to reproduce

Set in toml
DpiToggle = { CustomShortcut = "Cmd+Down" } or
DpiToggle = { CustomShortcut = "Win+Down" }

when button is pressed in an editor we can see that only the down button has been pressed (i.e. the page scrolls down) and not the window restored or minimized

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell

```

### macOS permissions (if applicable)

[...]

PR #1206, opened 2026-08-31
fix(windows): Use `Win` for `Cmd` CustomShortcut

kurimo:
As issue #893 reports, the Cmd or Win CustomShortcuts don't map to the Win-button. This happens because the current Windows configuration instead maps the command-button to the Ctrl-button.

I didn't find any tests that cover Windows-specific keybindings, and didn't see any easy ways to add that. This PR was only manually tested to work on a Windows with the `HaptipPanel` TOML config attached below by running `cargo run -p openlogi-agent --bin openlogi-agent` together with `cargo run -p openlogi-desktop --release`.

```toml
Up = { CustomShortcut = "Win+Up" }
Down = { CustomShortcut = "Win+Down" }
Right = { CustomShortcut = "Win+Right" }
Left = { CustomShortcut = "Win+Left" }
Click = { CustomShortcut = "Win+D" }
```

### DIFF
--- a/crates/openlogi-inject/src/inject/windows.rs
+++ b/crates/openlogi-inject/src/inject/windows.rs
@@ -284,7 +284,7 @@ fn post_custom_shortcut(combo: &KeyCombo) {
 fn combo_modifiers(combo: &KeyCombo) -> Vec<u16> {
     let mut modifiers = Vec::new();
     if combo.has_command() {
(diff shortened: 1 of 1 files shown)


kurimo:
Looks like `inject.rs` expects that Cmd and Ctrl are identical on Windows. This would break it.

PR #1211, opened 2026-08-31
fix(overlay): keep the Actions Ring above panels on Linux/Wayland

### LINKED ISSUE #1210: [Bug]: Actions Ring opens under panels/docks on Linux Wayland, and off-cursor on multi-monitor
## Summary

On Linux under a Wayland compositor (reproduced on KDE Plasma/KWin), the Actions Ring can open hidden behind a desktop panel or dock, and its position is unreliable on a multi-monitor setup — it can open on the wrong monitor entirely, or offset from the actual cursor position.

## Root cause

- **Panel occlusion**: the overlay opens the ring as a plain `WindowKind::PopUp`, which on Wayland is just another `xdg_toplevel` surface. Nothing about that guarantees it draws above a panel/dock's own always-on-top surface (typically a `zwlr_layer_shell_v1` overlay-layer surface, which genuinely is guaranteed to be on top). A panel can legitimately draw over the ring.
- **Wrong/stale position**: `openlogi_hook::cursor_position()` on Linux queries X11's `query_pointer` via XWayland. Under a native Wayland session this isn't a real Wayland pointer query and doesn't track the actual pointer — it reports a stale, display-ambiguous coordinate (confirmed live: identical coordinates returned across multiple cursor moves and clicks on different monitors).

## Steps to reproduce

1. Run OpenLogi on Linux under a Wayland session (e.g. KDE Plasma/KWin) with a visible panel/dock.
2. Bind a button to `ShowActionsRing`.
3. Press it while the cursor is under/near the panel, or on a secondary monitor.
4. Observe: the ring can render behind the panel, and/or appear on the wrong monitor or away from the actual cursor position.

## Environment

- KDE Plasma (KWin), Wayland session, multi-monitor (primary + secondary at a nonzero global X offset).

4ni1ak:
## Summary

On Linux under a Wayland compositor (reproduced live on KDE Plasma/KWin), the Actions Ring could open hidden behind a panel/dock, and its position was unreliable on a multi-monitor setup.

Fixes #1210

## Root cause

- **Panel occlusion**: the ring opened as a plain `WindowKind::PopUp`, which on Wayland is just another `xdg_toplevel` surface — nothing guarantees it draws above a panel/dock's own always-on-top `zwlr_layer_shell_v1` overlay-layer surface.
- **Wrong/stale position**: `openlogi_hook::cursor_position()` on Linux falls back to XWayland's X11 `query_pointer` under a Wayland session. That isn't a real Wayland pointer query and doesn't track one — confirmed live, it returned the exact same coordinate across multiple cursor moves and clicks on two different monitors.

## Changes

`crates/openlogi-overlay/src/ring.rs`:
- On Linux/Wayland, `open_ring` now opens an invisible full-screen `Layer::Overlay` host window first, then anchors the ring to it as a `WindowKind::AnchoredPopup` (center anchor + center gravity on the real cursor point). The overlay layer is guaranteed above everything else, including panels. Falls back to the previous plain `WindowKind::PopUp` at the old best-effort position guess on macOS, Windows, Linux/X11, or a Wayland compositor without `zwlr_layer_shell_v1`.
- The host's real cursor position comes from the compositor's own first pointer-motion delivery to it (`gpui::Window::mouse_position`, fed by native platform pointer events) rather than the X11 query — accurate regardless of which monitor the cursor is on. Bounded at 250ms, falling back to the old best-effort guess if the compositor never delivers one, so a ring still opens rather than hanging.
- Every dismissal path (slot click, cancel click, root click, click-away, the display-lifetime timeout) now closes the host window alongside the ring, so it never lingers as an invisible full-screen click-blocker.

`crates/openlogi-overlay/src/session.rs`:
- `dismiss_click_away` (previously reachable only via the macOS-only native click-away monitor) now also closes the ring's host window when it fires.

Side effect worth calling out: clicking the invisible host — i.e. anywhere outside the ring — now dismisses it on Linux too. Linux never had click-away-to-dismiss before this; the existing mechanism (a native OS click monitor) is macOS-only.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo fmt --all -- --check
cargo clippy -p openlogi-overlay --all-targets -- -D warnings   # RUSTFLAGS=-D warnings
cargo test -p openlogi-overlay   # 17 passed
```

**Runtime-verified live** on this machine: KDE Plasma (KWin) Wayland session, dual monitor (primary + a secondary at a nonzero global X offset). Iterated through several rounds against the real compositor — confirmed the ring now stays above the panel and opens correctly centered on the cursor on both monitors, including the timeout fallback path (observed firing once when the compositor was slow to deliver the first pointer event).

**Not tested**: macOS, Windows, Linux/X11, or another Wayland compositor (GNOME/Mutter, Hyprland, Sway) — the fallback path is unchanged from the pre-existing behavior on all of those, but I don't have hardware/sessions for any of them to confirm.


4ni1ak:
Fixed everything, pushed `d0958712`:

- **CI (macOS/Windows)**: my local gate only exercised the Linux cfg branch (`cargo clippy -p openlogi-overlay` on this machine), so the non-Linux stub's unused `oneshot` import and `unused_async` never showed up until CI's `cargo clippy --workspace`/`cargo test --workspace` compiled that branch too. Gated the import to Linux and documented+silenced the async lint on the stub (it's async only to match the call site's unconditional `.await`).
- **Greptile Issue 1 (popup-failure leaks host)**: confirmed — closing the host explicitly on that error path now, since no `RingView` exists yet for any dismissal path to find it.
- **Greptile Issue 2 (fallback uses wrong anchor)**: confirmed — the 250ms no-pointer-motion fallback was reusing `ring_placement()`'s global-coordinate origin as if it were already host-local, exactly the bug the host exists to avoid, reintroduced on a secondary display. Converts it the same way the plain-`PopUp` fallback path does now.

Re-ran the local gate (fmt/clippy/test) green on the final tree. Still no macOS/Windows toolchain here to verify those two CI lanes directly, but the fix is mechanical (cfg-gate the import, document the async-signature mismatch) — will watch CI.

Issue #431, opened 2026-07-20
[Feature]: Are keyboard remapping, shell-command actions, and wider keycode coverage planned?


costantinoai:
### Pre-flight checklist

- [x] I searched existing issues and the [Roadmap](https://github.com/AprilNEA/OpenLogi#roadmap), and this isn't already tracked.

Partial overlap, noted for triage: #101 and #64 touch keyboard remapping, #258 touches custom/macro actions. Neither covers the keycode gap below, and none of the three has an answer on whether the feature is planned. Happy to have this closed as a duplicate if you'd rather answer in those threads.

### Problem / motivation

I'm evaluating a move from Solaar to OpenLogi on Ubuntu 24.04 / X11, with an MX Master 3 and an MX Keys S (both Bluetooth-direct). Most of my setup transfers — per-app profiles via `WM_CLASS`, DPI, SmartShift, Back/Forward/Gesture bindings all work as documented.

Three things I currently rely on have no equivalent in OpenLogi, so I can't complete the migration. Listing them here to ask whether they're on the roadmap.

**1. Keyboard key remapping**

`ButtonId` in `crates/openlogi-core/src/binding.rs` enumerates mouse buttons only. My MX Keys S has seven keys remapped in Solaar — Lock PC, Calculator, Voice Dictation, Emoji Panel, Snipping Tool, Mic Mute, and the App-Contextual-Menu key — and there's no place in the config schema to express any of them. Same for fn-swap and backlight timeout, which Solaar exposes over HID++.

**2. Running a command from a binding**

Two of my bindings shell out (Solaar's `Execute:` rule action). The `Action` enum has no variant for this.

**3. Keycode coverage in `CustomShortcut`**

`macos_vk_to_linux` in `crates/openlogi-inject/src/inject/linux.rs` maps A–Z, 0–9, punctuation, Tab/Space/Esc/Enter/Backspace, F1–F12, Home/End/PgUp/PgDn/Delete, and the arrows. `modifiers_to_keycodes` in the same file handles Ctrl, Shift, and Alt, with `MOD_CMD` folding into `KEY_LEFTCTRL`.

That leaves these unreachable from a binding:

- Numpad keys — I need `KP_Enter` and `Ctrl+KP_Enter` (RStudio binds these separately from main-row Enter, so `Enter` isn't a substitute)
- F13–F24 — I need `F16` for a LibreOffice macro
- Print Screen as a bindable key
[...]

PR #1224, opened 2026-09-01
fix(linux): declare F13-F20 in uinput key capabilities

costantinoai:
## Summary

`CustomShortcut` bindings on F13–F20 never reach the compositor on Linux. `hid_usage_to_linux` maps HID usages `0x68..=0x6f` to `KEY_F13`..`KEY_F20` and `validate_keyboard_usage` accepts them, but `KEY_CAPABILITIES` declares only F1–F12. A uinput device cannot emit a key it has not declared, so the kernel drops the event: the binding parses, validates and resolves, then does nothing, with no error on any path.

The mapping is already covered by a test asserting `hid_usage_to_linux(0x6f) == KEY_F20`. Only the capability declaration was missing.

## Changes

- `openlogi-inject`: declare `KEY_F13`–`KEY_F20` in `KEY_CAPABILITIES` (`inject/linux.rs`).

## Testing

- `cargo fmt --check -p openlogi-inject`
- `cargo clippy -p openlogi-inject --all-targets -- -D warnings`
- `cargo test -p openlogi-inject` — 10 passed

Verified on hardware (Ubuntu 24.04, X11, MX Master 4 for Business over Bluetooth) by decoding the injector's capability bitmap from `/proc/bus/input/devices`:

- v0.8.3 packaged build: F1–F12 present, F13–F20 absent. A button bound to `F18` produced no event.
- This build: F13–F20 present.

Refs #431


AprilNEA (approved):
good catch

Issue #1148, opened 2026-08-29
[Feature]: Comprehensive action catalog for button/key remapping (Apps, Brightness, Audio, Screenshots, Custom Shortcuts)


nickdesi:
### Which part of OpenLogi?
GUI (desktop app) & Agent

### Operating system
All platforms (macOS, Windows, Linux)

### Summary / Problem Statement
When configuring button and key assignments in the device view, the current action picker only presents a limited subset of hardcoded actions (Mouse Clicks, Basic Navigation, and simple Edit commands like Copy/Paste).

Essential day-to-day actions commonly used on companion software (like Logi Options+ or Karabiner) are missing from the standard picker UI, even though several of them are already defined in the backend `Binding` enum.

### Proposed Action Catalog

To provide a complete and intuitive experience across mice and keyboards (such as MX Master and MX Keys series), we propose organizing the action picker into structured, comprehensive categories:

#### 1. System & Hardware Controls
- **Display Brightness**: Brightness Up, Brightness Down
- **Sound & Audio**: Volume Up, Volume Down, Mute/Unmute Audio, Mic Mute
- **Screen Capture**: Full Screenshot, Capture Region / Snipping Tool
- **Power & Security**: Lock Screen, Sleep Display / System Sleep
- **Media Playback**: Play/Pause, Next Track, Previous Track

#### 2. Navigation & Window Management
- **macOS / OS Windowing**: Mission Control, App Exposé, Show Desktop
- **Workspaces**: Switch to Previous Desktop / Space, Switch to Next Desktop / Space
- **Window Controls**: Minimize Window, Fullscreen / Maximize

#### 3. Applications & Custom Automations
- **Open Application**: Native file/app picker (`.app` on macOS, `.exe` on Windows, `.desktop` / binary on Linux) or path input.
- **Custom Keyboard Shortcut**: Ability to record and trigger arbitrary key combinations (e.g., `Cmd+Ctrl+Shift+4`, `Ctrl+Alt+T`).
- **Scripts & Commands**: Run Shell Command / AppleScript for advanced user workflows.

### User Value
This would make OpenLogi a complete, self-sufficient replacement for proprietary companion tools, allowing users to easily customize function rows, extra mouse buttons, and numpad auxiliary keys directly from the GUI without needing auxiliary remappers.

PR #1229, opened 2026-09-01
feat(cli): add `openlogi haptic` and the full 0x19b0 waveform set

costantinoai:
## Summary

`HapticWaveform` and `SupportedWaveforms` each modelled two of the sixteen waveforms the `0x19b0` feature defines, and nothing outside the Actions Ring could reach haptics at all — intensity was never read or written, and there was no way to play a waveform.

This adds the remaining fourteen waveform IDs, exposes get/set of the device-wide configuration alongside the existing `play_haptic`, and adds an `openlogi haptic` subcommand modelled on `openlogi backlight`:

```
openlogi haptic                 # config + advertised waveforms
openlogi haptic level 80        # intensity 0-100, 0 silences
openlogi haptic play knock
```

No config schema change, so a stock build still reads configs written alongside this.

## Changes

- `openlogi-hidpp`: extend `HapticWaveform` and `SupportedWaveforms` from 2 to 16 entries. `WHISPER_COLLISION` is `0x1B`, so the mask is not contiguous.
- `openlogi-device`: add `get_haptic_config`, `get_haptic_capabilities` and `set_haptic_intensity` beside `play_haptic`, following the same `with_route` pattern.
- `openlogi-hid`: route wrappers, mirroring `get_backlight` / `set_backlight_enabled`.
- `openlogi-cli`: `haptic` subcommand plus four parse tests.

## Testing

- `cargo fmt --check`
- `cargo clippy -p openlogi-cli -p openlogi-device -p openlogi-hid -p openlogi-hidpp -p openlogi --all-targets -- -D warnings`
- `cargo test -p openlogi-cli -p openlogi-hidpp -p openlogi-device -p openlogi-hid` — 472 passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` on the four crates

Verified on hardware: MX Master 4 for Business (`046d:b048`) over Bluetooth, Ubuntu 24.04.

The device's own capability mask decodes as all sixteen, which confirms the bit positions:

```
supported waveforms: SupportedWaveforms(SHARP_STATE_CHANGE | DAMP_STATE_CHANGE |
  SHARP_COLLISION | DAMP_COLLISION | SUBTLE_COLLISION | HAPPY_ALERT | ANGRY_ALERT |
  COMPLETED | SQUARE | WAVE | FIREWORK | MAD | KNOCK | JINGLE | RINGING |
  WHISPER_COLLISION)
```

All sixteen were played on the device and are physically distinct. Intensity round-trips: 60 → 25 → 100 → 60, each read back.


costantinoai:
One limitation worth recording, and a question about whether a follow-up would be welcome.

**`haptic level` does not survive a power cycle.** `setConfig` writes to device RAM, so the intensity reverts to the firmware default whenever the mouse is switched off and on. The subcommand says so in its output, but nothing re-asserts it.

Nothing in OpenLogi persists it either: `reapply_volatile_settings` re-pushes wheel resolution/inversion, DPI, SmartShift and lighting on every connect, but not haptic intensity, and the only haptic field in `config.toml` is `action_ring.haptics` (a bool). So the value is unreachable from config today.

Making it sticky looks like it would follow the existing pattern exactly — a per-device config field plus one more `effective_*` lookup in `reapply_volatile_settings` beside `dpi` and `smartshift`.

I deliberately left that out of this PR for two reasons. It would add config-schema surface, and as it stands this change adds none — a stock build still reads a config written alongside it, which seemed worth preserving. And the field's shape is a design decision rather than a mechanical one: top-level per-device versus under `action_ring`, and how it should interact with the existing `haptics` bool.

Would a follow-up adding persistence be welcome, and if so which shape do you prefer? Happy to leave it alone if this is better designed alongside #1148.

Issue #1205, opened 2026-08-31
[Bug]: OpenLogi overrides DPI settings after every reboot


SilentDecode:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.8.3

### Operating system

Linux

### OS version & architecture

Arch x86_64

### Device model

MX Master 3

### How is the device connected?

Bluetooth (direct, no receiver)

### Affected area(s)

- [ ] Device discovery / detection
- [ ] Button remapping
- [x] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [x] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

Every time I log into my Linux install (after password prompt), my mouse settings are different than the one I set in the OS itself. 

I only use OpenLogi for the ratchet function, because I hate freewheel. I have not touched any settings in OpenLogi except the "permanent ratchet" function. Very annoying that my perfected OS mouse settings are overwritten every time with annoying new settings I can't finetune enough (also the reason I have "autostart" turned off). There is no option in OpenLogi to also disable mouse acceleration, which I have always turned off. 

PS:
There is no Options+ for Linux as stated above. As far as I know there is no Linux version of Options+, but I couldn´t make this bugreport without ticking that box. The only OS on this hardware, is the current installation of Arch. No Windows to be seen here. 

### Steps to reproduce

1. Install OpenLogi
2. Reboot

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.8.3 (release)
- Agent: v0.8.3 (connected)
- IPC protocol: GUI 29 / agent 29
- Inventory: ready
- OS: Linux (x86_64)
- Locale: en-GB (UI: follow system)
- Accessibility: granted · Input hook: not installed
- Launch at login: no · Menu bar: yes · Update check: off
- Running from: source build (dev)
- Config: schema 7 · 2 configured device(s) · thumbwheel 14

[...]

Issue #1191, opened 2026-08-31
[Bug]: Brio Webcam is listed twice


stoneynutcase:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.8.3

### Operating system

Linux

### OS version & architecture

Omarchy 4.0.2-1

### Device model

Logitech Brio Webcam

### How is the device connected?

Wired (USB cable)

### Affected area(s)

- [x] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [x] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [ ] Other

### What happened?

The device is shown twice in the UI, only one of them has configuration when clicked.
I know 2 video devices are available in `/dev`, I believe one of them is the IR sensor.

<img width="1363" height="700" alt="Image" src="https://github.com/user-attachments/assets/664612c1-db3b-4306-a41f-8a54349e7de8" />

<img width="506" height="264" alt="Image" src="https://github.com/user-attachments/assets/e9b4aae1-7cf9-4205-b12a-5e06efcf4318" />

<img width="786" height="485" alt="Image" src="https://github.com/user-attachments/assets/ef315ade-9966-40e1-915f-a2da6e69dcb4" />

### Steps to reproduce

1. Buy a Logitech Brio webcam
2. Connect it to your (Linux) machine that has OpenLogi installed
3. Open OpenLogi
4. The webcam is listed twice

### Diagnostics report

```markdown

```

### `openlogi list` output

```text

```

### Logs

```shell
(inventory read from the running agent)
Unifying Receiver (E3B10D61, vid=046d pid=c52b)
  ├─ slot 1 ● MX Keys Wireless Keyboard (keyboard, wpid=408a, battery=50% good (discharging))
  │       model_ids=[b35b,408a,0000] ext=00 serial=— unit_id=a8557bbc transports=equad+btle
[...]

PR #1236, opened 2026-09-01
fix(agent): serialize the macOS tray tests that share NSWorkspace's center

4ni1ak:
## Summary

Two macOS-only unit tests in `crates/openlogi-agent/src/tray.rs` — `startup_stays_suspended_when_the_display_is_already_asleep` and `overlapping_suspend_sources_all_clear_before_device_io_resumes` — each register an `ActivityTarget` observer on `NSWorkspace::sharedWorkspace().notificationCenter()`, a process-global singleton. A notification either test posts is delivered to every live observer registered on that center, this test module's included, regardless of which test posted it. Run in parallel (Rust's default test runner), one test's posted notifications can flip the other's device-IO gate mid-assertion.

Not a hypothetical: two independent, unrelated PRs (#1191 and #1205) each hit the exact same failing assertion in `startup_stays_suspended_when_the_display_is_already_asleep` on `tests (macos, arm64)`, and both passed cleanly on an unmodified retry. Neither PR touches `tray.rs`. That's the signature of this race, not a real regression in either PR — but it means every PR pays an intermittent, unrelated CI failure until someone happens to hit a lucky interleaving.

## Changes

`crates/openlogi-agent/src/tray.rs`: added a module-private `workspace_notifications_lock()` — a `Mutex<()>` acquired at the top of each affected test and held for its body, so the two tests can no longer run concurrently with each other. Doesn't touch the notification-observer logic under test, only forces the two tests calling into the shared singleton to run one at a time.

## Testing

Can't compile or run this locally — `mod tray` is `#[cfg(target_os = "macos")]`-gated and this machine has no macOS toolchain. Verified the change parses/formats correctly with `cargo fmt -p openlogi-agent -- --check` (clean) and by careful manual review — the added code is plain `std::sync::Mutex`, no macOS-specific API surface, and `Mutex`/`PoisonError` are already imported in this file. Will watch the `tests (macos, arm64)` / `tests (macos, x86_64)` CI jobs on this PR for the real verification.

### DIFF
--- a/crates/openlogi-agent/src/tray.rs
+++ b/crates/openlogi-agent/src/tray.rs
@@ -502,8 +502,21 @@ mod tests {
     use super::*;
     use openlogi_hid::device_io_channel;
 
+    /// `NSWorkspace::sharedWorkspace()` and its `notificationCenter()` are a
+    /// process-global singleton: a notification posted by one test's
+    /// `install_activity_observer` call is delivered to every other live
+    /// `ActivityTarget`, this test module's included, regardless of which
+    /// test posted it. Rust's default parallel test runner would otherwise
+    /// let these tests corrupt each other's gate state through that shared
+    /// center — this held for a test's whole body serializes them.
+    fn workspace_notifications_lock() -> std::sync::MutexGuard<'static, ()> {
+        static LOCK: Mutex<()> = Mutex::new(());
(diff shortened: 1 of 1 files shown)


4ni1ak:
Superseded — the same fix landed directly on master in 86e9997 (same root cause, one shared `Mutex` instead of a per-test lock function, otherwise identical). Closing.

PR #1259, opened 2026-09-04
feat(device): G-series gaming button remap via HID++ 0x8110 spy

CharlieMc0:
## Summary

- G502 X PLUS (and other HID++ `0x8100`/`0x8110` G-series gaming mice) currently get no Buttons tab: the existing remap stack is built entirely on `0x1b04 ReprogControlsV4`, which these devices never report.
- Adds a second, independent capture mechanism using `0x8110 MouseButtonSpy` — a raw event **tap**, not a divert — restricted to the four buttons confirmed on real hardware to carry no native OS report (DPI Shift, Profile Cycle, DPI Up, DPI Down). The seven standard buttons (left/right/middle/side/wheel-tilt) keep their existing native/`0x1b04` behavior untouched.
- Deliberately does **not** implement Host mode, on-device profile writes, or `setMouseButtonMapping` suppression — confirmed unnecessary by hardware testing (see Testing) and out of scope per the maintainer's preference that OpenLogi never manage on-device profiles.

## Changes

- **`openlogi-hidpp`**: read-only `0x8100 OnboardProfiles` (`getDescription`/`getMode`) and `0x8110 MouseButtonSpy` (`getMouseButtonCount`/`start`/`stopMouseButtonSpy`/button-state event) codecs. Protocol layout is reverse-engineered — cross-checked against libratbag and `cvuchener/hidpp`, sourced in the doc comments.
- **`openlogi-device`**: new `MouseButtonSpy` session accessor (mirrors `ReprogControlsV4`); `session::gesture::spy` submodule with `SPY_BUTTON_MAPS` (one verified model: G502 X PLUS, `config_key` `"04099"`) and a pure `spy_edges` mask-diff decoder. `CaptureSpec` gains an independent `spy_buttons` field (not a variant of a new enum — the two capture mechanisms stay independently representable for a future device that has both); `ArmedControls` arms/rearms/restores the spy alongside the existing mechanisms. The spy's event receiver is registered via `listen()` *before* `start_reporting()` is awaited, since the underlying emitter never buffers and a receiver created afterward could silently miss early presses.
- **`openlogi-core`**: `ButtonId::{DpiShift, ProfileCycle, DpiUp, DpiDown}` + `GAMING_BUTTONS` (kept out of `ALL`, same reasoning as `KEYBOARD_KEYS`); `Capabilities::gaming_buttons` widens the existing `buttons` tab-gate so no GUI-gating code needed to change.
- **`openlogi-ipc`**: wire bump to `PROTOCOL_VERSION` 30 for the new `Capabilities` field; golden regenerated.
- **`openlogi-agent-core`**: `capture_plan::plan_for_device` now takes the device's model key (a `DeviceCaptureIdentity` struct, to stay under the arg-count lint) and arms the spy only when a gaming button is actually bound — an untouched device is never told to stream events.
- **`openlogi-desktop`**: `FallbackControls` (replacing a lone `thumbwheel: bool`) drives the synthetic-silhouette fallback; on a gaming-capable device it swaps the `DpiToggle`/`GestureButton` hotspots (which don't exist on this hardware) for the four gaming buttons.
- **`crates/openlogi-ui/locales/*.toml`**: four new `actions.*` keys across all 23 catalogs (English placeholder pending translation).
- `openlogi diag mouse-buttons` (from a preceding commit) is the diagnostic tool this was designed against — `--watch` streams live spy events for verifying a new model's button map.

## Testing

```
cargo fmt --all -- --check
cargo test --workspace
cargo test -p openlogi-ui locale
cargo test -p openlogi-desktop i18n
```
All green on macOS (full workspace, including the GPUI crates).

Hardware-verified on a real Logitech G502 X PLUS (Lightspeed receiver):
- `openlogi diag mouse-buttons` / `--watch` confirmed the button↔index map and that the spy streams cleanly in the device's default Onboard mode (no mode switch needed).
- Ran the dev build against the device: Buttons + Actions Ring tabs appear; bound Profile Cycle → an action fires correctly on press; standard buttons (left/right/middle/side/tilt) confirmed unchanged, no double-firing.

Not yet independently re-verified on this exact PR diff by the maintainer: DPI Up vs. DPI Down physical index order (5/10/11 swap risk — cosmetic label only, no data-migration risk if wrong), sleep/reconnect rearm, and agent-crash recovery. All three are inexpensive to check and don't gate review.

## Deferred (tracked as follow-up, not built)

- Host mode / on-device profile memory writes / `setMouseButtonMapping` suppression for the other 7 buttons.
- Additional `SPY_BUTTON_MAPS` entries for other G-series models.
- Real product-photo asset + Logitech slot-name metadata for the G502 X PLUS (the synthetic silhouette is used today; a naive image-only fix would actually *regress* the four gaming-button hotspots, since the asset-hotspot path has no slot-name vocabulary for them — confirmed by reading `map_slot_name`).

https://claude.ai/code/session_01L33m1b83ZgermHZPNCmg53


CharlieMc0:
Closing for now — not ready to send upstream yet, keeping this in my fork only.

PR #1268, opened 2026-09-04
feat(scroll): pulse-based smooth-scroll motion with step, duration, and acceleration

isleofgreg:
## Summary

Third PR of the series proposed in #1141 (hardware verification of the v0.8.x smooth scroll on an MX Master 4, macOS 15.7). This is the motion model itself plus its config surface. It is **stacked on #1156 and #1157** — the first two commits here are those PRs verbatim, and this PR will be rebased down to the remaining six once they land (or merge in any order; the diffs are independent).

What changes, in commit order after the two fixes:

- **Pulse-superposition motion model** (`runtime/scroll.rs`). Each accepted wheel tick becomes a finite pulse (Herf two-phase curve: damped-force head, viscous tail, C¹ joint, exact asymptote); a source's position is the sum of its live pulses evaluated at absolute timestamps. Overlapping pulses compound successive flicks into a glide, so momentum is the pulse tail rather than a separate mechanism. Upstream's invariants are preserved: finite motion, net distance conserved, no unbounded velocity state. Ticks arriving within a frame coalesce; a hard cap of 64 live pulses keeps per-frame cost trivial in free-spin bursts.
- **Config surface** (`openlogi-core` settings, nutype newtypes with serde defaults, so `smooth_scroll = true` alone keeps working): `smooth_scroll_step` (1–20), `smooth_scroll_duration_ms` (80–1000, default 360), `smooth_scroll_acceleration` (1–10 gain cap, default 7). `ScrollPreferences` widens to a single-snapshot `AtomicU64` so the worker always reads a consistent tuple. Live reload rides the existing `reload_config` path.
- **Acceleration keyed on wheel rate, not event gap.** A high-resolution wheel reports every few milliseconds regardless of speed, so interval-keyed gain saturated even for feather-light input. Distance over a 70 ms sliding window reproduces the classic `(1 + 30/dt)/2` curve for notched wheels and scales with real wheel speed for hi-res streams.
- **Gain 1 for OS-hook input on macOS.** The event-tap hook's line deltas already carry the OS acceleration curve (a hard spin arrives as ~1,400 line-units), so configured gain on top was acceleration². Diverted HID++ wheels and the raw-delta hooks on other platforms keep the configured cap.
- **Per-platform step default.** Follow-on from the point above and the open question in #1141 §2: the same step means different travel on pre-accelerated vs raw-count input. Measured: step 4 on macOS hook input lands at parity with the 90 px commercial reference that step 9 reproduces on raw ticks (fast-gesture benchmark ~66k px vs the reference's ~62k). Default is 4 on macOS, 9 elsewhere; the key keeps one meaning (per-tick multiplier). I went with the simplest option so the defaults feel right out of the box on every platform — happy to switch to input normalisation or a HID++ wheel divert on macOS if you'd rather have uniform semantics; that's the one design call in here I'd like your read on.

Measured against SmoothScroll.app on the same hardware: subjectively at parity or better after a multi-day daily-driver soak with the reference uninstalled, and measurably smoother (we frame at ~100 events/s vs its ~30). Full measurement write-up in #1141.

Not in this PR by design: GUI presets (follows once the core settles), per-app exclusions, keyboard smoothing.

## Test plan

- `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings` on aarch64-apple-darwin, plus `x86_64-pc-windows-gnu` for the touched crates; `RUSTDOCFLAGS=-D warnings cargo doc` for `openlogi-core`/`openlogi-agent-core`/`openlogi-inject`; `cargo test` for those three crates plus `openlogi-agent` — all green (232 + 272 + 11 tests in the scroll crates; motion model and config newtypes are deterministic and unit-tested, including pulse-curve monotonicity/normalisation, net-distance conservation, coalescing/cap behaviour, windowed acceleration, the macOS gain-1 path, and config round-trip/range rejection).
- On hardware (MX Master 4, free-spin and ratchet, macOS 15.7): listen-only CGEventTap capture of delivered distance for feather/medium/hard streams before and after; daily-driver soak on the branch since 2026-08-29.


isleofgreg:
Pushed `4689b4b9` addressing the review:

- **Merged ticks keep their own timing.** A merge (frame window or pulse cap) no longer adds amplitude to the older pulse and inherits its clock. It settles what that pulse has delivered, then restarts its remainder together with the new tick from the tick's own arrival and with the tick's own duration — position continuous, net distance conserved, no partial-progress start, no early finish at an older deadline, and a live duration change takes effect on the very next tick. The frame window is measured from the pulse's first tick (new `anchor` field) so a steady stream still hands off to a fresh pulse each frame rather than restarting one pulse forever.
- **Acceleration windows are per axis.** Each axis keeps its own rate; a fast spin on one no longer raises the gain of slow movement on the other.

Three new deterministic tests: a tick merged past the pulse cap into a 400 ms-old pulse animates for its full 1000 ms and delivers nothing of itself at arrival; a duration reload applies to a tick merged within the frame; a slow horizontal tick sharing a source with a fast vertical ramp gains nothing.

isleofgreg:
`ea913aea`: a frame-window merge now happens only into a pulse running the same duration; a tick accepted under a freshly reloaded duration gets its own pulse, so a reload never re-times motion accepted under the previous configuration. Only the pulse-count bound still merges unconditionally (documented as the model's one approximation — it still conserves distance and continuity). Test updated accordingly: after a reload inside the frame window the source holds two pulses ending at 100 ms and 504 ms.

isleofgreg:
The `tests (macos, x86_64)` failure on `ea913aea` is `tray::tests::startup_stays_suspended_when_the_display_is_already_asleep` in `openlogi-agent` (`assertion failed: gate.allows_io()`), which this PR doesn't touch — its only `openlogi-agent` change is one argument in `pairing/tests.rs`. It's the parallel-execution flake #1236 addresses: it accounts for 9 of the last 13 CI failures on this repo across unrelated branches (`fix/camera-depot-render-without-metadata`, `fix/direct-zero-unit-persistence`, `fix/windows-open-config-folder`, `ci/narrow-nix-ci-triggers`, …), and passes serially every time locally. I can't rerun jobs here, so I've re-pushed the same tree under a fresh commit to get another run.

isleofgreg:
Rebased onto master (a92aa43b, v0.8.6) after the 09-19 refactor split; still stacked on #1156 + #1157. No functional change: `runtime/scroll.rs` adopts master's `FRAME_INTERVAL` → `FRAME_PERIOD` rename, the config round-trip test moved to `config/tests/app_settings.rs`, the scroll-preferences reload test to `orchestrator/tests/capture_plans.rs`, and the injector hunks to `inject/macos/scroll.rs`. Each commit compiles and tests on its own. Previous head for reference: b3fe55b5.

isleofgreg:
Rebased onto master (e3bf24a4) after #1566 (v0.8.7); still stacked on #1156 + #1157. The only conflict was the `pub use settings::{…}` list in `openlogi-core/src/config.rs`, resolved as the union of `MouseProfileTarget` and the `SmoothScroll*` types. Every other patch is byte-identical (range-diff), each commit compiles on its own, and fmt, clippy `-D warnings` and `cargo test --workspace --all-targets` are green locally. Previous head for reference: e3d051bd.

PR #1269, opened 2026-09-04
fix(scroll): cool down and compress quick direction reversals of pre-accelerated input

isleofgreg:
## Summary

Fourth PR of the #1141 series. **Stacked on #1268** (which is stacked on #1156/#1157); only the last three commits are new here, and this will be rebased once #1268 lands.

The macOS scroll-acceleration curve is keyed on *unsigned* wheel speed and stays hot across a direction flip. Measured on an MX Master 4: an opposing tick arriving 68 ms after a hot stream came in at **14× the cold-start ramp** (the curve decays somewhere between ~70 and ~435 ms). Multiplied by the smooth-scroll step, a quick corrective flick after a fast spin overshot badly — the one soak complaint the motion model in #1268 had left.

Two measures, both confined to the pre-accelerated (macOS OS-hook) input path so raw-count sources are untouched:

- **Reversal cooldown** — on a direction flip within the hot window, re-impose the measured cold-start ramp on the new direction instead of trusting the still-hot OS curve. Direction state is tracked per axis as a signed integer (their lint set denies `float_cmp`).
- **Soft-knee compression of corrective bursts** — the burst that follows a reversal is compressed above a knee so a small correction stays small while a deliberate reverse spin still reaches full travel. Tune knobs are the `REVERSAL_*` constants in `runtime/scroll.rs`.
- A `chore` commit adds trace-level logging of impulses and cooldown decisions; it's what the measurements above were taken with and is off unless `openlogi_agent_core=trace`.

Soak verdict on this build (daily driver since 2026-08-29): the reversal overshoot is gone and the model is otherwise unchanged.

## Test plan

- Same gate as #1268: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings` (aarch64-apple-darwin + x86_64-pc-windows-gnu for the touched crates), `RUSTDOCFLAGS=-D warnings cargo doc`, `cargo test -p openlogi-agent-core` — all green. Five new deterministic tests cover the cold restart on a quick reversal, no cooldown for a resumed direction, compression above/below the knee, and that raw-count sources never enter the path.
- On hardware: fast spin → immediate corrective flick, before/after, with the trace commit's logging.


isleofgreg:
Re-stacked on the updated #1268 (which now carries the merge-timing and per-axis-acceleration fixes flagged here as well), plus `3ec4255f` for the cooldown finding:

- **Reversal state outlives the animation.** It now lives per source on the engine beside the animations instead of on the pulse set, so a source retired when its last pulse completes (360 ms default, 80 ms minimum — both shorter than the 400 ms cooldown) still cools an opposing tick that arrives while the OS curve is hot. State expires once no tick has arrived for a full cooldown; by then any flip is cold and starts fresh, which is exactly unscaled (this also corrects a test that had a cold flip still compressed — a leftover from the state being tied to pulse lifetime). Cancelling a source drops its state.

New test: a 100 ms animation finishes and is retired, and an opposing tick 168 ms after the departed direction's last tick still passes only 168/400 of its magnitude.

isleofgreg:
On the shared-state point: on macOS the hook's source identity is the global callback thread (`ScrollSource::OsHook(ThreadId)` — the enum's doc notes "macOS and Windows use one global callback thread"), because a scroll `CGEvent` at the tap carries no device identity. That already makes every per-source structure on this path shared between physical mice: the pulse set and the acceleration window in #1268, and the motion segments in the shipped v0.8.x engine before it. The persistent cooldown inherits that scope rather than introducing it, and the practical exposure is two mice scrolling the same axis in opposite directions within 400 ms of each other. Giving the macOS hook a per-device source would need the device identity plumbed through from the HID layer (the IOHIDEvent sender ID never reaches the CGEvent) — a worthwhile follow-up, but a separate one from the motion model. Happy to open an issue for it if you'd like it tracked.

isleofgreg:
The arm64 test failure on `74f6e043` is the same `tray::tests::startup_stays_suspended_when_the_display_is_already_asleep` flake described on #1268 (nothing in this PR touches `openlogi-agent`); re-pushed the identical tree for a fresh run.

isleofgreg:
Restacked on the rebased #1268 (e3d051bd) after the 09-19 refactor split; all four commits replayed without conflicts, no functional change. Previous head for reference: 69c0fdff.

isleofgreg:
Restacked on the rebased #1268 (1869f447) after #1566 (v0.8.7). All four commits replayed without conflicts, no functional change; fmt, clippy and the workspace tests are green locally. Previous head for reference: e01d0902.

PR #1270, opened 2026-09-04
fix(agent): reapply volatile settings on device reconnection and retry disagreeing DPI writes

isleofgreg:
## Summary

Two related fixes to the volatile-settings re-apply path, found while instrumenting an MX Master 4 on a Bolt receiver (macOS 15.7) over several days of logged use. Standalone — applies to current master, no dependency on the smooth-scroll series.

- **Reapply volatile settings when a device broadcasts its own reconnection** (`0x1d4b` WirelessDeviceStatus). A mouse waking from a nap re-establishes its link and comes back with firmware-default volatile settings (sensor DPI, SmartShift, wheel mode). The re-apply relied on an offline→online transition between snapshots, but a nap short enough that no reconciliation observed the device offline produced no transition, and the one write a reconnect did get could race the link still stabilising — leaving the wrong DPI live until the next nap. The reconnection broadcast already reaches the inventory watcher; it now survives the settle drain and forces a volatile re-apply on the following snapshot (like a system wake, with a single link-race confirming pass instead of the boot ladder). Forced requests coalesce keeping the larger budget. Verified overnight: after a 12 h nap the DPI was rewritten and read back correct 14.5 s after display resume, with the pre-fix `8705` warnings gone.
- **Retry a DPI write once when the read-back disagrees.** Firmware still drowsy from a link re-establishment can ACK `setSensorDpi` yet keep the old value. The read-back already caught the mismatch but only warned; one re-write now separates the two silent-failure shapes — a dropped write lands on the retry, a value the device clamps deterministically comes back unchanged again and keeps the existing warning.

(An earlier revision also reclaimed a stale diverted wheel route on re-apply. It was withdrawn after review: the agent has no signal to tell an abandoned diversion from a live external consumer, and the identity-free reconnect path touches every online device. It will come back separately with an ownership check.)

## Test plan

- `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings` on aarch64-apple-darwin and `x86_64-pc-windows-gnu` for `openlogi-device`/`openlogi-agent-core`; `RUSTDOCFLAGS=-D warnings cargo doc` for those crates; `cargo test -p openlogi-device -p openlogi-agent-core -p openlogi-agent` — green. New unit tests cover the forced re-apply request coalescing and budget, the reconnection-broadcast path, and the single DPI retry (dropped write vs deterministic clamp). (One pre-existing `tray::` test in `openlogi-agent` is timing-flaky under parallel execution on master — #1236 addresses it; passes serially.)
- On hardware: multi-day agent debug logs across naps and overnight sleep.

### DIFF
--- a/crates/openlogi-agent-core/src/orchestrator.rs
+++ b/crates/openlogi-agent-core/src/orchestrator.rs
@@ -153,10 +153,13 @@ pub struct Orchestrator {
     /// the distinction (as [`InventoryHealth`]) so the GUI can tell them
     /// apart.
     inventory: InventoryState,
-    /// Set after a system wake: devices may have power-cycled while their
-    /// set/route/online state looks identical across the sleep gap, so the
-    /// next refresh re-applies volatile settings to every online device.
-    reapply_all_next_refresh: bool,
+    /// Set after a system wake or a device's own reconnect broadcast: devices
+    /// may have power-cycled while their set/route/online state looks
+    /// identical across the gap, so the next refresh re-applies volatile
+    /// settings to every online device. The value is the confirming re-apply
+    /// budget the forced targets are queued with — a system wake keeps the
+    /// boot-race ladder, a device reconnect needs only a link-race confirm.
+    forced_reapply_budget: Option<u8>,
     /// Whether the last enumeration pass failed to open HID++ nodes; published
     /// atomically with the inventory so no observation pairs a fresh device
     /// set with a stale flag.
@@ -239,7 +242,7 @@ impl Orchestrator {
             current: 0,
             current_app: None,
             inventory: InventoryState::Pending,
-            reapply_all_next_refresh: false,
+            forced_reapply_budget: None,
             hid_open_failures: false,
             reapply_followup: HashMap::new(),
             camera_active: None,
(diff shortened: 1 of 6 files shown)


isleofgreg:
Fair point on the reclaim — it treated every diverted route in the forced target set as abandoned, and the identity-free reconnect path makes that set every online device. There's no staleness signal available from the agent's side today (a live external consumer on another device looks identical), so rather than guess I've **withdrawn that commit** from this PR. It's now the two fixes with an unambiguous owner: the reconnection re-apply and the DPI write retry. I'll bring the wheel-route reclaim back separately once there's a defensible staleness check (or scope it to the agent's own startup pass), if that's of interest.

isleofgreg:
Rebased onto master (post-#1286). The only change: the new `replay_tests.rs` fixture matches `InventoryEvent` exhaustively, so the `DeviceWake` variant this PR adds needed adding to its panic arm — folded into the first commit. No production changes.

isleofgreg:
Rebased onto master (a92aa43b, v0.8.6) after the orchestrator split. No functional change: the forced re-apply budget, the refresh path and the two entry points stay in `orchestrator.rs`; the `plan_reapply` rewrite and `RECONNECT_REAPPLY_CONFIRM_RETRIES` moved with `plan_reapply` into `orchestrator/devices.rs`; the tests into `orchestrator/tests/reapply.rs`. The DPI retry commit applied unchanged. Previous head for reference: 0114202d.

PR #1286, opened 2026-09-06
fix(hidpp): stop HID++ replies landing on the wrong request

isleofgreg:
## Summary

A HID++ reply is matched to the oldest pending request whose header it
carries — `(device, feature, function, sw_id)`, or `(device, sub id, register)`
for HID++ 1.0 — and nothing else on the wire ties a reply to its request. Two
requests sharing a header therefore get replies that cannot be told apart, and
this tree let that happen in two ways:

- **Inside the agent.** Startup runs several device sessions at once — the
  settings writer, the gesture capture, and any GUI request arriving over IPC
  — and each resolves its features through root `getFeature` with the
  identical header. A wireless receiver answers such requests out of order
  when a radio packet is retransmitted, so a session pins another session's
  feature index and stays wrong for its whole life: `native wheel mode
  unsupported … feature=0x2121`, `DPI write failed … InvalidFunctionId`,
  `capture session ended … InvalidArgument`, SmartShift `UnsupportedResponse`,
  and `control capture active … buttons=0 gesture_sources=0` — the Actions
  Ring and every extra button dead until the next lucky restart.
- **Across processes.** Every OS hands each input report to *every* open
  handle on a node, and the software-id allocator was a per-process static, so
  the agent and an `openlogi diag …` run (or the GUI) both used software id 1
  and matched each other's replies too. HID++ 1.0 register reads carry no
  software id at all.

Four commits, protocol layer up:
1. `0xB5` sub-register reads accept only the reply that echoes their
   sub-register byte. Without it two concurrent receiver probes handed one
   slot's reply to another slot's read — a phantom paired "slot 3" with
   `wpid=None`, the real slot dropped, `diag` targeting a device that does not
   exist.
2. Software ids are leased host-wide through OS file locks, so a second
   OpenLogi process never shares an id with the first and its replies land as
   unmatched reports.
3. A receiver's register phase (unique id, pairing count, arrival drain, slot
   identities) is serialised across processes with a per-node lock: an
   empty-slot *error* reply carries no sub-register byte, so (1) alone still
   let two concurrent probes answer each other's real slot read with their
   empty-slot errors — both then saw no devices and the agent retired a channel
   that was fine.
4. The channel never keeps two requests with one reply header in flight: a
   request whose header is pending is parked until that request is answered,
   cancelled, or times out. This is the fix for the wedged agent.

## Root cause, with wire evidence

Captured on an MX Master 4 (fw 27.3.19) on a Logi Bolt receiver, macOS 15.7,
with `OPENLOGI_LOG=debug,hidpp=trace` on the agent.

**The receiver reorders same-header replies.** A wedged startup, with the
settings GUI connected over IPC and no other process on the node. Session C
(the gesture capture, running inside the IPC handler's span) asks root for
`0x2150` at `.084`; session B (the settings writer) asks for `0x2121` at
`.099`. The replies come back in the *opposite* order, so FIFO matching hands
`0x2121`'s index 18 to C and `0x2150`'s index 19 to B:

```
23.084458 hidpp request dev=2 feat=0 func=1            ← C: getFeature(0x2150)
23.099479 hidpp request dev=2 feat=0 func=1            ← B: getFeature(0x2121)
23.114377 raw report received matched=true payload=[11, 02, 00, 01, 12, 00, 01, …]   ← index 0x12 = 0x2121 → taken by C
23.129225 raw report received matched=true payload=[11, 02, 00, 01, 13, 00, 00, …]   ← index 0x13 = 0x2150 → taken by B
23.129358 hidpp request dev=2 feat=19 func=1           ← B asks feature 19 (the thumbwheel) for wheel capabilities
23.144569 raw report received matched=true payload=[11, 02, 12, 01, 0f, 1c, 18, 18, …] ← C reads 0x2121's capabilities as thumbwheel info
23.159449 DEBUG native wheel mode unsupported index=2 … feature=0x2121
23.249300 DEBUG thumb wheel reports no single tap — click not capturable
23.294443 WARN  DPI write failed error=HidppFeature { operation: WriteDpi, feature_hex: 8705, kind: InvalidFunctionId }
23.309378 DEBUG capture session ended error=HID++ protocol error: Feature(InvalidArgument)
26.700370 INFO  control capture active index=2 gesture_sources=0 gesture_buttons=0 dpi_buttons=0 buttons=0 …
```

Twenty-one restarts without the GUI connected came up healthy; the first
with it connected wedged. With (4) the second request parks until the first
is answered, so replies can only arrive in request order.

**Foreign replies land on the agent's channel.** With the agent healthy and
idle, a concurrent `openlogi diag features` run shows up in the *agent's*
trace as unmatched reports — the CLI's ping reply, its `getFeature(0x0001)`
reply, then its whole feature-table walk, all with software id 1
(`func|sw = 0x11`, `0x01`):

```
21:53:58.062 raw report received len=20 matched=false pending_count=0 payload=[11, 02, 00, 11, 04, 05, ...]   ← CLI's ping reply
21:53:58.077 raw report received len=20 matched=false pending_count=0 payload=[11, 02, 00, 01, 01, 00, 02, ...] ← CLI's getFeature(0x0001) reply
21:53:58.092 raw report received len=20 matched=false pending_count=0 payload=[11, 02, 01, 01, 2d, ...]       ← CLI's featureSet.getCount
21:53:58.107 raw report received len=20 matched=false pending_count=0 payload=[11, 02, 01, 11, 00, 01, ...]   ← CLI's featureSet.getFeatureId(1)
…
```

`pending_count=0` is the only reason these were harmless in that run. The
agent's own startup issues *concurrent* `getFeature` calls with the identical
header — two device sessions ping and resolve features at once — so replies
are attributed purely by queue order:

```
21:39:58.000425 hidpp request dev=2 feat=0 func=17    ← ping, session A
21:39:58.000675 hidpp request dev=2 feat=0 func=17    ← ping, session B
21:39:58.028325 hidpp request dev=2 feat=0 func=1     ← getFeature, A
21:39:58.035247 hidpp request dev=2 feat=0 func=1     ← getFeature, B
21:39:58.050140 hidpp response dev=2 feat=0
21:39:58.058243 hidpp response dev=2 feat=0
```

A foreign `getFeature` reply arriving in that window is taken by the first
pending request, and every later reply shifts one request down the queue.

**Register reads cross-talk deterministically.** Two `openlogi diag`
processes racing each other (8 rounds of `diag dpi` ‖ `diag smartshift`):
15 of 16 runs reported `device: Slot 3 (slot 3 on receiver …)` and failed
with `device at index 0x03 did not respond to HID++`; one run
saw `paired-device count mismatch — expected=1 found=0`. The `0xB5` reply
echoes the sub-register as its first data byte (`[11 ff 83 b5 52 …]` for slot
2's pairing info, `[11 ff 83 b5 62 …]` for its codename), which is exactly the
byte the predicate was not checking. The same race, run while the agent was
up, made the agent's *own* inventory report `paired slot slot=3 online=true
wpid=None … has_event=false` and fail every write with
`DeviceUnreachable { index: 3 }` until the CLI stopped.

## Changes

- **openlogi-hidpp**
  - `protocol::v10::is_rap_response` takes an optional echo byte: a reply
    must repeat it as its first data byte. Error replies (no data byte) still
    match on the header.
  - New `HidppChannel::read_long_sub_register(device, address, sub_register,
    [p1, p2])`, which sends the sub-register as the first parameter and
    accepts only a reply echoing it. `read_long_register` is unchanged and
    shares the implementation.
  - Bolt and Unifying receivers use it for every `0xB5` read (pairing
    information, codename, Unifying receiver info).
  - Tests: a `0xB5` read for slot 2 ignores slot 3's reply and takes its own;
    a plain register read still matches on the header alone. The mock
    channel's `send_incoming` becomes `pub(crate)` for the first.
- **openlogi-device**
  - New `host_lock` module: exclusive OS file locks under
    `<temp dir>/openlogi-locks/` shared by every OpenLogi process on the host
    (`std::fs::File::try_lock`, stable since Rust 1.89; no new dependency, no
    `unsafe`). `try_lock` is non-blocking; `lock_within` polls up to a budget
    and gives up rather than hang on a stuck holder. The OS releases a lock
    with its holder, so a crashed process cannot strand one. The temp
    directory rather than a profile directory so a dev-profile agent and a
    release CLI meet at one path.
  - `inventory::probe` holds a per-node lock (FNV-1a of the backend node id,
    so different binaries agree on the name) for the register phase of a Bolt
    or Unifying probe — unique id, pairing count, arrival drain, slot
    identities — bounded at 5 s. The per-device feature walks stay unlocked:
    they address each device by index under the process's own software id.
    Needed even with the echo check: an empty-slot *error* reply carries no
    sub-register byte, so two concurrent probes still answered each other's
    real slot read with their empty-slot errors, both reported no devices, and
    the agent retired a channel that was fine.
  - The two direct `0xB5` reads in `inventory::probe` (Unifying codename,
    Bolt codename) go through `read_long_sub_register`.
  - Tests: a held lock excludes a second taker until dropped; `lock_within`
    takes a lock released inside its budget and gives up past it; node lock
    names are stable and distinct.
- **openlogi-hid**
  - Software-id leases are backed by a `host_lock` per id (`sw-id-N`), so a
    second OpenLogi process skips ids the first holds and its replies never
    match the first process's predicates — they land as unmatched reports. If
    the lock directory is unusable the lease falls back to the process-local
    table, with one warning.
  - The in-process bitmask becomes a table of leases guarded by a mutex; the
    id-exhaustion message names other processes as a possible holder.
  - Tests: a leased id's lock file is held (as a second open file description
    sees it) until freed; an id locked by another process is skipped.

- **openlogi-hidpp (channel)**
  - `PendingQueue` replaces the bare `VecDeque`: each pending message carries
    its correlation key (the report's first three bytes), and
    `PendingRequest::register_when_key_free` parks a request whose key is in
    flight, woken whenever a pending message leaves the queue (answered,
    cancelled, or timed out). Check and registration happen under one lock,
    so two parked requests woken together cannot both slip in. The wait
    counts against the caller's timeout. A parked request logs
    `hidpp request parked — same header in flight` at trace level.
  - Tests: a same-header request neither reaches the wire nor registers while
    the first is pending, a different header is unaffected, and cancelling
    the first releases it; two same-header requests issued together are each
    answered by their own reply, in order.

Not changed, on purpose: the agent still resolves features per call through
`root().get_feature`, and its startup sessions still overlap — (4) makes
that safe without touching them. Also unchanged: a process's `trigger_device_arrival` still
makes the receiver broadcast `0x41` arrivals that every other process sees as
a lifecycle event and reconciles on — noisy, but a reconcile, not a wrong
answer.

## Testing

```
cargo fmt --all
cargo clippy -p openlogi-hidpp -p openlogi-hid -p openlogi-device --all-targets -- -D warnings
cargo test -p openlogi-hidpp -p openlogi-hid -p openlogi-device
cargo check -p openlogi-hid -p openlogi-device --target x86_64-pc-windows-gnu
```

All green (209 + 18 + 205 tests). New tests:
`channel::tests::a_request_waits_while_the_same_header_is_in_flight`,
`channel::tests::same_header_requests_are_answered_in_order`,
`protocol::v10::tests::sub_register_read_ignores_another_sub_registers_reply`,
`protocol::v10::tests::plain_register_read_matches_on_the_header_alone`,
`host_lock::tests::{a_held_lock_excludes_a_second_taker_until_dropped,
lock_within_waits_for_a_release_inside_its_budget,
lock_within_gives_up_after_its_budget, node_lock_names_are_stable_and_distinct}`,
`transport::sw_id_lease_tests::lease_holds_the_ids_os_lock_until_freed`,
`transport::sw_id_lease_tests::lease_skips_an_id_held_by_another_process`.

Hardware (MX Master 4 fw 27.3.19 on a Logi Bolt receiver, macOS 15.7, this
branch cherry-picked onto my dev build and installed):

- 9 agent restarts in a row came up healthy (`control capture active …
  buttons=3 gesture_sources=1 dpi_buttons=1`): 6 with the settings GUI
  connecting over IPC at startup — the condition that wedged the unfixed
  build on its first launch — and 3 with two `openlogi diag` loops hammering
  the node through the startup window. The trace shows the guard engaging
  (`hidpp request parked — same header in flight`) 2–37 times per start.
- Two `openlogi diag` processes racing each other for 8 rounds (16 runs):
  before, 15/16 reported a phantom "Slot 3" and failed on it; with (1)+(2)
  only, 16/16 failed with "no device" (the error-frame case (3) fixes);
  with all four, 16/16 found the real device, no phantom slot, no
  `count mismatch` warnings.
- While the agent runs, `OPENLOGI_LOG=hidpp=trace openlogi diag dpi` shows
  the CLI's requests carrying software id 2 (`func=0x12`, `0x02`); the agent
  holds id 1. Lock files appear under `$TMPDIR/openlogi-locks/`.
- Unifying receivers: not runtime-tested on hardware (the same code paths,
  compile-checked and unit-tested).

Windows and Linux: compile-checked only (`x86_64-pc-windows-gnu`); the lock
primitive is std's, which uses `LockFileEx` / `flock` respectively.


isleofgreg:
Greptile's three findings were all real. Addressed in four commits on top of b92ff07 — one per finding, plus the rustdoc CI failure:

- **Stale replies matching new requests** (`channel.rs`) → 1cab091. A request abandoned unanswered — timed out, or cancelled by an outer deadline — now leaves its header reserved until its reply arrives and is discarded, or for `STALE_REPLY_GRACE` (1 s) at most, so a device that never answers costs the next request one grace window rather than a whole timeout. A byte-identical re-ask (what `read_feature_entry` does when the link drops a report) is the exception: it goes out at once, adopts the owed reply, and is answered by whichever comes first; the channel keeps counting the other so a later, *different* request cannot be answered by it either. New tests cover the late reply, the grace expiry, and the re-ask. The residual is a reply more than a second late, which is as unattributable as it always was.
- **Unrelated nodes exhausting the id pool** (`transport.rs`) → f984962. Leases are `(node, id)`: the in-process table is keyed that way and the lock file is named after the node's own lock name, in the one directory `host_lock` already owns — which also retires the transport's private copy of the file-locking code. `SwIdPolicy::Leased` now carries the lease as an owned handle dropped with the policy instead of a bare `fn(u8)`, since a free function cannot tell which node's id it is returning. New test: fifteen leases on one node do not refuse the first on another.
- **Lock timeout restoring the race** (`host_lock.rs`) → e7b8602. Contention past the wait settles a failed probe — "couldn't check", which the ledger already handles by replaying the node's last snapshot and retrying next tick — instead of probing unlocked. `lock_within` distinguishes that from the lock directory being unusable, the one case with nothing to wait for, where the probe runs unarbitrated as before. Two consecutive deferrals count as two failures for the ledger, so a holder that hogs a receiver's register phase for more than 5 s twice in a row retires the agent's channel; that holder is itself failing, so that seems the right side to err on.
- 912fda3 fixes the rustdoc job: a public doc linked a private matcher, and the job runs with `-D warnings`.

`cargo clippy --workspace --all-targets`, the non-GUI `cargo doc` with `-D warnings`, and the `openlogi-hidpp` / `openlogi-hid` / `openlogi-device` tests pass locally.

crates/openlogi-device/src/inventory/probe.rs
```diff
@@ -22,13 +26,63 @@ use super::events::EventSubscriptionHandle;
...
+    )
+    .await
+    {
+        Ok(Some(lock)) => Some(RegisterPhase { _lock: Some(lock) }),
+        Ok(None) => {
+            debug!(
+                node = %info.id,
+                "another OpenLogi process still holds this receiver's register phase — deferring the probe"
+            );
+            None
```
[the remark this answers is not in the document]
isleofgreg: Agreed — the trade-off I noted in the summary was the wrong side to err on. 04dddc24 gives deferral its own verdict, `ProbeVerdict::Deferred`: the ledger replays the node's last-good inventory through a new `NodeLedger::defer` that leaves the failure count and eviction alone (a channel that was never asked cannot have failed), while the verdict still fails the pass's `all_healthy` so the one-shot retry re-probes as before. Ledger test `deferred_ticks_replay_without_counting_toward_eviction` covers it: any number of deferrals between two real failures, and the second real failure is still the one that evicts.

crates/openlogi-device/src/inventory/probe.rs
```diff
@@ -124,6 +201,9 @@ async fn probe_bolt_receiver(
     now: Instant,
     subscriptions: Option<&EventSubscriptionHandle>,
 ) -> NodeProbe {
+    let Some(register_phase) = lock_receiver_registers(&info).await else {
```
AprilNEA: **P1 — Keep lock acquisition outside the receiver's I/O budget.**

`inventory.rs` starts the 13-second `RECEIVER_PROBE_BUDGET` around `probe_one`, so this new wait can consume almost five seconds before any receiver I/O starts. The subsequent Bolt probe still needs the arrival drain (up to 1.5 seconds), register reads, and a slot walk with its own 10-second budget. After a four-second lock wait, the outer timeout can therefore fire before the slot's intended timeout/cache fallback. That becomes `Failed`, not `Deferred`; two consecutive occurrences retire the channel and remove its routes despite successful lock acquisition and a working receiver.

Please separate the bounded lock-acquisition wait from the existing I/O deadline, or otherwise account for both explicitly. Add a composed test where the lock is released within its wait budget and a slow slot reaches its normal fallback without failing the entire receiver probe. The current lock and ledger tests exercise those components separately and miss this interaction.

crates/openlogi-hidpp/src/channel.rs
```diff
@@ -206,18 +229,189 @@ impl Drop for HidppChannel {
...
+        if self
+            .messages
+            .iter()
+            .any(|pending| pending.key == message.key)
+        {
+            return Err((message, Wait::InFlight));
+        }
+        let mut blocked_until: Option<Instant> = None;
+        for stale in self.stale.iter().filter(|stale| stale.key == message.key) {
+            if stale.request != message.request {
```
AprilNEA: **P2 — Byte-identical requests do not necessarily have interchangeable replies.**

The unconditional re-ask exception is unsafe for mutable state. For example: a DPI read produces an 800-DPI reply that is delayed; its caller is cancelled; a different function sets DPI to 1600 and receives its ACK; then a new DPI read sends the same bytes as the cancelled read. This branch lets the new read adopt the pre-write 800-DPI reply and treats its own 1600-DPI reply as an extra stale reply. All of this can happen inside `STALE_REPLY_GRACE`, so it is distinct from the documented residual risk beyond one second.

A local regression test through `send_v20` and the existing mock reproduced exactly this result: expected payload `[0, 6, 64]`, received `[0, 3, 32]`. Please retain quarantine by default. If immutable feature-table queries need this optimization, make it an explicit semantic choice at that boundary rather than inferring it from byte equality in the generic channel.

crates/openlogi-device/src/inventory/probe.rs
```diff
@@ -247,6 +331,7 @@ async fn probe_unifying_receiver(
         return NodeProbe::failed();
     };
     debug!(events = connections.len(), "drained device-arrival events");
+    drop(register_phase);
```
AprilNEA: **P2 — The Unifying register phase still performs RAP I/O after releasing this lock.**

`probe_unifying_slot` can subsequently call `read_codename_unifying` when capabilities are available but the marketing name is missing. That reads receiver `0xff`, register `0xB5`, after this guard has been dropped. Another process can then acquire the guard and start its register phase while the codename request is pending. Successful responses have the new echo check, but an error from one `0xB5` sub-register still matches another request's header because the error contains no sub-register byte. An unsupported codename read can thus interfere with the other process's receiver-info read, or vice versa; a failed codename is also published without making the enclosing probe unhealthy.

Please include these remaining RAP reads in the same cross-process arbitration rule, either by moving them into the protected identity phase or acquiring the guard at their actual I/O boundary. Audit non-inventory receiver-info reads too: `open_route_channel` and receiver selection also call `get_unique_id` without this guard. The lock only works if conflicting callers participate.

crates/openlogi-device/src/inventory/probe.rs
```diff
@@ -89,6 +156,16 @@ impl NodeProbe {
...
         }
     }
+
+    /// A probe that never ran: another process held the receiver's register
+    /// phase for longer than it was willing to wait.
+    pub(super) fn deferred() -> Self {
+        Self {
+            inventory: None,
+            verdict: ProbeVerdict::Deferred,
+            outcomes: Vec::new(),
```
AprilNEA, line 126: **P2 — Deferred probes still count as missing devices in the feature cache.**

The new verdict preserves the node ledger, but its empty `outcomes` still flows into `apply_outcomes` followed by `evict_unseen` at the end of enumeration. After four deferred passes (`CACHE_MISS_GRACE = 3`), the receiver's feature-cache entries are deleted even though its last-good inventory and channel remain present and no device I/O failed. When probing resumes with a device offline or its feature walk failing, the previously retained capabilities/model data are no longer available for fallback. Bolt cache eviction can also be persisted.

A local regression test using `NodeProbe::deferred`, `settle_probe`, `apply_outcomes`, and `evict_unseen` reproduced the cache deletion while the ledger continued replaying inventory without requesting channel eviction. Please suspend cache-miss aging for the deferred node's entries as well; preserve normal aging for nodes actually checked or removed.

AprilNEA (changes requested):
The correlation and cross-process arbitration changes address the reported startup failure, but four cases still need changes: lock acquisition consumes the existing probe deadline; identical requests can adopt replies from before an intervening state change; Unifying register reads outlive the register-phase lock; and deferred probes still age out the feature cache. Details are attached inline.

Validation on this head: `cargo test -p openlogi-hidpp -p openlogi-device -p openlogi-hid --lib` passed all 437 tests. Two additional local regression tests reproduced the read-after-write stale-response case and feature-cache eviction after four deferred probes. The deadline and Unifying lock-coverage findings are based on tracing the current code paths; not runtime-tested on hardware.

crates/openlogi-device/src/inventory/probe.rs
isleofgreg: Agreed — the wait sat inside a budget that was sized for the receiver's I/O alone. `1b6b565e`: `probe_one` now takes the register phase first and only then bounds the probe's I/O (`within_budget`), and every deadline of a pass lives in one `ProbeDeadlines` value (production defaults unchanged) so the composition is one place to read and one value for a test to shrink; the probes share a `PassContext` instead of threading cache, clock, event sink and deadlines separately.

The composed test is `a_receiver_probe_that_waited_for_its_register_phase_keeps_its_whole_io_budget`: a scripted Bolt receiver whose paired mouse never answers, the phase held 600 ms by the test and released inside a 2 s wait, a 100 ms drain and a 400 ms slot probe under a 900 ms budget. Wait + I/O outrun the budget by construction (asserted), the slot reaches its timeout-and-cache fallback, and the receiver settles `Healthy { complete: true }` on the cached slot. Alongside it: a phase held past the wait defers without a byte of I/O (`a_receiver_probe_defers_when_the_register_phase_is_held_past_the_wait`), and I/O alone outrunning the budget is still `Failed` (`a_receiver_probe_whose_io_outruns_the_budget_is_failed`).

crates/openlogi-hidpp/src/channel.rs
isleofgreg: Right — byte equality says what was asked, not what the answer is. `7e39daab` quarantines by default whatever the new request's bytes are. Adoption is now the caller's explicit choice, `AbandonedReply::AdoptIdentical` on `send_with` / `send_v20_with` / `FeatureEndpoint::call_with`, and the only caller that makes it is `FeatureSetFeature` (`count`, `get_feature`): the table is fixed for the life of the connection, so a walk re-asking for a dropped entry is not held for the grace on every retry. That opt-in is load-bearing rather than an optimisation — with a blanket quarantine, `read_feature_entry`'s 700 ms attempts alternate with the 1 s grace and only two of its four re-asks ever reach the wire (`lost_feature_entry_is_retried_before_giving_up` caught it).

Your read-after-write case is the test `a_re_asked_read_cannot_adopt_a_reply_from_before_an_intervening_write` (through `send_v20` and the mock: the re-asked read gets `[0, 6, 64]`, the late `[0, 3, 32]` is discarded). `an_identical_re_ask_waits_for_the_quarantined_reply` pins the default, `an_identical_re_ask_may_adopt_the_outstanding_reply_when_it_says_so` the opt-in, and `adoption_is_decided_by_the_re_ask_not_the_abandoned_request` that the abandoned request's own policy plays no part.

crates/openlogi-device/src/inventory/probe.rs
isleofgreg: Yes — the codename read was the one register read behind the walks. `1b6b565e` keeps the Unifying phase to the end of the probe (dropped after the slot walks join, i.e. after the last `0xB5` read the probe can make); the phase's shared wait allows for the walks running under it.

The audit is `1222cbbb`: the phase moved out of the probe into `host_lock::lock_receiver_registers` / `ReceiverRegisterPhase`, with one wait (`RECEIVER_REGISTER_WAIT`, 8 s) for every taker. `open_route_channel`'s Bolt and Unifying arms take it around `get_unique_id` — a receiver held elsewhere past the wait is passed over this time, the same `None` as a receiver that is not there, rather than read into the holder's replies. Pairing's `read_bolt_uid` takes it likewise, and a pairing session (`run_pairing`, `unpair`) takes it once in `open_receiver` and holds it through its last restore write, since every write in the flow is receiver register I/O and the flow waits on the user between them; inventory probes in every process defer to it meanwhile and pick the new pairing up on release.

A failed codename still publishes as `None` without failing the probe — that was the pre-existing behaviour, and it is the register phase rather than the name the finding is about — but I'm happy to make it count if you'd rather.

crates/openlogi-device/src/inventory/probe.rs
isleofgreg: Agreed. `f9c22af1`: the enumerator records the cache entries each node contributes (`node_cache_keys`), and a deferred tick passes those to `evict_unseen` as frozen — neither seen nor missed, counters standing until the node is actually probed again. Entries of nodes that were checked or removed age exactly as before; an evicted entry stops being any node's, and a vanished node's set goes with its ledger entry, so nothing stays frozen for good.

`deferred_ticks_hold_the_nodes_cache_entries_out_of_miss_aging` runs your composition — `NodeProbe::deferred` → `settle_probe` → `apply_outcomes` → `evict_unseen` — for `CACHE_MISS_GRACE + 1` passes with a checked node's unreported entry aging alongside, then resumes normal aging once the deferred node is probed again.

isleofgreg:
Thanks for the review — all four were real. Fixed as four commits on top of `04dddc24`, one per finding, replies in the threads:

- `7e39daab` — quarantine by default whatever the re-ask's bytes are; `AbandonedReply::AdoptIdentical` is the explicit opt-in, taken only by `FeatureSetFeature` (whose table cannot change), which is what keeps `read_feature_entry`'s re-asks reaching the wire.
- `1222cbbb` — every receiver register caller takes the node's phase (`host_lock::lock_receiver_registers`): route opening, pairing's receiver selection, and pairing sessions for their whole run.
- `1b6b565e` — the phase is taken before the receiver's I/O budget starts, with every deadline of a pass in one `ProbeDeadlines`; the Unifying probe holds its phase through its codename reads; composed test with a scripted Bolt receiver whose slot runs to its fallback after a lock wait that, together with the I/O, outruns the budget.
- `f9c22af1` — deferred ticks hold the node's cache entries out of miss aging.

Validation on `f9c22af1`: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, the CI rustdoc command (`--document-private-items`, `-D warnings`), and `cargo test --workspace --no-fail-fast --exclude openlogi-desktop` — 1187 passed, 0 failed (`openlogi-hidpp` 215, `openlogi-device` 210). Not hardware-tested at the time of writing; the dev bundle carrying these commits goes onto the MX Master 4 / Bolt setup that produced the original traces next.

crates/openlogi-device/src/inventory.rs
```diff
@@ -858,16 +911,53 @@ impl Enumerator {
...
+        frozen: &mut HashSet<CacheKey>,
+    ) {
+        if probe.verdict.is_deferred() {
+            frozen.extend(
+                self.node_cache_keys
+                    .get(node)
+                    .into_iter()
+                    .flatten()
+                    .cloned(),
+            );
```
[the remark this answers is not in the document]
isleofgreg: Real — the warm start was the gap: the persisted Bolt entries are in `cache` before any receiver has answered, so a receiver deferred from its first pass had nothing recorded to hold. `1c80799f`: a node deferred with no record yet holds every entry no node has claimed (`unattributed_cache_keys`) — nothing that was checked contributed them, so nothing that was checked can have found them missing — and its first real probe attributes what is its, after which its deferrals hold only that. Test `a_node_deferred_before_its_first_probe_holds_every_unattributed_entry`: a persisted entry survives `CACHE_MISS_GRACE + 1` deferrals of the never-probed receiver while a checked node's unreported entry ages alongside, then ages normally once the receiver has been probed.

crates/openlogi-device/src/inventory/probe.rs
AprilNEA: Verified on the updated branch: `probe_one` acquires the receiver register phase before entering `within_budget`. The composed lock-wait/slow-slot regression passes, as do the cases for lock contention without I/O and actual I/O-budget exhaustion. This addresses the original finding; resolving.

crates/openlogi-device/src/inventory/probe.rs
AprilNEA: Verified on the updated branch: the Unifying guard is dropped only after all slot walks, including codename reads, finish. Route opening and pairing receiver selection acquire the same per-node phase, and pairing sessions retain it through their final writes. The optional-codename failure behavior can remain unchanged; this finding concerned lock coverage. Resolving. Validation is by code inspection and scripted tests, not physical hardware.

crates/openlogi-device/src/inventory/probe.rs
AprilNEA: Verified with the cache bookkeeping regressions: deferred passes preserve known entries and warm, unattributed entries without advancing misses or persisting deletion. Coverage now includes an initial failed probe and a partial first probe, followed by repeated deferrals. Checked or confirmed-removed entries still age normally. The new partial-probe regression is in https://github.com/AprilNEA/OpenLogi/commit/fbe078e1f1b4dbe687b6115b98cf3daf66f68cf4. Resolving.

crates/openlogi-device/src/inventory.rs
```diff
@@ -864,16 +909,77 @@ impl Enumerator {
...
+                Some(keys) => frozen.extend(keys.iter().cloned()),
+                None => frozen.extend(self.unattributed_cache_keys()),
+            }
+            return;
+        }
+        // A failed first probe cannot establish cache ownership, even if it
+        // found some slots before failing. Keep unknown attribution distinct
+        // from a healthy probe that positively found no entries.
+        if !probe.verdict.is_healthy() && !self.node_cache_keys.contains_key(node) {
+            return;
```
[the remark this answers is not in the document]
AprilNEA: The direct-device sequence does not occur in the current code: `probe_one` returns `Deferred` only from the Bolt/Unifying register-lock branches. Direct probes do not acquire that lock.

For receivers, a failed identifying outcome is not proof that every paired slot was identified. Added `partial_first_probe_does_not_abandon_unread_slots_during_deferral` in https://github.com/AprilNEA/OpenLogi/commit/fbe078e1f1b4dbe687b6115b98cf3daf66f68cf4: two pairings are reported, only one identity is readable, then the receiver is repeatedly deferred. Allowing the failed identifying outcome to establish the ownership set makes this test fail by deleting the unread slot. The existing policy passes, and the test also verifies normal eviction resumes after a complete probe confirms the slot is gone.

Keeping unknown attribution until a healthy probe is intentional here. Resolving this finding as not applicable rather than applying a change that reintroduces cache loss.

crates/openlogi-hidpp/src/channel.rs
AprilNEA: Verified on the updated branch: quarantine is the default, and only immutable `FeatureSetFeature` queries opt into `AdoptIdentical`. The read-after-write regression returns the new 1600-DPI value and discards the old 800-DPI reply; the explicit-adoption tests also pass. This addresses the original finding; resolving.

isleofgreg:
Thanks for the rebase and the three commits on top — I've pulled the branch as-is and carried `10b56839` + `fbe078e1` into the build running on my MX Master 4 / Bolt setup. Agree that a failed first probe must leave attribution unknown rather than record an empty ownership set; the partial-probe regression pins exactly the case that motivated the last commit of mine. Nothing further from my side.

AprilNEA (approved):

isleofgreg:
Thank you for the careful review, the rebase, and the extra regression coverage — and for getting it into v0.8.5 so quickly. Much appreciated.

PR #1297, opened 2026-09-07
fix(overlay): fix Actions Ring position on a scaled Windows monitor

4ni1ak:
## Summary

A commenter on #1210 confirmed the same symptom reported there (fixed for Linux/Wayland in #1211) also happens on Windows 11: the ring opens offset from the pointer and isn't reliably above other windows.

## Root cause

`ring_window_options` fed `openlogi_hook::cursor_position()`'s raw `GetCursorPos` coordinates — physical pixels — straight into GPUI's window placement, which treats `WindowOptions` bounds as logical points.

Confirmed by reading `gpui_windows`'s own source rather than guessing: `WindowsDisplay::bounds()` (`display.rs`) divides each monitor's rect by its DPI scale before handing it to GPUI as `PlatformDisplay::bounds()`, and the window-creation path (`window.rs`) runs incoming coordinates through `logical_point(x, y, scale_factor)`. Neither macOS nor X11/Wayland has this split — both already report cursor position in the same unit GPUI expects. On any Windows monitor not running at 100% scaling, the ring lands at `cursor / scale`, offset toward the top-left corner — the higher the scale factor, the worse the offset.

## Changes

- `crates/openlogi-overlay/src/platform.rs`: added a Windows `display_containing` (`MonitorFromPoint` + `GetDpiForMonitor` + `GetMonitorInfoW`), mirroring the existing macOS CoreGraphics implementation in shape. `CursorDisplay` gained a `scale` field — `1.0` on macOS (already points), the monitor's DPI scale (`dpi / 96`) on Windows.
- `crates/openlogi-overlay/src/ring.rs`: extracted the cursor-to-display-space conversion into `cursor_in_display_space`, which now divides the raw cursor point by `display.scale` before subtracting the display origin.
- `crates/openlogi-overlay/Cargo.toml`: added `windows-sys` (already a workspace dependency, used elsewhere in `openlogi-hook`) as a `target.'cfg(target_os = "windows")'` dependency with the `Win32_Graphics_Gdi`/`Win32_UI_HiDpi`/`Win32_UI_WindowsAndMessaging` feature modules.

## On the "not always on top" half

`gpui_windows` already requests `WS_EX_TOPMOST` for `WindowKind::PopUp` (confirmed by reading `window.rs`), so this isn't a missing window style. I can't verify on real Windows hardware whether that's sufficient in practice, or whether the position bug alone explains what was reported — a ring rendered somewhere unexpected because of the DPI offset can read as a z-order failure. Flagging this honestly rather than guessing further; worth the reporter confirming after this lands whether the topmost behavior is still an issue on its own.

## Testing

Linux, x86_64, Rust 1.98.0:

```
cargo fmt --all -- --check
cargo clippy -p openlogi-overlay --all-targets -- -D warnings   # RUSTFLAGS=-D warnings
cargo test -p openlogi-overlay                                   # 19 passed
cargo check --workspace --all-targets
```

All clean. Two new unit tests cover the pure `cursor_in_display_space` conversion at 100% and a 150% DPI scale. Confirmed `Cargo.lock`'s `gpui`/`gpui-component` git pins didn't move (only a feature-list addition to the existing `windows-sys` entry).

**The new `display_containing` implementation itself is `#[cfg(target_os = "windows")]` and does not compile on this host at all — I cannot `cargo check`/test it here.** I wrote it against `windows-sys` 0.61's actual API signatures (verified in the crate's own source under `~/.cargo/registry`, not guessed) and matched the project's existing raw-pointer FFI conventions (`&raw mut`, `SAFETY` comments) as closely as possible, but this needs Windows CI or real hardware to confirm it actually compiles and behaves. Not runtime-tested against real hardware — that verification is the maintainer's/reporter's to do.

Related: #1210, fixed for Linux/Wayland in #1211.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5719,6 +5719,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/crates/openlogi-overlay/Cargo.toml
+++ b/crates/openlogi-overlay/Cargo.toml
@@ -44,5 +44,17 @@ objc2-app-kit = { workspace = true, features = ["NSApplication", "NSResponder", 
 # NSEvent global-monitor handlers are ObjC blocks (click-away dismissal).
 block2 = { workspace = true }
 
+[target.'cfg(target_os = "windows")'.dependencies]
+# MonitorFromPoint/GetMonitorInfoW/GetDpiForMonitor: placing the ring on the
+# cursor's display, and converting its physical-pixel position to GPUI's
+# logical points, needs a DPI query GPUI exposes no API for before a window
+# exists to ask.
+windows-sys = { workspace = true, features = [
+    "Win32_Foundation",
+    "Win32_Graphics_Gdi",
+    "Win32_UI_HiDpi",
+    "Win32_UI_WindowsAndMessaging",
+] }
+
 [lints]
 workspace = true
--- a/crates/openlogi-overlay/src/platform.rs
+++ b/crates/openlogi-overlay/src/platform.rs
@@ -163,6 +163,10 @@ pub fn display_containing(x: f64, y: f64) -> Option<CursorDisplay> {
     unsafe_code,
     reason = "MonitorFromPoint/GetMonitorInfoW/GetDpiForMonitor are plain Win32 FFI; GPUI exposes no DPI query before a window exists"
 )]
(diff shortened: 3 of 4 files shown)


4ni1ak:
Superseded — #1319 landed the same Windows Actions Ring mixed-DPI fix directly on master (same root cause, physical virtual-desktop coordinates instead of the old logical-display lookup). Closing.

PR #1024, opened 2026-08-26
fix(device): stabilize MX Ergo receiver and Bluetooth switching

igit-cn:
## Summary

- Add reliable MX Ergo S support across Bolt receiver and Bluetooth-direct connections.
- Preserve one logical device and one shortcut configuration while switching transports.
- Keep shortcut capture alive across reconnects, wakeups, and transport changes.

## Changes

- Bind cached HID++ features to the active channel generation.
- Serialize transactions that share a HID++ software ID.
- Read Unifying pairing data from the correct slot registers.
- Discover the Generic Desktop HID++ fallback exposed by MX Ergo Bluetooth devices.
- Prefer vendor HID++ collections per physical device when both collection types exist.
- Reuse the inventory channel for button and gesture capture.
- Migrate legacy route-specific aliases to stable physical-device identities.
- Hydrate offline linked devices without creating duplicate cards.
- Add Ergo M575 Bluetooth fallback regression coverage.

## Testing

- `cargo +1.98.0 fmt --all -- --check`
- `cargo +1.98.0 clippy --workspace --all-targets -- -D warnings`
- `cargo +1.98.0 test --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo +1.98.0 doc -p openlogi-hid -p openlogi-hidpp -p openlogi-hidpp-derive --no-deps --document-private-items`
- `cargo +1.98.0 run -p xtask -- ci clippy-windows`

## Hardware verification

- Tested on macOS 26.6.2, Apple Silicon.
- MX Ergo S tested with both a Bolt receiver and Bluetooth-direct.
- Shortcut remapping and stable device identity tested after switching transports.
- Ergo M575 checked for regressions; its Bluetooth fallback is also covered by unit tests.
- A stale macOS Input Monitoring grant had to be removed and re-added before the running agent could access HID input again.

Fixes #366

PR #1300, opened 2026-09-07
fix(macos): stabilize MX Ergo S and M575 connections and shortcuts

### LINKED ISSUE #366: [Device]: MX Ergo S, unsupported features
### Pre-flight checklist

- [x] I quit **Logi Options+** before testing.
- [x] I searched existing issues for my device.

### Device model

MX Ergo S

### Connection

### What's the problem?

Requesting support for a device not yet supported

### `openlogi list` output

```text
installed via brew but openlogi command doesn't seem to work, cli app isn't there
```

### `openlogi diag features` output

```text
installed via brew but openlogi command doesn't seem to work, cli app isn't there
```

### Details

Device is recognized automatically. Gestures works but when OpenLogi is running DPI button and horizontal scrolling doesn't work.

igit-cn:
## Summary

Replace #1024 with the consolidated receiver/Bluetooth and shortcut-recovery fixes, rebased onto `6c0ec1d9` on `master`. The previous PR was closed after follow-up testing exposed agent-registration and intermittent recovery problems.

This addresses related user-visible failures: a mouse appearing offline or twice after changing transports, configured buttons stopping working after a handoff or idle/wake, and helper lifecycle/configuration problems after replacing a local build. The reporter has tested MX Ergo S and Ergo M575, and confirmed the current installation works after the local service/permission recovery described below. Source tests, hardware feedback and remaining upgrade limitations are reported separately; this is not a claim that every macOS/hardware combination is covered.

### `openlogi-hidpp`: serialize the real transaction, correct Unifying identity reads

- Serialize the entire request/reply transaction on each channel, including raw writes and send-and-forget traffic. Requests sharing correlation fields must not consume each other's replies. A queued request starts its wire timeout only after acquiring the gate; cancellation removes only its own pending request.
- Read Unifying pairing, extended-pairing and device-name registers with their correct zero-based slot offsets (`0x20`, `0x30`, `0x40`), instead of applying Bolt's layout. Validate slot bounds, decode WPID and the low device-kind nibble, and expose the physical unit ID from extended pairing information.
- The reverse-engineered register layout is explicitly attributed to Solaar; no claim is made that it came from an official Logitech specification.

### `openlogi-hid`: discover the Bluetooth collection actually exposed by Ergo devices

- Accept the Logitech `0xb0xx` Generic Desktop mouse collection as a long-report HID++ fallback, including MX Ergo S and Ergo M575. Ordinary receiver/wired mouse collections remain excluded.
- Prefer the vendor HID++ collection when the same OS node and product expose both. Preference is not global by PID, so another same-model mouse is not hidden merely because the first one has a vendor collection.
- Preserve standalone-driver precedence and Linux receiver-child filtering.

### `openlogi-device`: stable physical identity and channel-scoped probing

- Enumerate Unifying paired-slot identity from receiver registers even while a mouse sleeps or uses Bluetooth; overlay current arrival notifications for liveness. A late arrival with a different WPID cannot mark a replacement slot occupant online.
- Preserve arrival-only enumeration for older receivers without those registers. Unreadable slots with no arrival are not invented as devices, and an absent/all-zero unit ID is never used as a physical cache key.
- Key the Unifying probe cache by physical unit ID rather than receiver slot. Receiver/Bluetooth reconciliation can then use the existing transport-independent identity model, without model-name merging.
- Reuse immutable feature data for a channel generation instead of walking the full feature table every 30 seconds. A replacement channel gets a validation attempt; a failed attempt preserves last-good data rather than dropping the device. Battery readings remain live, and warm-cache adoption restores lifecycle/battery notification indexes.
- Persist physical receiver identities and notification indexes in probe-cache schema 4; discard incompatible cache schemas for a safe cold start. This cache is disposable, not the user's settings.

### `openlogi-agent-core` / `openlogi-device`: recover shortcuts without stale cleanup

- Keep failed firmware-restore tokens scoped to their original physical device **and transport**. An unavailable old receiver route must not block capture through the device's new Bluetooth route (and vice versa).
- Keep the existing restore-before-rearm ordering on the same route. Do not replay old feature indexes into a different transport, underneath a running/draining successor, or into a receiver route reassigned to another mouse.
- Retain each route's cleanup debt for a later safe retry; do not discard it to make a handoff appear successful.
- Remove expired restart deadlines so pending restoration cannot keep the manager's timer permanently ready and busy-spin. Restoration continues to respect device-I/O suspension and receiver leases.

### `openlogi-desktop`: one useful record with correct model metadata

- Resolve offline receiver sightings using the receiver's nonzero physical identity and previously recorded route links. Hydrate their model name, kind and artwork from persisted identity instead of showing a generic chip/incorrect mouse category.
- Suppress never-observed, offline receiver pairings with no persisted device record, avoiding anonymous `Slot N` gallery cards. Known offline devices stay available; online unknown devices remain visible and actionable. This does not unpair or delete anything from the receiver.
- Map MX Ergo S's `SLOT_NAME_CHANGE_POINTER_SPEED` metadata control to `DpiToggle`. Cover the full six-control metadata mapping, including wheel tilt, forward/back, middle click and precision-mode button.

### `openlogi-agent` / `xtask`: supervised recovery and visible-wake gating on macOS

- Mark launchd starts with `--launchd`. If a manually launched or legacy agent already owns the singleton, the supervised successor waits instead of exiting successfully and leaving the service unsupervised after the predecessor fails. Only the lock holder starts the input hook.
- Preserve explicit menu-bar Quit with a one-shot marker and clear stale markers on a new explicit run. Keep existing protocol-version takeover behavior and GUI-owned `SMAppService` registration.
- Use generic workspace wake notifications only to schedule display confirmation. Require two separated visible-display samples before reopening HID; an asleep display keeps I/O suspended during DarkWake. Recover when a screen-wake notification is missed while retaining session-inactive suspension.
- Extend packaging tests to assert the launchd argument and the existing crash-restart/clean-Quit policy.

### `openlogi-desktop` registration: distinguish builds, not only release versions

- Compare both package version and the outer app's `CFBundleVersion` when deciding whether to refresh `SMAppService` registration. A different build of the same release now triggers a refresh attempt instead of being treated as already current; this does not guarantee recovery from every stale macOS service record (see installation notes below).
- Use typed `NSBundle` access; the only manifest change enables its existing Foundation feature. Do not override `RequiresApproval`, change bundle identifiers, or change signing requirements.

### `openlogi-core`: preserve settings from earlier custom builds

- Migrate private-build `device:<identity>` keys and explicit `device_aliases` into the current physical-key/route-link model, including selection and host-switch references.
- Keep conflict-preserving folds and a source backup before saving a migrated configuration, even when the schema version is already current. Aliases without a persisted target are not guessed from model names.
- Keep this one-way import in file loading rather than expanding the strict serialized `Config`/IPC shape. It is compiled only with the `fs` feature; ordinary current configs keep the existing schema and wire contract.

### Existing upstream behavior retained, not reintroduced

The current upstream already provides shared inventory-owned capture channels, transport-independent identity/link configuration, device removal with confirmation, wheel-tilt controls, agent-owned permission reporting and permission-grant relaunch. Those parts were reconciled with the newer architecture rather than copied back from the old GUI branch. No layout redesign, locale changes, dependency-version changes, release tags or installer binaries are included.

## Testing

Validated the final commit `715d3906` on Apple Silicon macOS with Rust 1.98.0:

- Full workspace tests: **1,399 passed, 0 failed**, including all 14 IPC wire-format goldens and locale parity; two pre-existing documentation examples are marked ignored.
- Formatting, strict full-workspace Clippy, and the repository's non-GPUI rustdoc gate: **passed**.
- Windows GNU cross-Clippy proxy, the portable WASM checks, and the Linux ARM64 musl cross-Clippy subset below: **passed**. These are compile/lint checks, not full native Windows/Linux CI or hardware tests.
- Release DMG build, `hdiutil verify`, and packaged launchd plist inspection: **passed**. The final local test bundle is ad-hoc signed and passes deep/strict verification. An earlier unsigned artifact failed installation; the distinction and recovery are documented below. The plist has `--launchd`, the nested agent path, and `KeepAlive.SuccessfulExit = false`.
- `git diff --check` passed. No `Cargo.lock`, dependency-version or GPUI-pin changes.

Exact source-validation commands:

```sh
export RUSTFLAGS='-D warnings'
cargo +1.98.0 fmt --all -- --check
cargo +1.98.0 clippy --workspace --all-targets -- -D warnings
cargo +1.98.0 test --workspace
RUSTDOCFLAGS='-D warnings' cargo +1.98.0 doc --workspace --no-deps \
  --document-private-items --exclude openlogi-ui --exclude openlogi-desktop \
  --exclude openlogi-overlay --exclude openlogi-agent
RUSTUP_TOOLCHAIN=1.98.0 cargo xtask ci clippy-windows wasm
cargo +1.98.0 clippy --target aarch64-unknown-linux-musl \
  -p openlogi-hook -p openlogi-inject -p openlogi-hid -p openlogi-hidpp \
  -p openlogi-core -p openlogi-agent -p openlogi-agent-core \
  -p openlogi-ipc -p openlogi-permissions --all-targets -- -D warnings
git diff --check upstream/master...HEAD
```

Packaging ran separately, without `RUSTFLAGS` or a signing identity:

```sh
RUSTUP_TOOLCHAIN=1.98.0 cargo xtask macos package \
  --output target/release/OpenLogi-pr-review-20260907-arm64.dmg
hdiutil verify target/release/OpenLogi-pr-review-20260907-arm64.dmg
plutil -p target/release/bundle/osx/OpenLogi.app/Contents/Library/LaunchAgents/org.openlogi.agent.service.plist
```

The unsigned build-check artifact was subsequently installed by the reporter and failed SMAppService signature validation. It is not an installable release candidate. It was rebuilt from the same source using the existing inside-out ad-hoc signing pipeline:

```sh
RUSTUP_TOOLCHAIN=1.98.0 cargo xtask macos package --sign-identity=- \
  --output target/release/OpenLogi-local-adhoc-20260907-arm64.dmg
```

The replacement DMG and installed bundle pass signature/integrity checks (`hdiutil verify` and `codesign --verify --deep --strict --verbose=2 /Applications/OpenLogi.app`). Installation still exposed a launch-constraint/stale-BTM service failure: launchd referenced an invalidated background-service record. Targeted service removal followed by the GUI's normal SMAppService registration restored supervised Agent startup, IPC responses and the normal GUI. TCC separately rejected the old Agent permission record because the ad-hoc code requirement had changed. After permission renewal, the reporter confirmed on 2026-09-08 that the current installation works and approved this PR.

No application code changed during that service/permission recovery, and the user's button configuration was preserved. This is a successful local recovery, **not** proof of seamless or unattended upgrades. Stable Developer ID signing/notarization was not available or tested. Release packaging also reports an existing unused-import warning in unchanged `windows/settings/diagnostics.rs`; the strict development-profile Clippy gate above passes.

Regression coverage includes queue wait versus wire timeout, cancellation ownership, per-node Bluetooth fallback preference, Unifying slot/register decoding, stale slot arrivals, warm/replacement channel cache adoption, cross-route restoration debt, device identity/artwork hydration, precision-mode metadata, launchd handoff/quit and visible-wake sequencing, and backup-preserving same-schema config migration. New migration/identity fixtures use synthetic device identifiers.

### Hardware verification and remaining boundaries

- On 2026-09-07, the reporter confirmed validation of both MX Ergo S and Ergo M575 using the installed local fixes and approved resubmission. Earlier testing in this report covered receiver/Bluetooth switching and shortcut behavior.
- On 2026-09-08, the reporter confirmed the final-tip installation works after the service/permission recovery above. The running launchd-supervised Agent serves the installed CLI's `list` through IPC, and that snapshot shows M575 online over Unifying with full model data. A separate exhaustive final-tip retest of every control on both models was not recorded.
- Local logs still contain node probe-retry warnings; their remaining paths have not been characterized by a long-duration soak. A working foreground session is not evidence of zero recovery warnings or unlimited idle stability.
- No claim of original MX Ergo, two simultaneous identical-model mice, exhaustive long-idle/DarkWake soak, Intel macOS or Windows/Linux hardware validation.
- Re-registering a helper cannot repair an arbitrary stale macOS TCC grant or make ad-hoc signatures stable. Input Monitoring and Accessibility still belong to **OpenLogi Agent**, not the GUI; this PR does not bypass user approval or alter the computer's permissions.

### Suggested maintainer hardware check

1. Use MX Ergo S with Bolt, and M575 with Unifying. Bind forward/back and verify the actions; for Ergo S also exercise precision mode and wheel tilt.
2. Switch receiver -> Bluetooth -> receiver repeatedly. Check one record per physical mouse, the active transport/online state, retained model artwork and bindings on both routes.
3. Leave the mouse idle, wake the computer, then test buttons before opening the GUI. Repeat with a disconnect/reconnect and with the other model connected. A maintenance DarkWake should not open HID or wake the display.
4. Rebuild/reinstall the same release with a different bundle build number. Check that the registered agent points to the replacement bundle, only one hook owner runs, explicit Quit stays stopped and a failed predecessor permits supervised recovery.
5. If available, use two same-model mice with distinct unit IDs and reassign a receiver slot; verify identities/configuration and old cleanup stay isolated.

Replaces #1024.

Fixes #366

### DIFF
--- a/crates/openlogi-agent-core/src/capture_plan.rs
+++ b/crates/openlogi-agent-core/src/capture_plan.rs
@@ -23,8 +23,10 @@ use tokio::sync::watch;
 
 /// Hardware identity of one HID++ capture session.
 ///
-/// Equality is the rearm contract: changing any field requires restoring the
-/// old firmware diversion before a replacement session may start.
+/// Equality is the rearm contract: changing any field requires the old session
+/// to finish teardown before a replacement may start. Failed restoration on
+/// another route stays parked until that route is used again; it must not
+/// disable capture on the device's newly active transport.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CaptureTarget {
     /// Physical identity used to serialize firmware ownership even when the
--- a/crates/openlogi-agent-core/src/watchers/gesture.rs
+++ b/crates/openlogi-agent-core/src/watchers/gesture.rs
@@ -21,8 +21,9 @@
 //! way regardless.
 
 mod dispatch;
+mod restores;
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
@@ -38,6 +39,7 @@ use tokio::time::Instant;
 use tracing::{debug, warn};
 
 use self::dispatch::InputDispatcher;
+use self::restores::{PendingRestore, RestoreQueue};
 use super::capture_session::{CaptureSession, CompletionAction, ReconcileAction};
 use crate::capture_plan::{CaptureTarget, DeviceCapturePlan, DispatchPlan, SharedCapturePlans};
 use crate::receiver_access::{ReceiverAccess, ReceiverRequestState, SessionReceiverLease};
@@ -139,14 +141,9 @@ enum SessionEvent {
     Done(SessionDone),
 }
 
-struct PendingRestore {
-    token: PendingCaptureRestore,
-    retry_at: Instant,
-}
-
 struct GestureManagerState {
     sessions: HashMap<PhysicalDeviceKey, RunningSession>,
-    pending_restores: HashMap<PhysicalDeviceKey, PendingRestore>,
+    pending_restores: RestoreQueue<PendingCaptureRestore>,
     restart_after: HashMap<PhysicalDeviceKey, Instant>,
     input_dispatcher: InputDispatcher,
     lease: std::sync::Weak<SessionReceiverLease>,
@@ -283,27 +280,22 @@ fn acquire_session_lease(
 }
 
 async fn retry_pending_restores(
-    pending_restores: &mut HashMap<PhysicalDeviceKey, PendingRestore>,
+    pending_restores: &mut RestoreQueue<PendingCaptureRestore>,
+    due: Vec<PendingRestore<PendingCaptureRestore>>,
     registry: &openlogi_hid::ChannelRegistry,
-    now: Instant,
+    device_io: &DeviceIoGate,
 ) {
-    let keys: Vec<_> = pending_restores
-        .iter()
-        .filter(|(_, pending)| pending.retry_at <= now)
-        .map(|(key, _)| key.clone())
-        .collect();
-    for key in keys {
-        let Some(pending) = pending_restores.remove(&key) else {
+    for pending in due {
+        if !device_io.allows_io() {
+            pending_restores.retain(pending);
             continue;
-        };
+        }
         if let CaptureSessionOutcome::RestorePending(token) = pending.token.retry(registry).await {
-            pending_restores.insert(
-                key,
-                PendingRestore {
-                    token,
-                    retry_at: Instant::now() + RETRY_DELAY,
-                },
-            );
+            pending_restores.retain(PendingRestore {
+                token,
+                retry_at: Instant::now() + RETRY_DELAY,
+                ..pending
+            });
         }
     }
 }
@@ -311,15 +303,14 @@ async fn retry_pending_restores(
 fn next_deadline(
     requests: ReceiverRequestState,
     device_io_allowed: bool,
-    pending_restores: &HashMap<PhysicalDeviceKey, PendingRestore>,
+    restore_deadline: Option<Instant>,
     restart_after: &HashMap<PhysicalDeviceKey, Instant>,
 ) -> Option<Instant> {
     if requests.any() || !device_io_allowed {
         return None;
     }
-    pending_restores
-        .values()
-        .map(|pending| pending.retry_at)
+    restore_deadline
+        .into_iter()
         .chain(restart_after.values().copied())
         .min()
 }
@@ -328,31 +319,50 @@ fn restart_deadline(unexpected: bool, now: Instant) -> Option<Instant> {
     unexpected.then_some(now + RETRY_DELAY)
 }
 
+fn retain_restart_delays(
+    restart_after: &mut HashMap<PhysicalDeviceKey, Instant>,
+    published: &[DeviceCapturePlan],
+    now: Instant,
+) {
+    // An elapsed startup delay no longer constrains capture. Keeping it while
+    // restoration is pending makes the manager's select deadline permanently
+    // ready, spinning instead of waiting for the next bounded restore retry.
+    restart_after.retain(|key, deadline| {
+        *deadline > now
+            && published
+                .iter()
+                .any(|plan| plan.target.physical_key == *key)
+    });
+}
+
 impl GestureManagerState {
     fn new(outputs: GestureOutputs) -> Self {
         Self {
             sessions: HashMap::new(),
-            pending_restores: HashMap::new(),
+            pending_restores: RestoreQueue::default(),
             restart_after: HashMap::new(),
             input_dispatcher: InputDispatcher::new(outputs),
             lease: std::sync::Weak::new(),
         }
     }
 
-    fn deadline(&self, requests: ReceiverRequestState, device_io_allowed: bool) -> Option<Instant> {
+    fn deadline(
+        &self,
+        requests: ReceiverRequestState,
+        device_io_allowed: bool,
+        published: &[DeviceCapturePlan],
+    ) -> Option<Instant> {
+        let busy = self.sessions.keys().cloned().collect();
         next_deadline(
             requests,
             device_io_allowed,
-            &self.pending_restores,
+            self.pending_restores.next_deadline(&busy, published),
             &self.restart_after,
         )
     }
 
     fn expedite_pending_restores(&mut self) {
-        let now = Instant::now();
-        for pending in self.pending_restores.values_mut() {
-            pending.retry_at = now;
-        }
+        self.pending_restores.expedite(Instant::now());
     }
 
     async fn reconcile(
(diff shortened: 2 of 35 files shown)


Chernicharo:
Tested this branch (`715d3906`) on Linux with an original MX Ergo (wpid `406f`, fw `MPM06.03_B0022`) on a Unifying receiver (`c52b`) — the Unifying/Linux combination listed as untested. Enumeration, features, controls and battery all work, same as master.

One identity problem, though. `assemble_unifying_device` says the extended pairing register "is the same physical unit id HID++ 2.0 reports over Bluetooth", and overwrites `model_info.unit_id` with it. On this unit they differ:

```
receiver  0xB5/0x30 (extended pairing, slot 1): 30 f7 b7 12 75 1e 00 00 00 01 …  -> f7b71275
device    0x0003 getDeviceInfo:                 03 21 62 d0 10 00 06 b0 1d 40 6f …  -> unitId 2162d010
```

master's `openlogi list` shows `unit_id=2162d010` (from `0x0003`); this branch shows `unit_id=f7b71275`. Since the orchestrator keys the physical identity on `DeviceIdentity::from_parts(serial, unit_id)` and this device has no serial, the receiver route and the Bluetooth route (which reports `2162d010` via `0x0003`) would get different identities — the opposite of the receiver/Bluetooth reconciliation this PR aims for. I haven't tested the Bluetooth side on this unit, so that last part is inferred from the `0x0003` value.

Suggestion: keep the `0x0003` unit ID when the feature probe has one, and use the register value only as a receiver-scoped cache key (or as the identity while the device is offline), rather than as the transport-independent ID. Solaar treats the two separately as well (receiver "serial" vs device `unitId`).

Happy to test a revision on this hardware.

Issue #1145, opened 2026-08-29
[Bug]: Mouse does not follow keyboard on Easy-Switch host change


Krisstof:
### Pre-flight checklist

- [x] I searched [existing issues](https://github.com/AprilNEA/OpenLogi/issues?q=is%3Aissue) and this is not a duplicate.
- [x] I am on the [latest release](https://github.com/AprilNEA/OpenLogi/releases/latest) or a recent `master` build.
- [x] I quit **Logi Options+** before running OpenLogi (the two apps fight over HID++ access and only one can own a receiver at a time).

### Which part of OpenLogi?

Both

### OpenLogi version

0.8.1

### Operating system

macOS

### OS version & architecture

macOS 26, Apple Silicon (M2 Max)

### Device model

MX Keys S + MX Master 3S, testé via Logi Bolt receiver ET en Bluetooth direct

### How is the device connected?

Logi Bolt receiver

### Affected area(s)

- [ ] Device discovery / detection
- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Battery status
- [ ] Settings / configuration (TOML)
- [ ] Auto-update
- [ ] Menu bar / tray
- [x] Other

### What happened?

Per the docs/FAQ, "OpenLogi drives Easy-Switch host switching over HID++ (0x1814/0x1815), and paired mice follow the keyboard when it switches hosts." In my setup (MX Keys S + MX Master 3S), pressing the keyboard's Easy-Switch key always switches the keyboard correctly, but the mouse never follows — tested exhaustively:

- macOS ↔ Ubuntu 26.04, both directions
- Both devices Bluetooth-direct, and re-paired via a Logi Bolt receiver (ruling out transport as the cause)
- Easy-Switch channel numbers confirmed matching between keyboard and mouse (1=macOS, 2=Ubuntu)
- All required Input Monitoring permissions granted on macOS, individually, to OpenLogi, OpenLogi Agent, and the terminal app used for diagnostics
- Ran the agent with RUST_LOG=debug during a full switch round-trip: no log line whatsoever references ChangeHost, 0x1814, HostsInfo, or any attempt to command the mouse — only routine gesture-session bookkeeping

[...]

PR #1334, opened 2026-09-10
perf(hid): keep a directly connected device's probe across reconnects

jpvieirapereira:
A directly connected device re-walks its entire feature table when it comes back from another computer. On an MX Keys S over Bluetooth that costs about 2.5s, during which the device is enumerated but nothing that needs its feature table can use it yet.

Two things cause it, and both come down to keying a device on where it was attached rather than on what it is.

## The key

`CacheKey::Direct` is the OS node id, and a device that leaves and returns arrives under a fresh one, so every return is a cache miss.

The backend already reports a serial for these devices. On Bluetooth it is the address, which survives both the trip and a reboot:

```
vid=046d pid=b378 serial=Some("d9:d2:69:b2:4b:96")   MX Keys S
vid=046d pid=b042 serial=Some("d1:ec:e1:ac:91:43")   MX Master 4
```

`NodeInfo::identity` already prefers a serial over the node id for exactly this reason, so the cache key does the same now, falling back to the node id only where no serial can be read.

## The eviction

The entry was also dropped while the device was away, past the three-tick grace meant for transient receiver timeouts. A trip to another computer is far longer than that, so the key alone would not have helped.

`CacheKey::names_the_device` makes the existing distinction explicit. An identity key cannot be handed to a different physical device, so what it memoizes stays true however long that device is out, and those entries are kept. A pairing slot or an OS node id can belong to something else the moment the device leaves, so those still evict on the same grace.

This also makes memory agree with disk. Bolt entries were already persisted across restarts while the in-memory cache dropped them after three ticks, which is hard to justify in either direction.

## What this does not do

It does not remove the walk for every reconnect. `is_stale` is `now - probed_at >= REFRESH_INTERVAL`, and a retained entry keeps its original `probed_at`, so a device away for longer than the 30s refresh window still returns to a full walk. What this removes is the walk for a device that steps out and comes back inside that window, which is the common case for host switching, and it is the whole of the gain being claimed here.

Closing the rest means serving the memoized entry immediately and letting the self-heal walk run without blocking it. Immutable data is never wrong, and the refresh exists only to catch something like a firmware update reshuffling the table, so it has no business holding up a device that is already known. That is a change to *when* the refresh happens rather than to what is cached, so it is not folded in here.

Resetting `probed_at` when an entry survives an absence would be one line and is the wrong line: a device that hops every few seconds would then never self-heal at all, and those are exactly the devices this targets.

## Persistence

Persistence follows the same rule rather than naming Bolt alone, so a serial-keyed direct device survives a restart as well. Schema 4; older snapshots are discarded as they were before.

## Testing

- `cargo test`: 1127 pass across the touched crates; clippy and fmt clean
- The eviction test is now two, one per side of the rule: a slot-keyed entry still evicts past the grace, an identity-keyed one is kept however long the device is away
- New tests cover key selection, including the fallback when the backend reports no serial or an empty one

I found this while working on #1145, where the window between a keyboard reconnecting and being usable is long enough for a user to press a key inside it. That case is not the argument for this change, though: every directly connected device pays the same walk on every reconnect today.

I do not have macOS or Windows hardware to measure there. The change is platform-neutral and degrades to the current behaviour wherever a backend reports no serial, but the benefit depends on those backends reporting one.

### DIFF
--- a/crates/openlogi-device/src/inventory.rs
+++ b/crates/openlogi-device/src/inventory.rs
@@ -871,6 +871,13 @@ impl Enumerator {
             .cloned()
             .collect();
         for key in missing {
+            // A key that names the device memoizes data that stays true while
+            // the device is away, and dropping it is what makes a device that
+            // steps out and back re-answer questions it has already answered.
+            // What it costs to keep is a few hundred bytes per device ever seen.
+            if key.names_the_device() {
+                continue;
+            }
             let misses = self.misses.entry(key.clone()).or_insert(0);
             *misses += 1;
             if *misses > CACHE_MISS_GRACE {
--- a/crates/openlogi-device/src/inventory/cache.rs
+++ b/crates/openlogi-device/src/inventory/cache.rs
@@ -31,11 +31,29 @@ pub(super) enum CacheKey {
     /// two receivers whose serials share a common prefix (e.g. "DA2699E1" and
     /// "DA2604F2" share "DA2").
     UnifyingSlot { receiver_uid: String, slot: u8 },
-    /// Direct (Bluetooth/USB): the OS-assigned HID node id (macOS registry-entry
-    /// id, Linux dev path, Windows interface path). Unique *per node*, so two
-    /// units of the same model never collide, and stable while connected so the
-    /// cache still hits across ticks.
+    /// Direct (Bluetooth/USB) with a serial the backend could read. For a
+    /// Bluetooth device that is its address, which survives both the device
+    /// leaving for another host and the machine rebooting.
+    DirectSerial(String),
+    /// Direct (Bluetooth/USB) with no readable serial: the OS-assigned HID node
+    /// id (macOS registry-entry id, Linux dev path, Windows interface path).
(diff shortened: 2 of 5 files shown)


jpvieirapereira:
Valid. `is_stale` is `now - probed_at >= REFRESH_INTERVAL`, and a retained entry keeps its original `probed_at`, so an absence longer than 30s still ends in a full walk.

That makes this a smaller change than the description claimed, and I have corrected the description rather than leave it overstating what it does. It removes the walk for a device that steps out and comes back within the refresh window, which is the common case for host switching, and does nothing for a longer absence.

Closing that gap properly means serving the memoized entry immediately and letting the self-heal walk run without blocking, since immutable data is never wrong and the refresh exists only to catch a firmware update reshuffling the table. That is a change to when the refresh happens rather than to what is cached, so I would rather propose it separately than fold it in here.

The alternative of resetting `probed_at` when an entry survives an absence would be one line, but it would mean a device that hops every few seconds never self-heals at all, and those are precisely the devices this targets.

PR #1344, opened 2026-09-10
fix(i18n): name the real update host in the consent string

### LINKED ISSUE #1342: [Bug]: Update consent dialog says "check GitHub" but the check goes to updates.openlogi.org
**Which part of OpenLogi?** GUI (desktop app)

**OpenLogi version:** 0.8.3

**Operating system:** macOS (string is platform independent)

**OS version & architecture:** macOS 26.x, Apple Silicon (arm64)

**Device model:** N/A (UI string, no device involved)

**How is the device connected?** Other / not sure

**Affected area(s):** Auto-update

### What happened?

The update consent text reads "OpenLogi can check GitHub for a new version once per launch" (`crates/openlogi-ui/locales/en.toml:417`). The actual endpoint is `https://updates.openlogi.org/channels/stable/latest.json` (`crates/openlogi-desktop/src/platform/updater.rs:25`).

For a user auditing outgoing connections this wording is misleading: the traffic does not go to github.com. Expected: the text names the real host.

The same string exists in every locale catalog at line 417 (pl, fr, de, es, and the rest), so the fix touches all of them. The catalog parity test should catch any locale missed.

### Steps to reproduce

1. Fresh install, open Settings.
2. Read the update consent description ("Check for updates?" dialog).
3. Compare with the manifest URL in `crates/openlogi-desktop/src/platform/updater.rs:25`.

dajiaohuang:
## Summary

The update-consent string told users OpenLogi checks **GitHub** for a new version, but the client fetches its manifest from `https://updates.openlogi.org/channels/stable/latest.json`, so the copy named a host the check never contacts.

The string is a leftover from before the updater moved to the static manifest: `release.yml` generates `latest.json` with xtask and uploads it to R2 rather than attaching it to the GitHub release, and the only GitHub URL left in the asset/update path is the `User-Agent`.

## Changes

- `openlogi-ui`: `update_consent_description` now names `updates.openlogi.org` in all 23 catalogs, so the key set, key order, and `%{}` placeholders stay aligned. Two locales needed the sentence restructured around the new token rather than a straight swap (`fi`, `pl`).

## Testing

- `cargo test -p openlogi-ui locale` — catalog parity passes (`locale_files_have_the_same_keys`).
- `cargo fmt --all -- --check` — clean.
- `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test --workspace` — one failure, `xtask` `ci_yml_runs_what_this_runner_runs`. It is **pre-existing and unrelated**: it reproduces on a pristine `git worktree` at the base commit with none of these changes applied, and it asserts on `.github/workflows/ci.yml` versus the xtask job plan, neither of which this change touches.
- Not runtime-tested on hardware. To verify: open Settings → Updates and confirm the consent text under the update-checking toggle reads `updates.openlogi.org` rather than `GitHub`.

## Overlap

Open PR #1326 rewrites this same `es.toml` line while keeping the `GitHub` wording, so whichever of the two lands second needs a trivial rebase on that one line. I did not fold this into #1326 because it is another contributor's branch and its scope is the `es` catalog only, whereas all 23 catalogs carry the same wrong host.

Fixes #1342


dajiaohuang:
The configured-host caveat is valid for builds that set `OPENLOGI_UPDATE_MANIFEST_URL`; this PR keeps the consent copy aligned with the shipping default `updates.openlogi.org` across the official locale catalogs. Making the host dynamic in localized consent text would be a separate design change, so I am leaving this non-blocking finding unchanged.

Issue #428, opened 2026-07-19
[Feature]: Smart Zoom support


taylor325:
### Pre-flight checklist

- [x] I searched existing issues and the [Roadmap](https://github.com/AprilNEA/OpenLogi#roadmap), and this isn't already tracked.

### Problem / motivation

I want to completely ditch Logi Options+ because it is buggy and bloated with telemetry. OpenLogi is the perfect lightweight alternative, but it is currently missing one crucial feature that keeps me from switching: Smart Zoom (the native macOS double-tap to zoom functionality).

<img width="517" height="849" alt="Image" src="https://github.com/user-attachments/assets/0d87b40b-9d43-44bb-bde2-64895dfc4685" />

Without this feature, it is difficult to fully transition away from the official Logitech software, as Smart Zoom is one of the most useful daily shortcuts on macOS. If this gets added, I will immediately switch over to OpenLogi as my primary driver.

### Proposed solution

Add a dedicated "Smart zoom" action to the button mapping list (as shown in the attached screenshot from the official Logitech software).

It should allow users to easily assign the native macOS Smart Zoom feature to any customizable mouse button. When the assigned button is pressed, it should trigger the exact same system action as a double-tap on a Magic Mouse or trackpad.

### Related area(s)

- [x] GUI
- [ ] CLI
- [x] Button actions / remapping
- [ ] DPI
- [ ] SmartShift
- [ ] Per-application profiles
- [ ] Configuration (TOML)
- [ ] Auto-update
- [ ] Other

PR #1354, opened 2026-09-11
feat(binding): add thumb-wheel zoom

mehmetnadir:
## What

Adds `Zoom In` / `Zoom Out` as bindable actions, so the thumb wheel can zoom.

Addresses the thumb-wheel half of #360 — the thread's original request ("the
thumb wheel can't be assigned the zoom function"). It deliberately does **not**
implement macOS *Smart Zoom* (the `0x9F` reprog-control task raised later in
that thread and in #428): that is a one-shot toggle bound to a **button**, a
different action with a different dispatch. Happy to follow up with it
separately if you'd like it in the same release.

## How

`Action::ZoomIn` / `ZoomOut` lower to a platform-neutral
`Effect::Zoom(ZoomDirection)`, dispatched per OS in `openlogi-inject`:

**macOS** synthesises a **magnification gesture** — a `CGEvent` carrying
`NSEventTypeGesture` (29) with `kIOHIDEventTypeZoom` (8) — rather than
⌘+scroll. This matters: macOS has no ⌘+scroll page zoom. AppKit and WebKit
zoom on *magnify* gestures, which is why Logi Options+ feels like a trackpad
pinch and zooms at the cursor rather than at the viewport centre.

I initially implemented ⌘+scroll and an event-level probe appeared to confirm
it worked — it did not. The probe confirmed the events were posted, not that
anything zoomed. In the field, ⌘+scroll zoomed a `<canvas>` app (which reads
the modifier itself) while every ordinary web page just scrolled. The gesture
path behaves like Options+ everywhere.

The wheel opens **one gesture session** rather than a gesture per tick:
`Began` on the first increment, `Changed` while it turns, and an idle watchdog
posts `Ended` once the wheel goes quiet. Browsers discard self-contained
per-tick gestures, so joining the session is what turns a visible tick-tick-tick
into continuous zoom.

**Windows / Linux** accumulate the fractional magnification into whole
Ctrl+wheel notches, which *is* the native page-zoom idiom on both.

## Continuous dispatch path

Zoom routes through `advance_zoom` (continuous), not `advance_action`
(discrete). The discrete path exists to debounce one-shot actions and applies a
threshold plus a 200 ms cooldown; running zoom through it made the wheel emit
about one 5% step per flick instead of tracking the rotation.

## Its own sensitivity setting

Zoom gets a `zoom_sensitivity` setting — app-wide default plus per-device
override, the same shape as `thumbwheel_sensitivity` — rather than borrowing
the scroll slider. Comfortable scrolling and comfortable zooming are not the
same number of degrees per unit; one slider cannot serve both, and the first
version that shared it was usable for neither.

Committing the app-wide default **clears** the device override, so a device
tuned once still follows later changes in Settings → General instead of pinning
whatever the default was that day.

The config field is additive and defaults, so existing `config.toml` files load
unchanged — there is a test for that rather than a claim.

## Also included

`Category::Zoom`, ring-gallery icons, the `Zoom` / `Zoom (reversed)` thumb-wheel
presets, and strings for all 23 locales (key set **and** order verified against
`en.toml`, so `catalog_parity` stays green).

`persisted_action_variant_names_are_stable` was updated deliberately: the two
new variants are appended, never reordered, per the on-disk schema contract.

One drive-by: the two sensitivity sliders in `pointer/smartshift.rs` were
near-identical copies, so they now share a `sensitivity_slider` helper. That
keeps `SmartShiftPanel::new` under the 100-line clippy limit and stops the pair
from drifting apart.

## Verification

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test --workspace` (GUI suite excluded) | 1563 passed, 0 failed |
| 23-locale key set + order parity | verified |
| macOS release build + bundle | ok |

New tests: zoom lowers to its own effect (not a scroll); zoom emits on every
increment instead of a threshold+cooldown; magnification scales with rotation;
zoom reads `zoom_sensitivity` while scrolling ignores it; device override beats
app default and clearing falls back; a `config.toml` without the field loads at
the default; committing the app default clears the device override.

The override-clearing test was checked by mutation — forcing the setter to
always store a value turns it red, so it is testing the behaviour and not just
running the code.

Built and used daily on a real MX Master 3S on macOS 15 (Darwin 25.6) while
developing this.


mehmetnadir:
Note on CI: the `CI` and `Build` workflows are sitting in `action_required`
under the fork-PR approval gate, so they haven't run — the only check that
reports here is the Greptile review. A maintainer approval would let them run.

For what it's worth, the same configurations were verified locally beforehand:
`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D
warnings` and `cargo test --workspace` are clean, and since neither quantized
backend builds on macOS I cross-checked both with `cargo check --target
x86_64-pc-windows-msvc` and `--target x86_64-unknown-linux-gnu`.

Also, for the record: #764 covers the same ground and predates this by three
weeks, reaching the same conclusion about magnification gestures independently.
#1175 and #1313 overlap too. I have no stake in which one lands — whichever
approach you prefer, I'm happy to fold anything useful from here into it, or to
close this in favour of another. The two pieces that might be worth carrying
over regardless are the separate zoom sensitivity setting (app-wide default
plus per-device override, since the wheel rotation that scrolls comfortably
isn't the one that zooms comfortably) and resetting the Ctrl+wheel remainder on
direction and idle boundaries.

atomiczsec:
this would be so useful!

LinyueSong:
heavy plus

PR #1376, opened 2026-09-12
perf(hook): bypass the macOS input filter for unheld pointer motion

tomhschmidt:
## Summary

Stop subscribing to ordinary `MouseMoved` events in the macOS filtering HID tap. Even returning `PassThrough` makes cursor delivery wait for the agent's event-tap thread. Ordinary motion has no active button hold to update, and the event monitor discards movement events.

This removes an unnecessary synchronous step from cursor delivery. It is a candidate fix for desktop-wide pointer stutter reported with a Bluetooth MX Master 3, where stopping the OpenLogi agent improved smoothness. It does not establish that this is the only cause or claim a measured FPS improvement.

## Changes

- **openlogi-hook:** Remove `MouseMoved` from the tap mask; retain all three `*MouseDragged` variants, buttons, scrolling, and keys.
- Add a regression test asserting ordinary motion bypasses the tap while gesture drags remain subscribed and translate with their original deltas. Confirmed the test fails with the original mask.
- No new dependencies or configuration changes.

## Testing

Passed on Apple Silicon / macOS 26.6.2 with Rust 1.98.1:

```sh
cargo fmt --all -- --check
git diff --check
RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --locked -- -D warnings
RUSTFLAGS="-D warnings" RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --exclude openlogi-ui --exclude openlogi-desktop --exclude openlogi-overlay --exclude openlogi-agent --locked
RUSTFLAGS="-D warnings" cargo test -p openlogi-hook -p openlogi-agent-core -p openlogi-agent
RUSTFLAGS="-D warnings" cargo test -p openlogi-desktop -p openlogi-overlay
```

542 unit tests and one doctest passed across the hook's affected dependency closure. Independent source review found no actionable regressions. Windows/Linux runtime validation was not performed; this patch only changes the macOS implementation.

**Hardware check:** A temporary native HID tap suppressed the physical MX Master 3 middle-button down/up. During the hold, both the HID and downstream session observers received 260 `OtherMouseDragged` events and zero `MouseMoved` events. This verifies the OS still supplies the retained drag event type when the button edge is suppressed. It is not an end-to-end configured-gesture test.

**Draft / remaining validation:** Compare visible smoothness with the patched agent and exercise configured middle/back/forward gestures. The installed v0.8.3 agent runs under Rosetta on this machine; architecture is a separate confounder, not a demonstrated cause. The installed app has not been replaced.

### DIFF
--- a/crates/openlogi-hook/src/macos.rs
+++ b/crates/openlogi-hook/src/macos.rs
@@ -579,8 +579,6 @@ const CALLBACK_WATCHDOG_POLL_INTERVAL: Duration = Duration::from_millis(20);
 const LIFECYCLE_WATCHDOG_POLL_INTERVAL: Duration = Duration::from_millis(100);
 const FREEZE_HAZARD_EXIT_CODE: i32 = 78;
 
-/// Event types the HID tap observes. Pointer *Dragged variants are required
-/// because a held button makes the OS emit those instead of `MouseMoved`.
 /// The macOS backend: a `CGEventTap` serviced by a private run-loop thread.
 pub(crate) struct Backend;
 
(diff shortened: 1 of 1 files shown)


tomhschmidt:
Closing this proposed fix: switching the affected machine from the Intel/Rosetta build to the native Apple Silicon build of the same version (v0.8.3) resolved the observed cursor/scroll stutter. The running agent was verified as ARM64. This patch was never installed for the comparison, so we do not have evidence that it is needed to address the reported issue.

PR #1396, opened 2026-09-15
feat(gui): add custom-shortcut Tap/Hold recorder to the button picker

fahadjibransheikh:
## Summary

- `Action::CustomShortcut` / `Action::HoldShortcut` and the `KeyCombo` parser already exist end-to-end in `openlogi-core` (config schema, IPC, agent runtime dispatch), but there was no way to reach either from a physical button binding — the catalog picker deliberately excludes both, and the only shortcut recorder in the app lives inside the Actions Ring slot editor.
- Adds a "Custom shortcut" section (Tap/Hold choice + free-text `KeyCombo` input) to the shared button-binding picker, so mouse buttons and the 5-direction gesture map (`HapticPanel`/`GestureButton`) can record a custom or held shortcut without hand-editing `config.toml`.
- Came up pairing a haptic-pad button with a third-party hold-a-shortcut launcher (e.g. Droppy's Ring) — that binding was only reachable by hand-editing the config file before this change.

## Changes

- `crates/openlogi-desktop/src/features/mouse/view.rs`: `MouseModelView` gains `shortcut_input`/`shortcut_hold` fields (built the same way as the existing `action_search`), a `set_shortcut_hold` setter, and a `localize_inputs` helper (kept `render()` under the workspace's `too_many_lines` limit) that re-derives both inputs' placeholders on a live language switch.
- `crates/openlogi-desktop/src/features/mouse/inspector.rs`: `ActionPickerContext` carries the two new fields; the shared `action_library()` (used by the plain-button, inherited-gesture, and gesture-direction inspectors) gets a new `custom_shortcut_editor` section appended after the catalog list, committing through the existing `on_pick: &PickFn` — no new commit plumbing.
- `crates/openlogi-ui/locales/*.toml` (all 23): four new keys under `[actions]` — `custom_shortcut`, `shortcut_placeholder`, `shortcut_hold`, `shortcut_tap` — at the same position in every file. Non-English values are the English source text for now, per the i18n rule (Crowdin improves them later).

## Testing

- `cargo clippy -p openlogi-desktop -p openlogi-ui --all-targets -- -D warnings` — clean (one real finding along the way: `MouseModelView::render` crossed the 100-line `too_many_lines` limit after adding a placeholder call; fixed by extracting `localize_inputs` rather than suppressing the lint)
- `cargo test -p openlogi-desktop -p openlogi-ui` — 221 tests pass, including `locale_files_have_the_same_keys` (catalog parity) and `services::i18n::tests::typed_translation_keys_resolve`
- GUI exercised end-to-end against `openlogi-agent-mock` (real hardware wasn't reachable for this dev build — TCC permissions for an ad-hoc-signed bundle, a local dev-environment limitation, not something this diff touches): bound a plain button (Middle Click) to Hold+chord and a gesture-direction row (the same code path `HapticPanel`/`GestureButton` use) to Tap+chord, both committed and rendered correctly and distinctly.
- **Not runtime-tested on real hardware** — the mock-agent verification above exercises the same code path, but real-device confirmation is the maintainer's call per `AGENTS.md`. Happy to attach screenshots from the mock-agent verification in a comment if useful.


fahadjibransheikh:
Fixed both Greptile findings in 3936e3ef:

- **Shortcut state persists**: `shortcut_input`/`shortcut_hold` now reset in the same opening-branch of `selection_card`'s click handler that already clears `action_search`, so switching to a different button/gesture-direction no longer shows a stale draft.
- **Invalid input fails silently**: Add is now disabled while the current text doesn't parse as a `KeyCombo` (covers empty input, modifier-only combos, unknown keys, malformed separators). Needed a live-updating subscription on `shortcut_input` (mirroring the existing one on `action_search`) so the disabled state tracks each keystroke rather than only the last unrelated re-render.

`cargo fmt --check`, `cargo clippy -p openlogi-desktop --all-targets -- -D warnings`, and `cargo test -p openlogi-desktop` (218 tests) all green on the new commit.

PR #1406, opened 2026-09-16
fix(gui): cap the label card's value row so long bindings ellipsise

### LINKED ISSUE #1401: [Bug]: Scaling Bug
### Pre-flight checklist

### Which part of OpenLogi?

GUI (desktop app)

### OpenLogi version

0.8.4

### Operating system

Windows

### OS version & architecture

W11 Pro 25H2

### Device model

MX Anywhere 3S

### Affected area(s)

- [ ] Button remapping
- [ ] DPI control
- [ ] SmartShift
- [ ] Battery status
- [ ] Auto-update
- [x] Menu bar / tray
- [x] Other

### What happened?

Theres a small display issue, see screenshot.
My display resolution is 3200x2000 with a 175% zoom in Windows

Image

### Steps to reproduce

just open the settings

### Diagnostics report

```markdown
### OpenLogi Diagnostics

**App**
- OpenLogi (GUI): v0.8.4 (release)
- Agent: v0.8.4 (connected)
- IPC protocol: GUI 30 / agent 30
- Inventory: ready
- OS: Windows (x86_64)
- Launch at login: yes · Menu bar: no · Update check: off
- Config: schema 7 · 1 configured device(s) · thumbwheel 14

**Assets**
- Cache path: C:\Users\moritz.schuenemann\.local/share/openlogi\assets · Bundle assets: absent

**Devices (1)**
- MX Anywhere 3S — mouse (codename: MX Anywhere 3S)
  - Connection: Bluetooth (direct) · Online: yes · Battery: 45% (discharging, good)
  - DPI: 1000 dpi (range 200–8000, 157 steps)
  - Model: 0b037 (model-ids: b037/0000/0000, ext-model: 00)
  - Transports: BTLE
  - Render: mx_anywhere_3s · direct
```

### `openlogi list` output

```text

```

### Logs

```shell

```

costajohnt:
## Summary

On the device page, the button label cards beside the mouse render overflow their fixed width when the binding text is long: the leading icon pokes out the left border and the chevron out the right, symmetrically. The report is from Windows at 175% (#1401), but it is not a scaling or platform bug: with the synthetic model on macOS the default "Forward (Button 5)", "Horizontal Scroll" and "Cycle DPI Presets" overflow the same way, and every longer locale is worse.

The card is a `gpui_base::Button`, which sets `items_center` on its base before the caller's styles land. Under our `flex_col` that is cross-axis centring, so the inner value row keeps its natural width and, once wider than the card's content box, is centred and overflows both edges. The `flex_1` / `min_w_0` / `text_ellipsis` on the value text never gets a chance because the row it lives in has no width constraint.

Fix: `.max_w_full()` on the row. Long rows are capped at the content box and the value text ellipsises as the comment there always said it would; short rows keep their centred natural width, so nothing else on the page moves.

## Changes

- `openlogi-desktop`: `features/mouse/view.rs`, `.max_w_full()` on the label card's value row, with a comment explaining the geometry. Two single-use locals in `LabelTrigger::render` are inlined so the function stays inside clippy's `too_many_lines` budget without a suppression. A sentence claiming the leading icon is absent for the gesture summary was wrong (every `BindingLabel` carries one); the comment drops it and the field doc now says why the field is still an `Option`.
- `openlogi-desktop`: `debug_selector` markers on the card and its value row, and a `#[gpui::test]` (`long_bindings_stay_inside_their_label_card`) that renders the view at 1200px and asserts each value row's bounds sit inside its card. With `.max_w_full()` removed it fails with the Forward row at 206px over a 156px card, the same symmetric overflow as the report. It pins the locale to English under `LOCALE_LOCK`: another test in the binary leaves it at zh-CN, whose labels are short enough to fit and would have made this a false pass.

## Testing

- `cargo clippy -p openlogi-desktop --all-targets -- -D warnings`
- `cargo test -p openlogi-desktop` (222 tests; three back-to-back runs to check for locale races, all green; one more run with the cap removed to confirm the new test fails)
- `cargo xtask ci` on macOS aarch64: 8 passed (rustfmt, clippy, MSRV check, rustdoc, tests with `--all-targets`, publish closure, wasm, cargo-deny), 3 not run on this host (typos, Linux tests, Windows clippy), 1 failed: `shell`, on an shfmt spacing diff in `.claude/skills/openlogi-macos-permissions/scripts/diagnose.sh`, which this branch does not touch (local shfmt version; the diff is only `view.rs`).
- Reproduced on macOS without hardware: `openlogi-agent-mock` plus the desktop app, English and Russian locales. Before/after below.
- Not tested on Windows or Linux. Not runtime-tested on hardware: the change is layout only and the mock inventory renders the same cards.

## Screenshots

Before (macOS, English, mock agent, synthetic model):

before: Forward (Button 5) and the right-hand cards spill past both borders

After:

after: long bindings ellipsise inside the card

Same pair under the Russian locale, where every card overflowed: [before](https://raw.githubusercontent.com/costajohnt/OpenLogi/75a70f8f6d9dc2e80aca7e3787ea6a78313a48ec/1401-before-ru-macos.png), [after](https://raw.githubusercontent.com/costajohnt/OpenLogi/75a70f8f6d9dc2e80aca7e3787ea6a78313a48ec/1401-after-ru-macos.png).

Reporter's original (Windows 11, 175%, MX Anywhere 3S): see #1401.

## Not in this PR

The cap fixes the overflow, but the ellipsis budget is tight: the card's content box is 130px (156 minus border and `px_3`), and the icon, gaps and chevron take 44px, so a long binding shows about 86px of text. The reporter's own "Forward (Button 5)" reads as "Forward…" on macOS and will on Windows. `aria_description` still carries the full string. If you would rather the full name be reachable for sighted users too, a `Tooltip` on the card or a wider `LABEL_W` are the two options I saw; both change more than this issue asks for, so I left them out and would take either as a follow-up if you want it.

Fixes #1401


AprilNEA:
Thanks for the precise diagnosis — it's the same gpui-kit 0.6 Button change that broke the Settings theme grid and a few other bespoke buttons, so #1408 fixes it at the root (`items_stretch` on the card, plus the other sites). Your regression test is the right guard for #1401, so I've folded it into #1408 with you as co-author and will close this one once that lands.

costajohnt:
#1408 is in and covers everything here, so closing this one. Thanks for folding the test in and for the co-author credit, and good catch that the Button change hit the other cards too.

